Issue #35, opened 2017-04-07
BundleExplorer overhaul


Traderain:
- [x] Rename to asset browser
- [x] Different view modes
- [x] Add cache files,w3speech etc.
- [ ] Asset preview
  - [ ] 3d models
  - [ ] Sounds
  - [ ] Images
  - [ ] text files
  - [ ] movies
- [x] Convert it to be purely event based so it can be kept open and import stuff as you need it.

Cooking needs to be eleminated to implement this so texture.cache/soundspc.cache etc. writers need to be implemented for this.

Traderain:
Asset preview could be achieved with [IPreviewHandler](https://msdn.microsoft.com/en-us/library/bb775315.aspx)

rfuzzo:
Related requests:
-[] make asset browser dockable #100 
-[] enhance the search #76 
-[] Pagination of results

rfuzzo:
outdated with Wolvenkit 8

PR #52, opened 2017-06-02
Implemented Witcher Interfaces for Asset browser

kote2ster:
#35


WolvenKit.Cache/Sound_Cache.cs
```diff
@@ -58,8 +53,8 @@ public void Read(BinaryReader br)
                 NamesOffset = br.ReadUInt32();
                 NamesSize = br.ReadUInt32();
             }
-            BufferSize = br.ReadInt64();
```
Traderain: This shouldn't be changed.Buffersize and Checksum is correct.

WolvenKit.Interfaces/IWitcherArchive.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Forms;
+
+namespace WolvenKit.Interfaces
+{
+    public interface IWitcherArchive
```
Traderain: Should be renamed to IWitcherArchiveManager since it manages archive now it's not for the actual archives.

PR #69, opened 2017-08-16
Implemented support for having a DLC and Mod in the same project.

### LINKED ISSUE #55: Multi purpose project (DLC&Mod)
We should have it so you can mark a file as a DLC file or as a mod file and when packing they will be installed to the appropriate folder. If we are updating mod explorer anyway we should probably track changed files and such so a "Please save dialog" can be made when closing the app or something.

Traderain:
- Fixed some bugs in assetbrowser.
- Recoded the whole packing process.
- Reworked the way the project's files are handled.
- Resolves #55 
- Added notification for unsaved files when closing the project.
Someone please check it if you have any issues with it. I am afraid to merge it until someone tests it.


Traderain:
Known bugs:
- TextureCache is not exported to correct folder.
- Script copy is messed up.

PR #71, opened 2017-08-27
Added option to show IDs in the strings importer.

Murzinio:


Traderain:
Can you check if the gui still work with the new DLC+Mod project thing?

rfuzzo:
71

Issue #72, opened 2017-09-30
Fix build target


Traderain:
Currently Irrlicht and VLC.net doesn't really like each other so the whole project has to be x86. This needs to be fixed so we can switch of the prefer 32bit and go back to anycpu.

ChiefyWombat:
glad I saw this, I just switched Sabre to AnyCPU as well.

PR #77, opened 2018-04-04
Fix solution and project configurations

dnandha:
Solution would not build because of target architecture mismatch.
See #72 

With this changes the solution builds successfully for x64 target platform (set as default).


Traderain:
Thanks for the contribution. Your help is very much welcome. I will review your PR when I get home and if there isn't anything wrong with it. In the meantime you are welcome to join us on discord you can find a link to it in the readme.

PR #78, opened 2018-04-07
small bugfix

rfuzzo:
fixes a small bug that prevented opening w2meshes.

### DIFF
--- a/WolvenKit/Forms/frmMain.cs
+++ b/WolvenKit/Forms/frmMain.cs
@@ -818,7 +818,7 @@ namespace WolvenKit
                     {
                         var rigDoc = new frmCR2WDocument();
                         var basePath = doc.File.FileName.Split(new string[] { "characters" }, StringSplitOptions.None)[0];
-                        var modelName = Path.GetFileName(doc.File.FileName).Split('_', '.')[3];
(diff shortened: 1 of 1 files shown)


WolvenKit/Forms/frmMain.cs
```diff
@@ -818,7 +818,7 @@ public frmCR2WDocument LoadDocument(string filename, MemoryStream memoryStream =
                     {
                         var rigDoc = new frmCR2WDocument();
                         var basePath = doc.File.FileName.Split(new string[] { "characters" }, StringSplitOptions.None)[0];
-                        var modelName = Path.GetFileName(doc.File.FileName).Split('_', '.')[3];
+                        var modelName = Path.GetFileName(doc.File.FileName).Split('_', '.');
```
Traderain: If you do that modelname will be an array not a string and you need a string here:
```
 var rigPath = $@"{basePath}characters\base_entities\{modelName}_base\{modelName}_base.w2rig";
```

kote2ster:
@rfuzzo for now for mesh loading you should use the [Renderer branch](https://github.com/Traderain/Wolven-kit/tree/Renderer). I will test it on Monday.

rfuzzo:
thx. true, that was a silly suggestion. :)

PR #102, opened 2018-09-09
FIXED: DataTable Set Var now uses correct indices for DataTableEntry.…

Strahlemann83:
… Row/Column was switched

# DataTable Set Var now uses correct indices for DataTableEntry. Row/Column was switched

Fixed:
- Usage of DataTableEntry indices


Traderain:
Thanks for the contribution.

Issue #111, opened 2018-09-19
Fix compiler warnings


philippTheCat:
^title

Traderain:
I agree but we should supress CS1591

saurabhs:
I may take this, but how you guys are planning to suppress CS0649, <variable> is never assigned to, and will always have its default value 0? Remove/comment out it from class/struct for now?

NamStalker:
Can I pick this up? It looks like the broken build for their PR request is stale.

rfuzzo:
outdated with Wkit 8

PR #119, opened 2018-10-21
Fixed compile warnings.

NamStalker:
Added Task.Run() to each async task that was being called. Added a disable warning for unassigned variables within structs. Removed variables from exception catches. Removed async from Initialize, as it is already starting a seperate task.

# <PULL REQUEST TITLE>

Implemented:
- <Features implemented>

Fixed:
- <Fixes>Compiler warnings.

<Aditional notes>


NamStalker:
#111

PR #123, opened 2018-12-20
w2ent child class type support

rfuzzo:
# <PULL REQUEST TITLE>

Implemented:
- registered all CEntity child classes (unknown bytes support)

Fixed:
- disabled SDynamicDecalMaterialInfo registration (error)

<Additional notes>


rfuzzo:
I registered all child classes for CEntity *as* CEntity 

This is only if the children don’t add further custom variables and then equivalent to adding and registering the children as their own types.

If we want to go with Limes Method in the end, we should probably have them as their own types and inherit from CEntity etc.

But for now this is equivalent.

WolvenKit.CR2W/Types/CR2WTypeManager.cs
```diff
@@ -114,13 +114,335 @@ public CR2WTypeManager()
...
+            Register("W3HorseManager", new CEntity(null)); //ws
+
             // Testing area. Put your CVectors which don't work fully yet here.
 #if DEBUG
             // *.w2p
             //Register("CParticleEmitter", new CParticleEmitter(null));
             Register("CDecalSpawner", new CDecalSpawner(null));
             Register("SSeedKeyValue", new SSeedKeyValue(null));
-            Register("SDynamicDecalMaterialInfo", new SDynamicDecalMaterialInfo(null));
+            //Register("SDynamicDecalMaterialInfo", new SDynamicDecalMaterialInfo(null)); //rfuzzo: disabled because it couldn't read the stream
```
Traderain: These don't really need disabling. This section is for WIP vectors but not an error.

PR #127, opened 2019-10-23
Miscellaneous Fixes and asset browser modifications

sw3dg1n:
# Miscellaneous fixes and asset browser modifications

Implemented:
- Changed the "Search from current folder" checkbox and its functionality to "Search only in current mod folder" as due to the data structure present, the former one cannot be supported. However, searching in the current mod folder is somewhat possible at least, it just doesn't work properly if you are in a mod called <someName> and there is a second mod called <someName><suffix>.
- Significantly increased the default width of the marked files list in the asset browser as it is not resizable.

Fixed:
- Fixed a bug where the "Search from current folder" (or now "Search only in current mod folder") checkbox was ignored if the "Regex" checkbox was enabled.
- Fixed a bug which caused the bundle manager to load indefinitely during startup if a mod in the mods folder had two resources with the same name defined like with Immersive Lighting.
- Fixed a bug which caused exceptions when opening w2p files in debug mode.
- Added missing build flags for the x64 release configuration.

Feel free to only include the changes you want. Especially the changes in the asset browser are very much debatable I would say.


Traderain:
Thanks. Looks great. The w2p thing is not really a bug because it was there for testing but should be disabled in the repo anyway.

PR #130, opened 2020-02-09
Indented Chunk layout

### LINKED ISSUE #113: Indented view for chunks
Wolven Kit shows indented view for attributes and such like a treeview.

It would be really nice if we could view the chunks at an indent also. We already have parent and sometimes child attributes inside the chunks.

Also dumping the gamefile we can easily see the objects and their children in the created XML's with a little XML magic the file can be cleaned to look noice

Attaching a indented XML - dumpfile

[t_05a_mg__bear_lvl5_meshes.w2ent.xml.zip](https://github.com/Traderain/Wolven-kit/files/2437692/t_05a_mg__bear_lvl5_meshes.w2ent.xml.zip)

rfuzzo:
# Indented Chunk View

Implemented:
- indented view for chunks based on the parent-child relation between chunks with TreeListView.

<Additional notes>
closes #113 

unknown


rfuzzo:
hmm, this is very slow for larger files.

PR #144, opened 2020-03-11
Refreshing mod assets where adding file from mod.

vonLeebpl:
# Mod asset loader fix

Implemented:
Refreshing mod assets browser when adding file from mod.

Fixed:
- Removed CFont.cs from project.

<Additional notes>
- When user adds file from mod the asset list is refreshed on opening asset browser form.


Traderain:
Thanks :)

PR #163, opened 2020-05-04
CR2W Table writing fix

rfuzzo:
# CR2W Table writing fix

Fixed:
- cr2w stringtable and imports table writing fix
- many unknown byte parsing fixes

<Additional notes>
#61 #5


CR2WTests/StressTest.cs
```diff
@@ -7,487 +7,598 @@
...
+        static Dictionary<string, MemoryMappedFile> mmfs;
 
         [ClassInitialize]
         public static void Setup(TestContext context)
         {
+            mmfs = new Dictionary<string, MemoryMappedFile>();
             mc = new BundleManager();
-            mc.LoadAll("D:\\SteamLibrary\\steamapps\\common\\TW3\\bin\\x64");
-            //mc.LoadAll("C:\\Steam\\steamapps\\common\\The Witcher 3\\bin\\x64");
+            //mc.LoadAll("D:\\SteamLibrary\\steamapps\\common\\The Witcher 3\\bin\\x64");
```
Traderain, line 62: We should probably change this so we don't have to edit it in every commit. How about %TW3_DIR%/bin/x64?

PR #201, opened 2020-08-03
- Reworked wcclite dump to xml to allow for individual file export.

Maxzor:
Before this commit exporting a single file triggered a recursive
  export for the entire parent directory.
  Note : this is a terrible hack...

- Made the modexplorer treelistview to not reveal after expand,
  in case of long lists this became quickly annoying.


Maxzor:
Better option than no revealafterexpand would be to memorize the scroll location, but need to fsck this.

PR #224, opened 2020-09-10
Postgres wip

Maxzor:


Maxzor:
Dunno why the rebase stuffed me these first three commits :/
There is a cr2w rework in here :
- virtual chunk parenting
- deleting chunk parentptr
- chunk referrer reverse-lookup

PR #225, opened 2020-09-11
Postgres + Virtual chunks + custom classes

rfuzzo:
# Postgres + Virtual chunks + custom classes

Implemented:
- frmChunkList rework
- Dump CR2W to Postgresql from console.
- added custom class compiling
- all cvariable copying
- custom themeing

Fixed:
- slightly faster chunk list generation
- slightly faster create Cvariable
- removed parsing of CSpotlight Blockdata, since it reads too far
- refactored chunkpropertylist to actually work with the models

<Additional notes>


Traderain:
Is this done?

CR2WTests/StressTest.cs
```diff
@@ -58,7 +58,7 @@ public class StressTest
         [ClassInitialize]
         public static void Setup(TestContext context)
         {
-            var exedir = LookUpW3exe();
+            var exedir = new FileInfo(LookUpW3exe()).Directory.FullName;
```
Traderain, line 61: Can we make an env var like "W3_DIR" and check if thats avaliable, otherwise w3lookup otherwise error?

Traderain (changes requested):

CR2WTests/StressTest.cs
rfuzzo: Yes, we should do that, but the unit test is broken right now anyways because max changed some things in the way memorymapped bundles are read...

Maxzor: on it, my bad

rfuzzo:
> 
> 
> Can we make an env var like "W3_DIR" and check if thats avaliable, otherwise w3lookup otherwise error?

implemented in https://github.com/Traderain/Wolven-kit/pull/225/commits/9e5ed6975e19e068e379d869fb51a42eb18bd829

Issue #122, opened 2018-10-31
Directory structure on compiled project


KoalaNalle:
Currently the root folder contains all files but licence files. And it's housing some 100+ files.

I would suggest that the structure is updated to atleast /bin - folder and the root should house only standalone-executable-, readme- and config files.

Traderain:
Cleaned it up a bit. Should be better now. Let me know If you have any other ideas how to make it better structured @KoalaNalle

rfuzzo:
we could use this: https://www.nuget.org/packages/PrettyBin
![image](https://user-images.githubusercontent.com/37657287/76293759-ea6d8180-62b1-11ea-89c5-755d3c35e35e.png)

Traderain:
Oh that's cool. I tried Fody but wasn't satisfied. This looks neat.

rfuzzo:
ah I used fody on my wpf aps - for weaving, which worked nice. 
but yeah, PrettyBin seems very simple and does the trick I guess.

Traderain:
What happened with this?

rfuzzo:
maybe we can finnaly get smart about NET5 publish single file?

PR #226, opened 2020-09-11
Doing my part to save the planet

Maxzor:
Minor github action update 
  - Do not build if the HEAD commit has not changed overnight.
    Programatically this is a bit garbage as the workflow does not cancel cleanly. But whatever :),
    environmentally this is much better.
  - No need choco and 7zip, powershell has zip built-in

and some root folder cleaning for #122


Maxzor:
Will do it the right way -.-

PR #227, opened 2020-09-12
Misc code cleaning

Maxzor:
Few renames
Most compiler warnings fixed (lazy with pragmas here and there)
Misc things


Maxzor:
will fix sln tonight hopefully

PR #284, opened 2020-11-20
frmRender lods

michaelpolakatwork:
frmRender levels of detail for exporting

Fixed:
- loading all lod meshbuffers

Loading all mesh buffers using the scenegraph mesh loading rather than WKMesh since WKMesh only loads the first lod.  We need to clean up frmRender and fastRender as they both do similar things but in different ways which makes maintenance difficult.


rfuzzo:
Thank you! 
Would it be possible to merge this into branch 0.7.0-preview2 instead of master?
So that we don’t skip merging and tagging the preview1 release?

PR #325, opened 2021-01-30
Fixed warnings

phrisk:
#### Initial Warning Fixes

Fixed a handful of warnings but there's still many more to address

> Disclaimer: I haven't tested *everything* after making these changes, however the solutions build, the projects run and functionally nothing should have changed :)

Fixed:
- Exception handling:
  - Caught exceptions where the exception was incorrectly re-thrown
  - Caught exceptions where the exception was correctly re-thrown but nothing else was being done
  - Caught exceptions where the exception was correctly re-thrown but the exception instance was unused
- Async/Task methods:
  - Renamed some async methods that didn't have Async in their name
    - e.g. `WolvenKit.CR2W.Read` -> `WolvenKit.CR2W.ReadAsync` - assuming this will move back to an async MemoryMappedFile implementation in the future
  - Methods marked as async that didn't contain awaiter have had their async modifier removed. These will now return `Task.CompletedTask` or `Task.FromResult` where appropriate. Not sure why these are returning Task at all - maybe left behind from some older refactoring...?!
- Removed a few places were methods were being overriden to do nothing more than call their base implementation
- Removed a few places where code was unreachable
- Removed `RoslynPad.Editor.Windows` from WolvenKit as it was both unused and demanding a conflicting version of `Microsoft.CodeAnalysis.Common`
- Removed a bunch of unused usings 

---

#### Remaining warnings

Mostly fall under the following categories:
 - Classes that hide an inherited member of their base class without using the `new` keyword
   - I think most of these should be safe to have `new` added. However, there's some instances, like `W3Mod` where the inherited member that's being hidden has the exact same implementation? Can these just be removed?
 - Dependencies that aren't fully compatible with `.net5.0-windows`
 - Unused fields
 - Usage of `System.Reflection.Assembly.Location` that should be replaced with `System.AppContext.BaseDirectory`
   - Fairly sure this should be a safe change in all places, just didn't have time to test

### DIFF
--- a/CP77.CR2W/CR2W/CR2WExport.cs
+++ b/CP77.CR2W/CR2W/CR2WExport.cs
@@ -290,9 +290,9 @@ namespace CP77.CR2W
                         variable.Variables = TryGetClassVariables(variable.Size);
                     }
                 }
-                catch (Exception e)
+                catch (Exception)
                 {
-
+                    // TODO: Are we intentionally swallowing this?
                 }
                 
 
@@ -300,7 +300,7 @@ namespace CP77.CR2W
                 br.BaseStream.Seek(endoffset, SeekOrigin.Begin);
                 o.Variables.Add(variable);
                 }
-                catch (Exception e)
+                catch (Exception)
                 {
                     return o;
                 }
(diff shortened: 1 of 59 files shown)


Offline-R503B:
[Fix Warnings](https://trello.com/c/XVnfFBUU/114-fix-warnings)

Traderain:
Thank you!

PR #334, opened 2021-02-02
Mexh export fix

Akiway:
# Mesh export fix

Changes:
* Changed default extension in export definition
* Added `mesh` in the list of ECookedFileFormat
* Moved buffers creation in a dedicated function


WolvenKit.Common/Enums.cs
```diff
@@ -31,6 +31,7 @@ public enum EFileReadErrorCodes
     /// </summary>
     public enum ECookedFileFormat
     {
+        mesh,
```
rfuzzo, line 34: This enum should only contain extensions for which uncooking (ie. converting to a raw file format) has already been implemented. For all other cases, simply extract the buffers.

CP77.CR2W/Modkit/Uncook.cs
```diff
@@ -128,87 +128,85 @@ public static partial class ModTools
...
+            // decompress buffers
+            var buffers = GetBuffers(cr2w, br);
 
-                // decompress buffers
-                var buffers = GetBuffers(cr2w, br);
 
+            // uncook textures, meshes etc
+            return extAsEnum switch
+            {
+                ECookedFileFormat.mesh => GenerateBuffers(cr2wStream, cr2wFileName),
```
rfuzzo, line 154: we should add here the actual mesh uncooking: conversion to fbx/glb. 
until that is done, simply extract buffers normally.

rfuzzo (changes requested):
Requested: add the mesh-to-glb conversion here.

rfuzzo:
Upon further thought, let's merge this in, and keep exporting restricted to uncookable file formats. the actual conversion is still needed but can be adressed later.

PR #335, opened 2021-02-03
Updated MSBuild.Sdk.Extras & Turned on TreatWarningsAsErrors for all …

phrisk:
One last commit to end all warnings (:

Fixed:
- Updated `MSBuild.Sdk.Extras` -> there was a bug in the version previously referenced that caused targets to be imported twice
- Fixed warning in `animAnimNode_ReferencePoseTerminator`
  - Here `visWhenActive` was hiding a base member - I've added the `new` keyword, assuming the ordinal in the inherited class is correct
- Fixed unreachable code warnings in `ArchiveManager`
- Turned on **TreatWarningsAsErrors for ALL projects**
  - This might be annoying to some, but it's a good thing to have IMO... and will hopefully stop contributors introducing new warnings, whether intentional or not.

Warnings are now completely gone 👍

### DIFF
--- a/CP77.CR2W/Archive/ArchiveManager.cs
+++ b/CP77.CR2W/Archive/ArchiveManager.cs
@@ -121,7 +121,7 @@ namespace CP77.CR2W.Archive
         public override void LoadModArchive(string filename)
         {
             return;
-            if (Archives.ContainsKey(filename))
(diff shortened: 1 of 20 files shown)


Traderain (approved):

Traderain:
Thanks, this is awesome!

PR #344, opened 2021-02-08
Custom Orchestra Libraries + TitleBar redesign.

Offline-R503B:
<See Title>

Implemented:
- Custom Orchestra Libraries. (And renamed my audiocontrols lib) 
- Mahapps window as shell.
- I did some stuff on home page to make it look cleaner.

Fixed:
- Nothing , Absolutely nothing. If it works don't fix it.

<Additional notes>


Offline-R503B:
Ofcourse this isn't just changing a simple change that's why I want this PR to be mentioned and reviewed here so we can later take it as reference if bugs pop up regarding using a Mahapps window. (so far nothing , and I don't expect anything special)

Offline-R503B:
Needs to be merged in a different branch.

PR #352, opened 2021-02-09
Removed homepage visibility.

Offline-R503B:
Removes navigation possibility to homepage.
removed homepage from startup.

View is still there


Offline-R503B:
Unrelevant

PR #356, opened 2021-02-11
Home page progression

Offline-R503B:
Just more stuff , Some1 please get that update dialog outta here it has the possibility of crashing the app ;p


Offline-R503B:
Wrong branch

PR #395, opened 2021-03-11
consolidate tw3 csprojects

rfuzzo:
# consolidate tw3 csprojects

Implemented:
- moved tw3 csprojects to one consolidated project

<Additional notes>


Traderain:
Good idea

PR #433, opened 2021-04-22
Updated Pack.cs removing "basegame_" from output filename

Halvkyrie:
Updated to remove the "basegame_" part of output file naming as it is no longer necessary for loading archives as of game version 1.2 (and it was annoying me)


rfuzzo:
grand!

PR #461, opened 2021-05-17
tw3 manager fixes

rfuzzo:
- fixed a bug where tw3 managers would not load


WolvenKit.Witcher3Formats/W3Speech/SpeechManager.cs
```diff
@@ -54,7 +54,15 @@ public override void LoadAll(string exedir)
                     .SelectMany(dir => Directory.GetFiles(dir ?? "", "*.w3speech", SearchOption.AllDirectories)
                         .OrderBy(k => k)))
                 {
-                    LoadArchive(file);
+                    try
+                    {
+                        LoadArchive(file);
+                    }
+                    catch (Exception e)
+                    {
```
rfuzzo, line 62: @Traderain we are getting constant stream length exceptions here for the DLC. needs investigation

PR #533, opened 2021-07-27
Feature 532 remove orchestra

### LINKED ISSUE #532: Remove Orchestra
Remove Orchestra Shell, since we're not using any of its features anymore: We're using Syncfusion ribbon instead of Fluent and we're using a syncfusion window. And it it is difficult to debug. There seems to be no reason to keep using orchestra.

rfuzzo:
# Feature 532 remove orchestra

Implemented:
- use mainview.xaml instead of shell 
- rebuilt mainview.xaml after shell

Fixed:
- removed some unused nugets

<Additional notes>
closes #532


Offline-R503B (approved):
Taking a well deserved vacation after reading this.

PR #555, opened 2021-08-19
no archive list re collection for mesh exports

HitmanHimself:
Fixed:
- <Fixes>
mesh dependency files haven been moved in archives in patch 1.3 so re collecting certain archives causes missing dependencies, removed re collection


rfuzzo (approved):
ok

PR #574, opened 2021-08-29
mlmask images offset fix

HitmanHimself:
Fixed:
- <Fixes>
mask images in mlmask imports being horizontally off from the original images has been fixed


rfuzzo (approved):
dx10 👍

PR #584, opened 2021-09-06
Fixed error "Value cannot be null. (Parameter 'path1')" when uncooking basegame_4_appearance.archive

gammaparticle:
### Fix error "Value cannot be null. (Parameter 'path1')" when uncooking basegame_4_appearance.archive

Implemented:
- A workaround. Replaced `settings.Get<MorphTargetExportArgs>().ModFolderPath` in call to ExportMorphTargets() in Uncook.cs with the raw output directory (i.e. -o argument of WolvenKit.CLI) when `settings.Get<MorphTargetExportArgs>().ModFolderPath` is null (which occurs when WolvenKit.CLI is used, but not the App).

Fixed:
- The error: "unexpected error occured while uncooking: Value cannot be null. (Parameter 'path1')" which occurs when uncooking 
basegame_4_appearance.archive.


rfuzzo (approved):
approve, at some point we need to add more verbose logging.

HitmanHimself (approved):

PR #585, opened 2021-09-06
Fixed exception on writing null buffers in CArrayBase

rennr:
# Fixed exception when writing null buffers in CArrayBase

Fixed:
- Fixed: CArrayBase caused an exception when writing certain data types (UInt8) if internal buffer is null.


rfuzzo (approved):
thank you for fixing this.

PR #593, opened 2021-09-28
Fixed audio replay bug

Prongt:
Fixed:
- Bug: Clicking the play button on an audio track wouldnt play the track again if it had previously been played


rfuzzo (approved):
thank you!

PR #596, opened 2021-10-03
Feature/tweak records

rfuzzo:
# Feature/tweak records

Implemented:
- Yaml tweakDocuments and records
- Serialize records

<Additional notes>


WolvenKit.RED4.TweakDB/Types/Fundamentals/BaseFundamental.cs
```diff
@@ -11,5 +12,7 @@ public abstract class BaseFundamental<T> : IType
         public abstract void Serialize(BinaryWriter writer);
 
         public override string ToString() => $"{Name}, Value = {Value}";
+
+        public string GetValueString() => Value.ToString();
```
WopsS, line 16: Why not just adapting `ToString`?

WopsS (commented):
In my opinion `Serialization` and similar things should be in WolvenKit, the TweakDB should not be responsible about that.

`TweakRecord` can be renamed to `Record` and add getters (for flats), maybe something like

```csharp
class Record
{
    public string Name { init; get; }
    public string Type { init; get; }
    // etc

    public void Add(...)
    {
        // Add a flat, combine record's name with flat's name.
    }
}
```

Overall, up to you to merge or change.

WopsS (commented):
Seems good, we can do other changes when we don't have anything to do and want a rework 🤣

WopsS (approved):

PR #607, opened 2021-11-03
Adds a .tweak->TweakDB .bin conversion command to the CLI

jackhumbert:
Checks for in & out paths, and assigns the current directory if one isn't assigned. Also adds an option to not delete the output directory, if supplied.


rfuzzo (approved):
lgtm

PR #611, opened 2021-11-09
Allow simultaneous reads from a single archive file by multiple WolvenKit.CLI processes

gammaparticle:
### **Implemented Feature:**

On initial read of an archive file, a memory mapped file (MMF) is checked for existence. If it does not exist, the IOException that's thrown is caught and a MMF is created.  Names are added to MMFs on creation to enable sharing of the mapped files across multiple processes, allowing multiple WolvenKit.CLI processes to simultaneously access the same archive file.  Once MMF handling is completed, the MMF is explicitly disposed.

### **Results:**

Changes allow for  multiple simultaneous unbundling instances on the same archive file, and also allows multiple simultaneous uncooking instances on the same archive file (with each instance writing to its own output location).

### DIFF
--- a/WolvenKit.Modkit/RED4/Unbundle.cs
+++ b/WolvenKit.Modkit/RED4/Unbundle.cs
@@ -31,10 +31,7 @@ namespace WolvenKit.Modkit.RED4
             var extractedList = new ConcurrentBag<string>();
             var failedList = new ConcurrentBag<string>();
 
-            MemoryMappedFile mmf;
-
-            // name of memory mapped file that's derived from the filestream
-            var mapName = "archiveMapFS";
+            var mapName = "ExtractAll_Map";
 
             // check search pattern then regex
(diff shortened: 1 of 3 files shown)


WolvenKit.Modkit/RED4/Unbundle.cs
```diff
@@ -59,7 +64,16 @@ public void ExtractAll(Archive ar, DirectoryInfo outDir, string pattern = "", st
             }
 
             using var fs = new FileStream(ar.ArchiveAbsolutePath, FileMode.Open, FileAccess.ReadWrite, FileShare.ReadWrite);
```
rfuzzo, line 66: Should this be moved into the catch clause, or is that not a problem?

gammaparticle: Yes, that can be moved into the catch clause.  Let me go ahead and move it.

rfuzzo (approved):
great! thank you for the changes.

PR #616, opened 2021-11-15
Adds UInt64 parsing for raRef:CResource in TweakDB compilation

jackhumbert:
Only parses decimal since that's the only format I've ever seen CResources in - please correct me if I'm mistaken!

### DIFF
--- a/WolvenKit.RED4.TweakDB/Types/CResource.cs
+++ b/WolvenKit.RED4.TweakDB/Types/CResource.cs
@@ -15,17 +15,16 @@ namespace WolvenKit.RED4.TweakDB.Types
         public static implicit operator CResource(string value)
         {
             var cres = new CResource();
-            var rx = new Regex(@"(\d+)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
-            var m = rx.Matches(value);
-            if (m.Count == 1)
+            UInt64 key;
+            if (UInt64.TryParse(value, result: out key))
             {
                 cres.Text = value;
-                cres.Key = Convert.ToUInt64(m[0].Groups[1].Value, 10);
-                return cres;
+                cres.Key = key;
+            } else
+            {
+                cres.Text = value;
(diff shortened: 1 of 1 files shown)


WolvenKit.RED4.TweakDB/Types/CResource.cs
```diff
@@ -11,11 +10,21 @@ public class CResource : IPrimitive
...
 
-        public static implicit operator CResource(string value) => new()
+        public static implicit operator CResource(string value)
         {
-            Text = value,
-            Key = (!string.IsNullOrEmpty(value) && value != "0") ? FNV1A64HashAlgorithm.HashString(value) : 0 
-        };
+            var cres = new CResource();
+            UInt64 key;
+            if (UInt64.TryParse(value, result: out key))
```
WopsS, line 17: `ulong.TryParse(value, out var key)` and the declaration above is not needed.

WolvenKit.RED4.TweakDB/Types/CResource.cs
```diff
@@ -11,11 +10,21 @@ public class CResource : IPrimitive
...
+            {
+                cres.Text = value;
+                cres.Key = key;
+            } else
```
WopsS, line 21: `else` on new line.

WopsS (changes requested):

WolvenKit.RED4.TweakDB/Types/CResource.cs
jackhumbert, line 17:
```suggestion
            if (ulong.TryParse(value, out var key))
```

WolvenKit.RED4.TweakDB/Types/CResource.cs
jackhumbert, line 21:
```suggestion
            } 
            else
```

WopsS (approved):

PR #621, opened 2021-12-04
Tweaks to the pack/build process

jackhumbert:
* Only delete .archive files in `cp77Proj.PackedModDirectory` (`archive/pc/mod`) instead of the whole directory, allowing other files to be placed here & packed
* Removes `mod` prefix from .archive filenames*
* Preserve folder names from `cp77Proj.TweakDirectory` when compiling to `cp77Proj.PackedTweakDirectory`

\* This will cause duplicate .archive files if you're zipping up the packed folder & dropping it in - do we have a way to notify people of a change like this?

### DIFF
--- a/WolvenKit.App/Functionality/Controllers/RED4Controller.cs
+++ b/WolvenKit.App/Functionality/Controllers/RED4Controller.cs
@@ -393,8 +393,14 @@ namespace WolvenKit.Functionality.Controllers
             foreach (var f in tweakFiles)
             {
                 var text = File.ReadAllText(f);
+                var folder = Path.GetDirectoryName(Path.GetRelativePath(cp77Proj.TweakDirectory, f));
+                var outDirectory = Path.Combine(cp77Proj.PackedTweakDirectory, folder);
+                if (!Directory.Exists(outDirectory))
+                {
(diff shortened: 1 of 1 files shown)


ja-to:
In regards to breaking changes, I believe the best way to communicate is by staying ahead with patch notes.

Below is an example of a breaking change in a previous release.
https://wiki.redmodding.org/wolvenkit/features/release-notes/8.3/projects

Moving forward, the landing page for each major release (i.e. .../features/release-notes/8.x) should contain a warning with a link to detailed notes about the change.

For example, let's say the next release will be version 8.4.4 (minor release). We can begin writing the notes before the release is published. This way there's something to track so nightly users aren't caught out by duplicate installed mod archives. After release we can add a breaking change hint to the header of the 8.4 page with a very brief summary and link to the specific change below.

We could also try:

- Adding a one-time popup for new installs with information or a link to the wiki
- Adding some webview content or link within the WolvenKit home/landing page
- Sending OTA in-app notification (system for this is already integrated I believe)
 
 I believe getting ahead with patch notes is probably the most efficient/simple solution.

rfuzzo (approved):

PR #638, opened 2022-01-09
Remove "SelfContained=False" from publish workflows

ADawesomeguy:
This prevents libraries needed for WINE to function properly from being excluded in the build/

Redid the PR because I accidentally used the master branch for the other one.


rfuzzo (approved):
CI

PR #663, opened 2022-02-04
8.5.0 bug fixes

seberoth:
Changed:
- Texture Import: Default value for "keep file" changed from true to false

Fixed:
- Morphtarget export


rfuzzo (approved):
ok

PR #666, opened 2022-02-08
fix a bug where Oodle would crash on load

rfuzzo:
# yes

Fixed:
- fix a bug where Oodle would crash on load

<Additional notes>
linus


ADawesomeguy:
# linus
<sub>looks good to me</sub>

ADawesomeguy (approved):
Byootifull

PR #667, opened 2022-02-12
Add Timer to Check Folder Path Validity in ProjectWizard

WRodewald:
Added a timer that checks whether the folder path is valid every 1s. This is necessary since the project wizard will otherwise not notice when a folder path has been made valid since it last time was checked. 

Example:
appending a (non-existing) sub-folder in the text-field will cause the wizard to correctly note that the folder doesn't exist and mark the path as invalid. When opening the browser through the button and adding that folder, in the browse UI, the path won't change, thus no check will be made, even though the folder now exists and the path should be valid. 

I have no experience with C# so I'd appreciate any feedback on the changes.

### DIFF
--- a/WolvenKit/Views/Wizards/ProjectWizardView.xaml.cs
+++ b/WolvenKit/Views/Wizards/ProjectWizardView.xaml.cs
@@ -6,6 +6,7 @@ namespace WolvenKit.Views.Wizards
 {
     public partial class ProjectWizardView : ReactiveUserControl<ProjectWizardViewModel>
     {
+        private System.Windows.Threading.DispatcherTimer timer;
 
         public ProjectWizardView()
         {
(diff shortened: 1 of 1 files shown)


WolvenKit/Views/Wizards/ProjectWizardView.xaml.cs
```diff
@@ -1,11 +1,13 @@
 using System.Reactive.Disposables;
 using ReactiveUI;
 using WolvenKit.ViewModels.Wizards;
+using System.Windows.Forms;
```
rfuzzo, line 4: Could you use the WPF variant of the timer?
This one:
https://docs.microsoft.com/en-us/dotnet/api/system.windows.threading.dispatchertimer?view=windowsdesktop-6.0

rfuzzo (changes requested):
before accept, please remove Windows.Forms dependency

rfuzzo (approved):

PR #673, opened 2022-02-18
Added ScrollViewer to prevent overflow in CreateClassDialog

deadlymustard:
Added ScrollViewer to prevent overflow in CreateClassDialog

Fixed:
- Fixed an issue where the Create Class dialog would be longer than screen and not allow the user to interact with the panel
- Added a default height of `800` to the panel
- Wrapped the main panel content in a `ScrollViewer` so that a large amount of content would scroll the screen rather than expanding infinitely

Before:
Capture

After:
Screenshot 2022-02-17 232045


rfuzzo (approved):
Great stuff! thank you

PR #681, opened 2022-02-19
Add Base/Mod Files Toggle to Asset Browser

effs:
Implemented:
- Toggle button to switch between the game base archives and installed mod archives. This was previously only available in the ribbon.

Fixed:
- The 'home' button in the Asset Browser was a little confusing, changed it to a FoldUp icon and a 'back to top level' tooltip

Didn't remove the existing toggle from the ribbon view yet.


rfuzzo (approved):
lgtm

PR #688, opened 2022-02-27
Check build only on branch pushes and not tags

ADawesomeguy:
Also I got rid of workflow dispatch for checks. Feel free to add it back.


rfuzzo (approved):
I review

PR #693, opened 2022-03-01
Fix warnings on publish

ADawesomeguy:
tbh no clue what I'm doing but I looked at the compiler warnings and did what they said to do to fix them because they were an eyesore.

also for some reason rider did some weird stuff with whitespace but it seems to look ok.

### DIFF
--- a/WolvenKit.RED4.Types/ClassesExt/Properties/cpGameplayLightComponent.cs
+++ b/WolvenKit.RED4.Types/ClassesExt/Properties/cpGameplayLightComponent.cs
@@ -7,7 +7,7 @@ namespace WolvenKit.RED4.Types
         public new CHandle<entIPlacedComponentEditorData> PlacedEditorData
         {
             get => GetPropertyValue<CHandle<entIPlacedComponentEditorData>>();
-            set => SetPropertyValue(value);
+            set => SetPropertyValue<CHandle<entIPlacedComponentEditorData>>(value);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


WolvenKit.RED4.Types/ClassesExt/Properties/cpGameplayLightComponent.cs
```diff
@@ -4,10 +4,10 @@ public partial class cpGameplayLightComponent
     {
         [OrdinalOverride(Before = 9)]
         [RED("placedEditorData")]
-        public CHandle<entIPlacedComponentEditorData> PlacedEditorData
+        public new CHandle<entIPlacedComponentEditorData> PlacedEditorData
         {
             get => GetPropertyValue<CHandle<entIPlacedComponentEditorData>>();
-            set => SetPropertyValue<CHandle<entIPlacedComponentEditorData>>(value);
+            set => SetPropertyValue(value);
```
seberoth, line 10: Even if it isn't needed, please revert so it matches the rest again :)

seberoth (changes requested):

WolvenKit.RED4.Types/ClassesExt/Properties/cpGameplayLightComponent.cs
ADawesomeguy: Hm I don't remember doing that. Might've been some automatic fix by Rider or something I clicked. I'll make sure to revert it.

seberoth (approved):
Looks good to me

ADawesomeguy:
@rfuzzo mergeable?

PR #695, opened 2022-03-01
Fix nuget publish action

ADawesomeguy:
This is an attempt to fix the nuget action -- it ran on my fork without failing or complaining so maybe it'll work in the main repo?


rfuzzo (approved):
Ah good idea

PR #704, opened 2022-03-05
Add Launch Game button to main toolbar

### LINKED ISSUE #702: Add Launch Game Option/Button
It'd be nice to be able to launch the game directly from WK, either as an option to pack+install or a separate command.

A button used to exist for this at some point.

effs:
Implemented:
- Launch Game button in the main toolbar (Closes #702)

Notes:
- Just launches the game. Could have it pack & install automatically, but I suspect quite often folks will want to do or simply do that separately. So it's two clicks rather than possibly rebuilding the mod when you don't want to.
- No implementation for the Ribbon (which is to be retired)

### DIFF
--- a/WolvenKit.App/Services/ISettingsManager.cs
+++ b/WolvenKit.App/Services/ISettingsManager.cs
@@ -58,6 +58,8 @@ namespace WolvenKit.Functionality.Services
 
         string GetW3GameRootDir();
 
+        public string GetRED4GameExecutablePath();
+
         string GetRED4GameRootDir();
 
         public string GetRED4GameModDir();
--- a/WolvenKit.App/Services/SettingsManager.cs
+++ b/WolvenKit.App/Services/SettingsManager.cs
@@ -169,6 +169,8 @@ namespace WolvenKit.Functionality.Services
             return fi.Directory is { Parent: { Parent: { } } } ? Path.Combine(fi.Directory.Parent.Parent.FullName) : null;
         }
 
+        public string GetRED4GameExecutablePath() => CP77ExecutablePath;
+
         public string GetRED4GameModDir()
         {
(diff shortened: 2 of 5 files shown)


effs:
image

rfuzzo:
cool, thank you! Since the pack & launch etc options are getting more and more, I think we should have a drop-down button with all: 
- Pack, 
- Pack & Install, 
- Pack, Install & Launch, 
- Launch

rfuzzo (approved):
ye

ADawesomeguy:
If there are applicable launch parameters (e.g. `-debugscripts` for TW3) maybe it would be a good idea to add those and hook onto the log or something like that?

effs:
> cool, thank you! Since the pack & launch etc options are getting more and more, I think we should have a drop-down button with all:
> 
>     * Pack,
> 
>     * Pack & Install,
> 
>     * Pack, Install & Launch,
> 
>     * Launch

Good idea — there’s the related workflow enhancement of creating an installable mod (aka zipping the pack) and publishing it, but that might not fit neatly into this set so it can be a separate option.

The only question is really which one is the most common workflow.

ADawesomeguy:
Like this? 
image

rfuzzo:
Yeah like in that picture!

ADawesomeguy:
Déjà vu 👀

ADawesomeguy:
```
unattended
forceRawTweakDB
profiler
windowCaption
scriptVersion
tweakdbBlobPath
scriptsBlobPath
pvdDumpToFile
renderPreset
forceLowSettings
qualityLevel (RTX)
texQualityLevel
meshQualityLevel
interopstartingport
watchdogTimeout
automator
```
List of startup params, maybe we could make a checkbox section of these in settings that the user can check for debug launches.

effs:
I’ll create a separate Issue + PR for adding the Launch Menu and the command-line switches, great ideas!

PR #710, opened 2022-03-08
Add libs to main directory and reference them by name and not path

ADawesomeguy:
The whole lib subdirectoy was a problematic so I think it's better to reference them by name in the main directory as it's not platform-specific.

Also fixes the issue where it tries to load a Linux lib on Windows.


rfuzzo (approved):
I trust you that it works on Linux.

ADawesomeguy:
I trust you that you checked.

ADawesomeguy:
> I trust you that it works on Linux. 

I'll test it out again before you merge.

ADawesomeguy:
Ok it worked, maybe test on Windows. Also keep in mind this will move all the libs to the main build directory instead of the lib subdirectory but is infinitely more convenient for P/Invoke.

rfuzzo:
Ok

ADawesomeguy:
Sooo.... does that mean this should be merged?

PR #711, opened 2022-03-09
Export textures referenced in .mltemplate files

dragonzkiller:
Currently, when exporting a mesh with materials into an empty depot, materials referenced in an mltemplate file are not exported along with the json. This leads to unreferenced textures, for example, when importing an mltemplate texture into Blender.

This pull request finds all CBitmapTextures (or rather their CResourceReferences) and exports the associated file if not previously done so.


rfuzzo (approved):
lgtm

Issue #717, opened 2022-03-09
Improve Developer Docs For Getting Started, Debugging, Etc.


effs:
There’s already a bunch but we could try to compile it a bit more easily accessible — just an .md file or two in the repo itself could do the trick and have a better chance to be up to date than even the wiki (to which docs could be auto-committed?)

- [ ] Basic dev guide in repo
- [ ] Not-basic stuff
- [ ] Push to/pull into the wiki automatically?

PR #719, opened 2022-03-10
Add Barebones Dev Guide

effs:
Should figure out how to get this synced directly to the wiki.

Relating to #717


rfuzzo (approved):
omg thank you. these docs are so useful

Issue #718, opened 2022-03-10
Add Launch Options And Improve Toolbar Pack/Install/Launch Features


effs:
Continuing off #702/#704 as this requires some reworking of the toolbar and the launch options are a major addition.

## Toolbar

Simple implementation looks like this:

![image](https://user-images.githubusercontent.com/13802421/157657192-b2eb1258-1954-4c59-9a4b-f300e7f3bf03.png)

…But it needs a better control than the SplitButton implementation in sf (weirdly uncustomizable as you can see with the black text) or MahApps (doesn’t actually do what splitbuttons do). I’ll see if it’s possible to import the WinUI3 SplitButton as an XAML Island… or just write a very simple Button + Menu with some cross-triggers.

Leastwise, I would prefer to have the most-used option directly clickable with the dropdown for additional options. (Could even make the most-used dynamic, or last-used?) If anyone has better suggestions, more than welcome!

## Launch options

@ADawesomeguy compiled a list of the command-line options:
```
unattended
forceRawTweakDB
profiler
windowCaption
scriptVersion
tweakdbBlobPath
scriptsBlobPath
pvdDumpToFile
renderPreset
forceLowSettings
qualityLevel (RTX)
texQualityLevel
meshQualityLevel
interopstartingport
watchdogTimeout
automator
```

This needs:

- [ ] Support in Settings to either add one set of switches, or
- [x] Actual 'launch profile' support (named + opt set), and both need
- [ ] Validation of the selections (and could do some more validation on settings save/load too)

Additionally if using launch profiles

- [ ] UI to select the profile to use (maybe default with a _secondary_ sub-action dropdown?)

jackhumbert:
Here's some additional ones I was able to find. The `-` is necessary, and I've only tested a handful of them:

```
-engine=
  serverServerGameEngine
  serverDebugServerGameEngine
  FunctionalTestsServerEngine
  HeadfullTestServerEngine
  HeadllessTestServerEngine
  BackendGameEngine
-multiplayerServer
-multiplayerClient
-ignoreInput
-ignoreGamepads
-hudNoParallel
-noInputIcons
-newInkLayersEnabled
-mpMenu
-noFullscreenUI
-noHUD
-forceLoadAllMeshAppearances
-breakOnPoolDefault
-profileio
-disableioopt=
  IOLRUCache
  BufferInlineData
  MergeOps
  CancelOps
  KeepAliveForGameCache
-autoMemDump
-memoryDumpOuput
[...]

PR #735, opened 2022-03-16
Added function/button to launch CP77 via Steam

Halvkyrie:
Adds a function and a button to the shell next to the existing "Launch Game" button that runs the default steam id for Cyberpunk 2077. Not fully added to the ribbon view however as i don't know how to do that. Seemingly "RibbonView.g.i.cs" would have to be generated again.
I hardly know what I'm doing so please be careful with this (and someone would have to add it to the ribbon view properly)

### DIFF
--- a/WolvenKit.App/ViewModels/Shell/AppViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/AppViewModel.cs
@@ -139,6 +139,7 @@ namespace WolvenKit.ViewModels.Shell
             ShowSettingsCommand = new RelayCommand(ExecuteShowSettings, CanShowSettings);
 
             LaunchGameCommand = new RelayCommand(ExecuteLaunchGame, CanLaunchGame);
+            LaunchSteamGameCommand = new RelayCommand(ExecuteLaunchSteamGame, CanLaunchSteamGame);
 
             CloseModalCommand = new RelayCommand(ExecuteCloseModal, CanCloseModal);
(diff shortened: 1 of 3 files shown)


ADawesomeguy:
I have no clue what to do with this so I requested some people that are smart and/or have worked with this or are good at GUI and are also cool people 👍🏽

effs (commented):
Nice! Yes,  Steam support would be nice - dunno if other managers need specific support too?

To avoid overcrowding the toolbar, I think the GUI element could be a part of #718?

What if at least for now we'd actually add a static launch command + flags to Settings instead? It'd also allow adding command-line flags. Then rather than use the exe we can always launch the game with that. Most likely people will only use one launcher, yes?

WolvenKit.App/ViewModels/Shell/AppViewModel.cs
```diff
@@ -510,6 +511,29 @@ private void ExecuteLaunchGame()
...
+                {
+                    FileName = steamrunid,
+                    ErrorDialog = true,
+                    UseShellExecute = true
+                });
+            }
+            catch (Exception ex)
+            {
+                _loggerService.Error("Launch: Error! Please check if you have Steam installed, and a valid Steam installation of Cyberpunk 2077");
+                _loggerService.Info($"Launch: error debug info: {ex.Message}");
```
effs, line 532: 💎

effs:
This was a really good spark, @Halvkyrie - if #748 works for this use case, I think we can close this one and then try to figure out what the optimal UI would be in #718?

rfuzzo (approved):
steam

PR #748, opened 2022-03-21
LaunchCommand/LaunchOptions In Settings, Used By Launch Button (Plus Versioned Settings Support..)

### LINKED ISSUE #741: Launchcommand + Flags In Settings
A quicker measure ahead of #718.

Full-fledged support will be a bigger effort, so maybe for right now we can add two fields to Settings:

1. [x] `LaunchCommand` is the exe or other command to run to launch the game, e.g. Steam's launcher (`"steam://rungameid/1091500"` from #735 ). *Defaults to the current exe field*
2. [x] `LaunchOptions` is a static string of command-line options to pass to the `LaunchCommand`. *Defaults to empty*

With these, we can support 'launch configurations' even if it's static. All the elements are already in place, so this is a quicker change.

Additional stuff:

- [x] FirstSetupWizardView shouldn't depend on anything in SettingsManager. If it's not healthy, we are not gonna assume _any_ of the data in the settings is valid anymore. So start from a blank settings file.
- [ ] SettingsManager.IsHealthy() may need a bit of tweaking
- [ ] TESTS

effs:
New stuffs:

- SettingsDto is versioned; the idea is to keep the Dto always on the current version but maintain a set of migrations that are always run on load/save. When eventually there's a v3 settings, then for any v2 settings we should run MigrateV2ToV3 (which does whatever fixes) and for any undefined/v1 settings it's MigrateV1ToV2, then MigrateV2ToV3.
- v2 settings include LaunchCommand and LaunchOptions to support the thing I was ACTUALLY doing, which is enabling using alternative launch stuffs for the Launch Button (as req'd in #735 and was part of #718)

Fixed stuffs:

- TweakDB loading could crash on startup on first run because settings might not exist at that point yet

Pictures:

- Initial setup image

- With options (don't recommend this particular one) image

Closes #741

### DIFF
--- a/WolvenKit.App/Services/ISettingsManager.cs
+++ b/WolvenKit.App/Services/ISettingsManager.cs
@@ -10,45 +10,46 @@ namespace WolvenKit.Functionality.Services
     {
         #region Properties
 
-        bool CheckForUpdates { get; set; }
+        // I wish there was a better way to inject deps
+        // than creating unnecessary interfaces.
+        public string SettingsVersion { get; }
+        public bool CheckForUpdates { get; set; }
+        public EUpdateChannel UpdateChannel { get; set; }
+        public bool ShowGuidedTour { get; set; }
+        public string ThemeAccentString { get; set; }
+        public string CP77GameDirPath { get; set; }
+        public string CP77ExecutablePath { get; set; }
+        public string CP77LaunchCommand { get; set; }
+        public string CP77LaunchOptions { get; set; }
+        public bool ShowFilePreview { get; set; }
+        public string ReddbHash { get; set; }
+        public string MaterialRepositoryPath { get; set; }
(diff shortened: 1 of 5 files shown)


ADawesomeguy:
I don't know what this means but it looks cool so I'll approve when I'm at my PC.

effs:
> I don't know what this means but it looks cool so I'll approve when I'm at my PC.

💜

ADawesomeguy:
Oh wait I understand more from the title, good PR imo.

effs:
@Halvkyrie D' you want to see if this works for you on Steam?

ADawesomeguy (approved):
There is a lot of code 👍🏽

rfuzzo (approved):
thank you

PR #759, opened 2022-03-31
Search Enhancements

effs:
Generally:

- Search syntax unified a bit, improved (?) syntax and filtering (see below)
- Search game files or mods based on mod toggle in Asset Browser
- Removed secondary search bar, there's just one
- Basic error handling + regexp timeouts

Search syntax:

- Search is a sequential filter split on ` > `, i.e. you can do `judy > quest` to _first_ match `judy` and then in those results `quest`. (Alternative would be `quest judy` because quest is first in the path, but it might be more ergonomic to just tack on further refinements rather than modify the earlier.)
- Spaces to split 'terms', anything allowed to match between terms. I.e. `quest judy` => `quest.*?judy`.
- Partial match anything anywhere. Term `ma` will match a whole lot, e.g. `../market/.../..`, `../machete.xbm` and whatever.
- File extension matching because it's just strings? `.ent` will match, also mid-path, but who does that?
- OR logic like `.app|.ent` (no spaces allowed, currently)
- Exclusion like `judy !shoes .app|.ent|.mesh` or `judy .app|.ent|.mesh > !shoes` if you want to match out of order. Works on extensions too since, again, they're just strings
- You can probably still write most regexps, just don't leave whitespace in them

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -260,6 +260,9 @@ paket-files/
 .idea/
 *.sln.iml
 
+# VSCode
+.vscode
+
 #macOS
 .DS_Store
 
--- a/WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
@@ -548,21 +548,20 @@ namespace WolvenKit.ViewModels.Tools
                         _ => searchRefinement
                     })
                     .Select(AsMatchFunctions)
-                    .Select(archive =>
-                    archive)
                     .ToArray();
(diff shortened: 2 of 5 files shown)


effs:
The code is kind of stupid but primarily I want to find out if this improves search ergonomics for the user.

rfuzzo (approved):
lgtm, at some point we might need some in-app help tooltips :D I need to add that to the curve editor too

effs:
@rfuzzo Yeah, should be liberal with them. I'll see about creating a template for it for search, hopefully reusable then.

PR #760, opened 2022-04-02
Archive refactor

seberoth:
Archive refactor

Implemented:
- Moved archive reading to RED4.Archive
- Guess file type for unknown hashes

Fixed:
- Permission issue while reading archive files (UAC)
- If multiple version of a file exists, the one from the last loaded archive is used


rfuzzo (approved):
Thanks

PR #761, opened 2022-04-04
Fix/archive load order

seberoth:
Archive fixes

Implemented:
- Proper load order for archives for now
- Added descriptions to file types
- Added some new filetypes (commented out if they aren't used currently)

Fixed:
- Crash while opening not supported xbms


rfuzzo (approved):
nice

PR #763, opened 2022-04-06
Hotfix For Mod Browser Read Error Crash

effs:
Quick guard so it won't crash on activating the mod browser. Underlying error is still unresolved.


effs:
Errors in #764 (I have an access denied and a buffer size error)

seberoth (approved):
Looks good to me

PR #772, opened 2022-04-18
Update Shields.io workflow badges in README.md

ADawesomeguy:
The checks always said pending so I just had it use the workflow for build checks.


rfuzzo (approved):
thank you

PR #780, opened 2022-04-24
Fix file path separator for Linux during uncook

dragonzkiller:
Fixes the path separators for Linux when performing an uncook operation with WolvenKit.CLI. Without the change, single files are created with the literal name of the relative path in the archives, rather than creating a directory structure.

Confirmed to still work on Windows as well.

### DIFF
--- a/WolvenKit.Modkit/RED4/Uncook.cs
+++ b/WolvenKit.Modkit/RED4/Uncook.cs
@@ -57,7 +57,7 @@ namespace WolvenKit.Modkit.RED4
                     relFileFullName += ".bin";
                 }
 
-                var mainFileInfo = new FileInfo(Path.Combine(outDir.FullName, $"{relFileFullName}"));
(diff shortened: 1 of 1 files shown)


ADawesomeguy:
The thing is that I could’ve sworn I did this earlier. I have no clue where it went though.

njibhu:
> The thing is that I could’ve sworn I did this earlier. I have no clue where it went though.

This was already fixed for the unbundle command, but not for the uncook one.

Thanks for making the PR @dragonzkiller

ADawesomeguy:
> > The thing is that I could’ve sworn I did this earlier. I have no clue where it went though.
> 
> This was already fixed for the unbundle command, but not for the uncook one.
> 
> Thanks for making the PR @dragonzkiller
> 
> Tested the fix working on Linux

Aha, that makes sense. Thanks!

PR #791, opened 2022-05-06
Sandbox2

someaka:
# Streamingsectors things mostly

<PULL REQUEST TITLE>

Implemented:
- refresh button for streamingsectors
- delete selection in treeview for streamingsectors (still called Test Items in Selection for now, needs a better way to get all SelectedItems)
- import from JSON for streamingsectors : the JSON formats are keanu's who made Object Spawner and contributed some code to this sandbox too
- Imports for .rig files and meshes with bindmatrices

Fixed:
- added a bunch of try catch and is not null all over the place :D

<Additional notes>
probably forgot things


seberoth (changes requested):
Without diving to deep into it:
1. Commits without messages aren't that good
2. Why did you revert the Classes changes?

someaka:
if anything got reverted it's very probably a mistake,
and yes i should probably find a way to make those fruit commits private somehow,
also should probably find a way to cancel that PR the bot is kinda spamming about it
and use the cherry pick function maybe :D i'm sorry i'm not quite sure

PR #792, opened 2022-05-13
Fixed using wrong editor for float values

seberoth:
Fixed:
- Fixed using wrong editor for float values 

Changed format string for floats from "R" to "G9"


rfuzzo (approved):
lgtm G9

PR #807, opened 2022-06-05
🐛 Asset Search Requires Type Prefix For Hash Search

### LINKED ISSUE #804: Asset Browser Search: Require hash: To Search For Hashes
A string of numbers can also match .wem files, so let's actually require the `hash:` type indicator rather than force specifying .wem (it's possible there will be other cases where long number strings occur). Should need hash searches less anyway.

effs:
Hash search requires `hash:` prefix like `hash:123455633`. Otherwise numbers are handled like any other substring.

Closes #804


rfuzzo (approved):
.Select.Select.Select

PR #799, opened 2022-05-28
Clean branch

someaka:
# StreamingSectors and stuff

<PULL REQUEST TITLE>

Implemented:
- added Copy Paste and Delete selection
- added export for nodeData
- added imports for Objects Spawner and Unreal formats
- added option to import gltf with no original (WIP)
- added an option to import boneRigMatrices for meshes
- added an option to import rigs (very very WIP)
- added a refresh file button for streamingsectors
- probably more stuff i forgot

Fixed:
- YES

<Additional notes>

🍌 🍐 🍓 🍎 🍏 🍒 🍑

PR #810, opened 2022-06-05
Random fixes

seberoth:
Fixed:
- Hashing for CName and NodeRefs (again...)
- Duplicated fileName writing in CR2W
- Wrongly shifted size field in RedPackageNameHeader
- Enum writing for enums where value is "None"
- A bug introduced while merging #799 
- Moved buffer generation, so the newly generated bytes are written
- Removed showing success message on import/export when the action failed
- Prevent empty soundbank creation while processing all imports/exports
- Filter out "old" files from rig selection


rfuzzo (approved):
👍

PR #821, opened 2022-06-09
Wwise file support

ADawesomeguy:
Implemented:
- [x] Command for converting WEM to OGG in CLI

GUI is too complicated for me so I'm gonna push this and someone else can add it to the GUI if needed.


WolvenKit.Modkit/RED4/Tasks/WwiseTask.cs
```diff
@@ -0,0 +1,47 @@
...
+                outpath = Path.ChangeExtension(path, ".ogg");
+            }
+
+            if (wem)
+            {
+                var file = File.ReadAllBytes(path);
+                using var ms = new MemoryStream(file);
+                using var br = new BinaryReader(ms);
+
+                var inBuffer = br.ReadBytes(file.Length);
```
rfuzzo, line 28: As far as I can see you just need `var inbuffer = File.ReadAllBytes(path);` here. No need for streams if you just use the byte array in the end right?

WolvenKit.Modkit/RED4/Tasks/WwiseTask.cs
```diff
@@ -0,0 +1,47 @@
+using System.IO;
+using WolvenKit.Core.Wwise;
+
+namespace CP77Tools.Tasks
+{
+    public partial class ConsoleFunctions
+    {
+        #region Methods
```
rfuzzo, line 8: No need for regions if it's just a small file and has no other regions.

WolvenKit.Modkit/RED4/Tasks/WwiseTask.cs
```diff
@@ -0,0 +1,47 @@
...
+
+                File.WriteAllBytes(outpath, oggBuffer);
+
+                _loggerService.Success($"Finished converting {path} to OGG: {outpath}");
+            }
+
+            return 1;
+        }
+
+        #endregion Methods
```
rfuzzo, line 45: No need for regions if it's just a small file and has no other regions.

rfuzzo (changes requested):
Looks good, just the comment about the streams :)

WolvenKit.Modkit/RED4/Tasks/WwiseTask.cs
ADawesomeguy: Ah yeah sorry, I had that in my test but for this file I'd copied the Kraken stuff. I'll fix it ASAP.

rfuzzo (approved):
Lgtm

PR #831, opened 2022-06-12
Removed magic hash from GeometryCacheService

flibX0r:
# Tiny Cleanup

Fixed:
- Replaced magic hash value with known path string in GeometryCacheService


flibX0r:
Tiny PR, saw a magic number while learning the current source, it upset me deeply

jackhumbert (approved):

PR #833, opened 2022-06-12
CLI: Add ability to directly export to JSON from the archives

njibhu:
# Description

There's no direct way to export to json from the CLI. The current flow would be unbundle the file, then use the cr2w command to transform the exported file to json. This is rather cumbersome when doing batch exports. This PR adds a way to directly extract to JSON from the CLI.

# Details

- This PR adds a new exposed CLI flag to the `uncook` command: `-s` or `--serialize`

- Using this flag adds a new step in the uncook process which exports the file to json with the `.json` file extension appended to the end of its path.

- To prevent errors with unsupported files to be uncooked, when the serialize option is used, it does not try to uncook the buffers.

# Notes

For now this argument only serializes the main chunk, but I plan to implement in further PRs serializing the other chunks for embedded resources. But I'm not sure yet on how to expose this to the CLI.


rfuzzo (approved):
lgtm

PR #845, opened 2022-06-27
Turn file preview on by default

ja-to:
File preview is now set true by default aiming to help new users.


ADawesomeguy (commented):
`= true;`

seberoth (approved):
lgtm

PR #846, opened 2022-06-29
fix a bug where certain ui commands would not work

rfuzzo:
Fixed:
- add missing ObservesProperty to DelegateCommands

<Additional notes>
I'm sorry


seberoth (approved):
lgtm

Could replace `AsyncCommand` with `DelegateCommand` and remove `AsyncAwaitBestPractices` but also can do that later

ADawesomeguy:
> can do that later™️

rfuzzo:
> lgtm
> 
> Could replace `AsyncCommand` with `DelegateCommand` and remove `AsyncAwaitBestPractices` but also can do that later

I'll push a fix for it here now. Thanks

PR #848, opened 2022-07-01
🐛 InsertChild Checks Item Can Be Assigned To Array, Not The Other Way Around

### LINKED ISSUE #847: Array Item Duplicate/Copy/Etc. Doesn’t Work For All Arrays
Works: array:meshMeshAppearance, array:CMaterialEntry
Doesn't: array: IMaterial, array:entlComponent

effs:
At a glance looked like the rest of the file was using the correct logic but might be worth a check.

closes #847


rfuzzo (approved):
Effs seems also to be fine

PR #863, opened 2022-07-09
move animation buffer read to where it's actually needed

rfuzzo:
# move animation buffer read to where it's actually needed


rfuzzo:
is this still needed?

seberoth:
Its one of the parsers I always comment out if I'm running the tests because it takes so long...
So I will say yes xD

PR #866, opened 2022-07-16
Feature/msix package

rfuzzo:
# msix package and connected refactors

Implemented:
- msix package project
- removed gpm dependencies
- added wolvenkit resources and move reddb there


seberoth (approved):
Don't know if it works, but we always blame fuzzo if something doesn't work so it doesn't matter

PR #867, opened 2022-07-17
Fix WKitJsonVersion Check

dragonzkiller:
If a JSON file was imported from an OLDER version of wkit, this would fail. The check should be looking for larger version numbers, not smaller.

Fixed:
- version check for WKit exported JSON files


rfuzzo (approved):
Pr goes brrrr

PR #869, opened 2022-07-17
Fix for TweakXL

ja-to:
**Fix for TweakXL**

Implemented:
- WolvenKit copies .yaml format files to the packed directory
- Supports TweakXL pipeline, removes support for deprecated TweakEXT


WolvenKit.App/Controllers/RED4Controller.cs
```diff
@@ -364,47 +364,49 @@ private void CompileTweakFiles(Cp77Project cp77Proj)
...
-                    using var writer = new TweakDBWriter(ms);
-                    writer.WriteFile(db);
-                    File.WriteAllBytes(outPath, ms.ToArray());
-                }
-                catch (Exception e)
-                {
-                    _loggerService.Error(e);
-                    continue;
-                }
+                File.Copy(f, outPath, true);
```
effs, line 379: Are errors handled upstream somewhere so that the user knows this is where it fails if it does?

effs (approved):
⚰️ the commented code, and maybe change the method name to reflect that we’re (processing and) copying the file?

PR #872, opened 2022-07-20
Fix GPU leak

ja-to:
- Commented out several animated icons causing a GPU leak

- Small improvement to Properties mesh preview:
    Adjusted lighting
    Added reset camera context button


rfuzzo (approved):
thanks!

PR #878, opened 2022-07-26
♻️ Use types instead of inline conditionals in AB search

effs:
Moved the new regex and verbatim logic into refinements so that we don't need to deal with it inline.

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
@@ -505,137 +505,115 @@ namespace WolvenKit.ViewModels.Tools
             _progressService.IsIndeterminate = false;
         }
 
+        // Asset search impl
+
+        // Safer regexps
+        private static readonly RegexOptions RegexpOpts = RegexOptions.Compiled | RegexOptions.IgnoreCase;
+        private static readonly TimeSpan RegexpSafetyTimeout = TimeSpan.FromSeconds(60);
+
         private enum TermType
         {
             Unknown,
             Include,
             Exclude,
-            Hash,
-            Verbatim
         }
 
         private readonly record struct Term(TermType Type, string Pattern, string NegationPattern);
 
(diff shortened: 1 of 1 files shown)


WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
```diff
@@ -505,138 +505,116 @@ public async Task PerformSearch(string query)
...
         private static readonly Regex Whitespace = new("\\s+", RegexpOpts, RegexpSafetyTimeout);
         private static readonly Regex Or = new("\\|", RegexpOpts, RegexpSafetyTimeout);
         private static readonly Regex Negation = new("^(?'Open'\\(\\?:)*\\!(?<term>.+?)(?'Close-Open'\\))*$", RegexpOpts, RegexpSafetyTimeout);
         private static readonly Regex SquashExtraWilds = new("((\\(\\?:)?\\.\\*\\??\\)?){2,}", RegexpOpts, RegexpSafetyTimeout);
 
-        private static readonly Func<Term, Term> HonorFileExtension = (Term term) => term.Type == TermType.Verbatim ? term : term with { Pattern = Regex.Replace(term.Pattern, "(^|\\W)\\.(?<term>\\w+?)", "$1\\.${term}") };
-
-        private static readonly Func<Term, Term> DropUnnecessaryGlobStars = (Term term) => term;
-
-        private static readonly Func<Term, Term> LimitOrToOneTerm = (Term term) => term.Type == TermType.Verbatim ? term : Or.IsMatch(term.Pattern) ? term with { Pattern = $"(?:{term.Pattern})" } : term;
```
effs, line 543: These were really hard to read, dunno if the new editorconfig allows simpler formatting for multi-conditional expressions? (I simplified the expressions anyway in this case)

rfuzzo: yeah I changed the editorconfig so that it doesn't auto-change everything into one big line :)

PR #880, opened 2022-08-06
🐛 Tweak Browser: explicitly notify on property change & force refresh…

effs:
Not sure this is the right fix, but the problem is here (since this _does_ fix it).

It's possible this would be better addressed by managing when the load is called (it's called twice on startup if the game path needs to be set), or then there's something funky with how the CollectionViewSource works but this is about 7 levels of bindings too deep for me.


effs:
And rubber-ducking suggests that because the tweak load is run twice, wpf only binds to the first one but we're making changes to the second.. so this works by explicitly telling wpf to check the property's current value. Maybe? In that case the right fix would be to not load twice (but on the other hand this fix doesn't do anything catastrophically wrong.)

rfuzzo (approved):
INPC go brrrr

PR #894, opened 2022-09-08
Temporarily Fix RedMod not being detected

justarandomguyintheinternet:
- Check for the redmod executable instead of the non-existent version.txt file
- Temporary fix until the version.txt file and the remote version exists


rfuzzo:
That's a good idea. But I do have a version.txt file with the DLC... do you not? Which platform are you on?

justarandomguyintheinternet:
Using the GoG version, no version.txt file for me

rfuzzo:
Hmm... okay, that's weird thanks

rfuzzo (approved):
thanks

PR #897, opened 2022-09-13
Fix/menubar

rfuzzo:
# menubar fixes

Implemented:
- Menubar inside titlebar
- bugfixes

<Additional notes>


WolvenKit.App/Controllers/MockGameController.cs
```diff
@@ -24,35 +21,25 @@ public MockGameController()
 
         public IObservable<IChangeSet<RedFileSystemModel>> ConnectHierarchy() => _rootCache.Connect();
 
-        public List<IArchiveManager> GetArchiveManagers(bool loadmods) => new List<IArchiveManager>();
+        public List<IArchiveManager> GetArchiveManagers(bool loadmods) => new();
```
WopsS, line 24: I know it is not related to this class or your change, but why does it have `loadmods` parameter? I find it strange that archive manager should do that.

WolvenKit.App/ViewModels/Documents/ScriptDocumentViewModel.cs
```diff
@@ -56,13 +56,13 @@ public override async Task<bool> OpenFileAsync(string path)
         {
             _isInitialized = false;
 
-            LoadDocument(path);
+            await Task.Run(() => LoadDocument(path));
```
WopsS, line 59: Why this change?

WolvenKit.App/ViewModels/Tools/ImportExportViewModel.cs
```diff
@@ -619,7 +619,7 @@ private async Task ExecuteProcessAll()
                 var toBeExported = ExportableItems.ToList();
                 foreach (var item in toBeExported)
                 {
-                    success = await ExportSingle(item);
+                    success = await Task.Run(() => ExportSingle(item));
```
WopsS, line 622: Why `Task.Run`?

WolvenKit.Common/Tools/ProcessUtil.cs
```diff
@@ -13,7 +13,7 @@ public static class ProcessUtil
     /// <param name="fileName"></param>
     /// <param name="arguments"></param>
     /// <returns>true if exit code is 0</returns>
-    public static async Task<bool> RunProcessAsync(string fileName, params string[] arguments)
+    public static async Task<bool> RunProcessAsync(string fileName, string arguments, string workingDir = "")
```
WopsS, line 16:
```suggestion
    public static async Task<bool> RunProcessAsync(string fileName, string arguments, string workingDir = string.Empty)
```

rfuzzo: "Default parameter value needs to be compile-time constant"

PR #925, opened 2022-10-02
 Fix for automatic Cyberpunk 2077 FileNamePath lookup.

houseofcat:
# Fix for automatic Cyberpunk 2077 FileNamePath lookup.

## Implemented
- No new features.

## Fixed
- Automatic "Cyberpunk2077.exe" install location lookup sometimes did not work on first time run.

### Additional Notes
Made some optimizations.
* Introduced high probability default install location check for Steam first.
* Code duplication reduction.
  * Partially moved from the `FirstSetupViewModel.cs` into a new static `RegistryHelpers.cs` class.
* Parallel Loop state introduced to gracefully exit without finishing the loops (similar to `break;`).
* Skip second parallel loop if we already have the `FileNamePath`.
* Removal of unneeded local `void` method for assignment since assignment.
  * Really shouldn't be parallel assignments occurring other than really weird multi-official installation locations (if that is even possible).
* Removed or added blank lines where appropriate in the areas I was viewing. Any other whitespace changes can be from applying Code Document Formatting (CTRL + K,D , CTRL + K,E)
* Moving exception handling to inside the Parallel Loop as the exceptions are cleaner and not ugly `AggregateExceptions` from Parallel/TPL consolidation.

UnitTest projects were incompatible with `net6.0-windows` in Visual Studio 2022 Pro `v16.3.5`. Possible proper solution is to include unit test project with designation of `net6.0-windows` for any code you wish to be testable in the `WolvenKit.App` project.

### DIFF
--- a/WolvenKit.App/ViewModels/Dialogs/FirstSetupViewModel.cs
+++ b/WolvenKit.App/ViewModels/Dialogs/FirstSetupViewModel.cs
@@ -11,6 +11,7 @@ using Microsoft.WindowsAPICodePack.Dialogs;
 using Prism.Commands;
 using ReactiveUI;
 using ReactiveUI.Fody.Helpers;
+using WolvenKit.App.Helpers;
 using WolvenKit.Core.Interfaces;
 using WolvenKit.Functionality.Services;
 using WolvenKit.ViewModels.Dialogs;
@@ -24,7 +25,6 @@ namespace WolvenKit.App.ViewModels.Dialogs
     {
         private readonly ISettingsManager _settingsManager;
         private readonly ILoggerService _loggerService;
-        private string _cp77Eexe = "";
 
         public FirstSetupViewModel(
             ISettingsManager settingsManager,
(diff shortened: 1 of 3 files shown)


WolvenKit.App/ViewModels/Dialogs/FirstSetupViewModel.cs
```diff
@@ -166,66 +165,78 @@ private void ExecuteOpenDepotPath()
...
-                            if (Directory.Exists(installLocation.ToString()))
-                            {
-                                if (File.Exists(installLocation.ToString()))
-                                {
-                                    cp77 = Directory.GetFiles(installLocation.ToString(), "Cyberpunk2077.exe",
-                                        SearchOption.AllDirectories).First();
-                                }
-                            }
+                            if (subKeyName is null)
+                            { return; }
```
rfuzzo, line 193: We're using expanded braces.

rfuzzo (approved):
thanks! very nice

PR #927, opened 2022-10-03
Feature/node editor 2

rfuzzo:
# node editor

pull Jack's node editor branch to main


Traderain:
Can we merge this?

WolvenKit/WolvenKit.csproj
```diff
@@ -35,6 +35,8 @@
 
   <ItemGroup>
     <PackageReference Include="CommunityToolkit.Mvvm" Version="8.0.0" />
+    <PackageReference Include="Nodify" Version="1.7.0" />
```
miroiu, line 38: Hi, I'm happy to see that you use Nodify! 
You seem to be using an older version which is missing lots of new features and fixes which you can read about in the [changelog](https://github.com/miroiu/nodify/blob/master/CHANGELOG.md).

WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
```diff
@@ -0,0 +1,230 @@
...
+            CurrentMousePosition = e.GetPosition(this);
+
+            if (CurrentMousePosition != PreviousMousePosition)
+            {
+                MouseLocation = TransformPosition(CurrentMousePosition);
+
+                if (IsMouseCaptured)
+                {
+                    // Panning
+                    if (e.MiddleButton == MouseButtonState.Pressed)
```
miroiu, line 38: Just for my understanding: are you overriding `OnMouseMove` only to change the mouse button used for panning? (from right-click to mouse-middle)
This seems like a lot of work to do just that. I will create an issue to allow configuring keybindings for common editor actions.

WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
```diff
@@ -0,0 +1,230 @@
...
+            var nodes = new Dictionary<int, Microsoft.Msagl.Core.Layout.Node>();
+            var socketNodeLookup = new Dictionary<int, int>();
+            foreach (var item in ItemsHost.Children)
+            {
+                if (item is ItemContainer ic)
+                {
+                    if (ic.DataContext is INode nvm)
+                    {
+                        // without UpdateLayout(), these were 0, so backups were needed
+                        var width = ic.ActualWidth != 0 ? ic.ActualWidth : 300;
```
miroiu, line 150: You can bind to the `ActualSize` of the `ItemContainer` in v2.0.0 and move the layout logic into the view model.

WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
```diff
@@ -0,0 +1,230 @@
...
+                            var node = nodes[nvm.GetHashCode()];
+                            nvm.Location = new Point(
+                                node.Center.X - graph.BoundingBox.Center.X - ic.ActualWidth / 2,
+                                node.Center.Y - graph.BoundingBox.Center.Y - ic.ActualHeight / 2);
+                        }
+                    }
+                }
+
+                // center to 0,0 since we're subtracting the graph's center from each location
+                BringIntoView(new Point(0, 0));
```
miroiu, line 219: I don't see why this would be necessary. You never move the editor's `Location`.

WolvenKit/Views/Documents/RDTGraphView.xaml.cs
```diff
@@ -23,6 +24,28 @@ public partial class RDTGraphView : ReactiveUserControl<RDTGraphViewModel>
         public RDTGraphView()
         {
             InitializeComponent();
+
+            EventManager.RegisterClassHandler(typeof(BaseConnection), MouseLeftButtonDownEvent, new MouseButtonEventHandler(OnConnectionInteraction));
```
miroiu, line 28: There's a `SplitCommand` and `DisconnectCommand` on the connection that you can bind to in v2.0.0.

WolvenKit.App/ViewModels/Documents/RDTGraphViewModel.cs
```diff
@@ -69,21 +113,581 @@ public void RenderNodes(CArray<CHandle<graphGraphNodeDefinition>> nodes)
...
+                    {
+                        item.Disconnect();
+                    }
+                }
+            };
+        }
+
+        public virtual Size GetSize()
+        {
+            var size = new Size(200, 0);
```
miroiu, line 303: You could bind to the `ActualSize` of the `ItemContainer` if that's what you need here.

WolvenKit.App/Globals.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace WolvenKit.App;
+public static class Globals
+{
+    public const bool ENABLE_NODE_EDITOR = false;
```
E1337Kat, line 10: Could this be set with a application parameter with a default of false? so that like "wolvenkit.exe --with_node_editor" or something similar can be used to enable it outside of needing to build it yourself?

WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
miroiu: Here's an example of properties that you can bind to the editor if you need more control over it. And there are also some useful methods exposed. https://github.com/miroiu/nodify/blob/feature/new-blueprint/Nodifier/Graph/Graph.Editor.cs

WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
jackhumbert: That would be great! It's been a while since I wrote this, but I recall thinking it would be nice to be able to handle some custom mouse actions here as well (right-click drag, ctrl + left-click, etc).

WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
jackhumbert: I think this was made obsolete by some changes elsewhere - thanks!

WolvenKit.App/WolvenKit.App.csproj
```diff
@@ -67,6 +67,8 @@
     <PackageReference Include="WinCopies.WindowsAPICodePack" Version="2.12.0" />
     <PackageReference Include="WinCopies.WindowsAPICodePack.Shell" Version="2.12.0.2" />
     <PackageReference Include="Discord.Net" Version="3.8.0" />
+    <PackageReference Include="AutomaticGraphLayout" Version="1.1.12" />
```
E1337Kat, line 70: I don't see where this is used?

E1337Kat (commented):
Looks fairly clean, just need to remove the one thing I added before the appsettings changes.

WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
```diff
@@ -0,0 +1,232 @@
...
+                    InitialMousePosition = e.GetPosition(this);
+                }
+            }
+            else
+            {
+                base.OnMouseDown(e);
+            }
+        }
+
+        protected override void OnMouseUp(MouseButtonEventArgs e)
```
miroiu, line 89: Hi! I just released **v4.0.0** to address the configurable mouse/key bindings issue https://github.com/miroiu/nodify/issues/46. Updating to this version should remove the need to override mouse event handlers. You can have a look at [EditorGestures](https://github.com/miroiu/nodify/blob/master/Nodify/EditorGestures.cs) to see what can be configured and how. And you can also create custom interactions if needed by overriding the `NodifyEditor.GetInitialState`.

Here's the [changelog](https://github.com/miroiu/nodify/blob/master/CHANGELOG.md) 🚀

PR #928, opened 2022-10-03
General fixes and improvements

seberoth:
Implemented:
- Scroll/Pan to the properties image preview
- Notifications now always appear in the lower right corner of the main window
- Enable long path support

Fixed:
- Copy/Pasting multiple array items
- Some types were not copied correctly and a reference to the original data was left

The Sound Modding Tool can now only opened when a project is loaded


rfuzzo (approved):
Thanks lgtm

PR #929, opened 2022-10-03
mod asset browser improvements

seberoth:
Implemented:
- Archives are now sorted by load order
- Property info showing relative archive path for modded archives

Fixed:
- Always showing the first found file in the file selection instead of the current selected archive

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
@@ -376,7 +376,7 @@ namespace WolvenKit.ViewModels.Tools
                 switch (o)
                 {
                     case RedFileViewModel fileVm:
-                        AddFile(fileVm);
+                        AddFile(fileVm.GetGameFile());
                         break;
                     case RedDirectoryViewModel dirVm:
                         AddFolderRecursive(dirVm.GetModel());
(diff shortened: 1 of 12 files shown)


WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
```diff
@@ -777,10 +778,20 @@ private void CyberEnhancedSearch()
             RightItems.SuppressNotification = false;
         }
 
-        public IGameFile LookupGameFile(ulong hash)
+        public IGameFile LookupGameFile(string archiveRelativeName, ulong hash)
         {
-            DynamicData.Kernel.Optional<IGameFile> f = _archiveManager.Lookup(hash);
-            return f.HasValue ? f.Value : null;
+            foreach (var archive in _archiveManager.ModArchives.Items)
```
rfuzzo, line 783: Had this always only loaded mod archive files?

seberoth: Ofcourse.... Not

seberoth: Removed the method. Should be better now^^

rfuzzo (approved):

Issue #930, opened 2022-10-03
Logging Level Customizability and Purge Old Logs


houseofcat:
**Describe the bug**
Set current log level default to `warn` from `debug` but make it customizable for the user.

Log Pruning
Implement a file watcher that keeps `x` latest files, triggered on every file creation. Log4Net and Serilog do not have this feature built-in if I recall correctly but I have done this before.

**To Reproduce**
1. Trigger a verbose debug error. Log files fill up to 1 GB.
2. Doesn't appear to have a prune behavior.

**Expected behavior**
Ideally one shouldn't fill up a drive with logs, log at the appropriate level (but configure it), and to prune old logs from the log directory.

**Screenshots**
![image](https://user-images.githubusercontent.com/1414138/193603195-f2ce6f10-ee05-4433-8439-390da09b2720.png)

**Desktop:**
Report from Discord User: `Simarilius`
```
Just noticed that theres several 1Gb+ applog files in the folder in Appdata. Their a few days old at this point so may be todo with other things I posted bugs for that are fixed but thought I should flag it up.
```

houseofcat:
This should be resolved going forward. Any current users with large log files and older ones, this one time effort would be to purge them manually. It should be managed healthily going forward 🤞

PR #931, opened 2022-10-03
Serilog Enhancements

houseofcat:
# Serilog Enhancements
For #930
 * Better Serilog WriteTo.File default options.
 * Introduce Serilog Async for performance.
 * Customize Logging visibility level based on `#if DEBUG`.

## New Feature
Introducing [Serilog.Sinks.Async](https://github.com/serilog/serilog-sinks-async). This allows logging to be buffered and performed on background threads, thus freeing up the MAIN/UI thread when possible. I have seen some non-zero CPU utilization when logging was involved.

## Fixes
Added some healthy defaults for Serilog File sinks and introduced them as async.
`Log.CloseAndFlush()` missing on Application Exit();

## Additional Notes
Users - unless using a debug build - will no longer get the verbose logging by default. I feel this should be configurable as it could potentially interfere with troubleshooting. Unless of course its mainly just warning/error logs that are used today for triage. In which case, I recommend leaving Prod/Release builds with `Warning` level. `Information` could also be relevant happy to go either defaulting to less chatty for now. I suppose it really all depends on the quality of the logging to date of which I don't know.

* Writing log to file is now `async` and `buffered`.
* Writing log to Console is now also `async` and `buffered`.
* New logs are started every `day`.
* New logs are started every `100 MB`.
* We keep only `10` files at any given time.
* Default log level Prod/Release builds would be `Warning`.
* Console included sink is only enabled when `#if DEBUG`.
* Buffering enabled will prevent many writes to disk but achieve the same amount of logs... eventually.
* Serilog.CloseAndFlush() on Application Exit so that the application can log shutdowns gracefully.

This change will not cleanup any currently large directories.

### DIFF
--- a/WolvenKit/App.xaml.cs
+++ b/WolvenKit/App.xaml.cs
@@ -3,7 +3,6 @@ using System.IO;
 using System.Reactive.Linq;
 using System.Threading.Tasks;
 using System.Windows;
-using Microsoft.EntityFrameworkCore.Query.Internal;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using ProtoBuf.Meta;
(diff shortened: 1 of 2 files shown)


rfuzzo (approved):
Thanks, fixes the filesize neatly!

PR #934, opened 2022-10-03
Fix for ImportExportViewModel Crash

houseofcat:
# Fix for ImportExportViewModel Crash
Early exit out of `ImportSettings` with no `ActiveProject`.

## Implemented
Used `RelayCommand` attribute's property `CanExecute` and a new property to determine if we `HasActiveProject`.

## Fixed:
@seberoth mentioned how a user could try importing/load settings without having an `ActiveProject`. This leads to an `NullReferenceException` and crash to desktop. I just short circuit in this scenario and early return out without error.

## Additional Information
The crash was in a method called `SaveSettings()` but it didn't seem like any of this code should execute without the `ActiveProject` so moved the check further up the call stack.

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/ImportExportViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/ImportExportViewModel.cs
@@ -304,9 +304,8 @@ namespace WolvenKit.ViewModels.Tools
         #endregion properties
 
         public ICommand AddItemsCommand { get; private set; }
+        
         public ICommand RemoveItemsCommand { get; private set; }
-
-
 
         private bool CanAddItems(ObservableCollection<object> items) => true;
 
@@ -826,7 +825,6 @@ namespace WolvenKit.ViewModels.Tools
         /// </summary>
         public ICommand ProcessSelectedCommand { get; private set; }
 
-
         /// <summary>
         /// Execute Process selected in Import / Export Grid Command
         /// </summary>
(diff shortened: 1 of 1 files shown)


WolvenKit.App/ViewModels/Tools/ImportExportViewModel.cs
```diff
@@ -999,6 +998,13 @@ private void ResetSettings()
         [RelayCommand]
         private void ImportSettings()
         {
+            // Don't import if there is no active project.
+            if (_projectManager.ActiveProject == null
+                || string.IsNullOrWhiteSpace(_projectManager.ActiveProject.ProjectDirectory))
+            {
+                return;
```
rfuzzo, line 1005: Maybe we can log an error here or a notification? Or add a CanExecute to the command?

houseofcat: image

houseofcat: We are getting disabled which is ideal, but its not... refreshing once there is an ActiveProject. Looking for solution.

seberoth: Add `ImportSettingsCommand.NotifyCanExecuteChanged();` to `this.WhenAnyValue(x => x._projectManager.ActiveProject).Subscribe`

houseofcat: Nice I will check it out, can be done on `null` member though right (since it stats of null)?

rfuzzo (approved):

Issue #939, opened 2022-10-05
Generate Materials Errors


houseofcat:
**Describe the bug**
Fixes and performance tweaks for Generate Materials call.

**To Reproduce**
1. Go to Tools
2. Click on 'Depot Generator'
3. Click on 'Generate Materials'
4. Wait for error.

**Expected behavior**
A couple of exceptions occurring when running Generate Materials.
IndexOutOfRangeException, DivideByZeroException, and Memory Access Violation with DirectX.

**Screenshots**
Uncaught error when previously was in Parallel.ForEach().
```csharp
// Need to add a DivideByZero prevention, something like this.
var smallScale = mWidthLow == 0 ? 0 : maskWidth / mWidthLow;
```

Unsure what triggered this error but happened only once in `Debug`.
```ini
[10/5/2022 11:48:27 AM] Unhandled exception in WolvenKit v8.7.0.0
[10/5/2022 11:48:27 AM] Message: The data area passed to a system call is too small.
Source: WindowsBase
StackTrace:    at MS.Win32.UnsafeNativeMethods.GetWindowText(HandleRef hWnd, StringBuilder lpString, Int32 nMaxCount)
   at System.Windows.Automation.Peers.WindowAutomationPeer.GetNameCore()
   at System.Windows.Automation.Peers.AutomationPeer.UpdateSubtree()
   at System.Windows.ContextLayoutManager.fireAutomationEvents()
   at System.Windows.ContextLayoutManager.UpdateLayout()
   at System.Windows.ContextLayoutManager.UpdateLayoutCallback(Object arg)
   at System.Windows.Media.MediaContext.FireInvokeOnRenderCallbacks()
   at System.Windows.Media.MediaContext.RenderMessageHandlerCore(Object resizedCompositionTarget)
   at System.Windows.Media.MediaContext.RenderMessageHandler(Object resizedCompositionTarget)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
```

Uncaught error when previously was in Parallel.ForEach().
![image](https://user-images.githubusercontent.com/1414138/194115626-31792a33-834e-4816-841c-57de707832c3.png)

Triggered in `Release` at the very end. Unable to get here in `Debug` as we run out of memory.
[...]

PR #941, opened 2022-10-05
Rework Depot Operations

houseofcat:
# Rework Depot Operations

## Implemented
- Ja-to's command improvements.
- Added new parallelism style that should be more responsive to the entire system and UI.
- Added ability to feature flag revert to legacy parallelism.
  - Enhance legacy parallelism with options.

## Fixed
- Some arithmetic bug fixes on range / division operations. Relates to #939
- Better logging on Uncook exceptions that were getting lost in the ether.

## Preliminary Benchmark
```ini
i9 12900K @ 5.0 GHz, 32 GB of DDR4 4000 MHz

// DEBUG
12:37:29 pm - tweaked parallel foreach starttime
12:40:17 pm - endtime
2 minutes 48 seconds

// expected to be slower but better system responsiveness
// DEBUG
12:59:35 pm - new parallelforeach async starttime
13:01:54 pm - endtime
2 minutes 19 seconds
```

While the results showed it was faster, I would caution that the results are dependent on the system and level activity of the system. I would expect it to be slightly slower than the previous path forward. It should be assumed to be about 3 minutes on a comparable system and faster when in `Release` mode.

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
@@ -340,7 +340,13 @@ namespace WolvenKit.Modkit.RED4
             var heightInTiles0 = DivCeil(maskHeight, maskTileSize);
             var smallOffset = widthInTiles0 * heightInTiles0;
 
-            var smallScale = maskWidth / mWidthLow;
+            // DivideByZero preventions.
+            // maskWidthLow == 0
+            // maskWidth < mWidthLow => maskWidth / mWidthLow = 0 because (int) Math
+            var smallScale =
+                mWidthLow == 0 || mWidthLow !< maskWidth
+                ? 1
+                : maskWidth / mWidthLow;
 
             for (uint x = 0; x < maskWidth; x++)
             {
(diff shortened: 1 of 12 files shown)


WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -340,7 +340,13 @@ private static void Decode(ref byte[] maskData, uint maskWidth, uint maskHeight,
...
             var smallOffset = widthInTiles0 * heightInTiles0;
 
-            var smallScale = maskWidth / mWidthLow;
+            // DivideByZero preventions.
+            // maskWidthLow == 0
+            // maskWidth < mWidthLow => maskWidth / mWidthLow = 0 because (int) Math
+            var smallScale =
+                mWidthLow == 0 || mWidthLow !< maskWidth
+                ? 1
+                : maskWidth / mWidthLow;
```
houseofcat, line 350: @seberoth Was getting DivideByZero here.

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -362,7 +368,11 @@ private static void DecodeSingle(ref byte[] maskData, uint maskWidth, uint maskH
             var tileIndex = (widthInTiles * yTile) + xTile + tilesOffset;
 
             var paramOffset = tilesData[tileIndex * 2];
-            var paramBits = tilesData[(tileIndex * 2) + 1];
+
+            // IndexOutOfRange Prevention
+            var paramBits = tilesData.Length > (tileIndex * 2) + 1
```
houseofcat, line 373: @seberoth Getting length less than expected on the array here.

Relates to #939

rfuzzo (approved):

Issue #805, opened 2022-06-04
Command line option to uncook meshes with Materials


njibhu:
**Is your feature request related to a problem? Please describe.**
It's currently very hard to export meshes with Materials using the CLI.
We need first to uncook the mesh, then run the CLI with cr2w to serialize the mesh, extract the material buffer from the serialized mesh, the run cr2w again but this time on the extracted material buffer. That's very unperformant, and waste a lot of disk space, particularly when running on big batches of meshes.

**Describe the solution you'd like**
Would be great to have the option to uncook meshes with Materials, and I guess with an extra command line argument to take the path of the MaterialDepot.

**Additional context**
Currently building a streamingsector importer for a third-party program, I'm interfacing with the Wolvenkit CLI to automatically export the right meshes, to import them in the other program.

njibhu:
Found a workaround: 
In the settings file (`WolvenKit.CLI.exe settings`):
```
  "MeshExportArgs": {
    "MeshExportType": "WithMaterials",
    "MaterialRepo": "path\\to\\materialrepo",
    "ArchiveDepot": "path\\to\\cyberpunk\\archive\\pc\\content\\"
  }
```

njibhu:
Quick idea on how to fix it without requiring a lot of new cli options.
The main issue I have with it is when using automated extract scripts:
- A way to solve it would be by being able to pass as an arg the path of a custom wolvenkit config file. 
- if I need to change the export settings I can just have two custom wolvenkit settings files and switch between the two by changing the custom config path argument

PR #942, opened 2022-10-05
Add --mesh-export options to the uncook CLI command

njibhu:
# Add mesh export options to the uncook CLI command

Add CLI options to export meshes.
`--mesh-export-type` which matches the export types in the UI
`--mesh-export-material-repo` which optionally allows to have a material repo in a different folder than the output

Added as well a small refactor for the uncook options to make it easier in the future to add more options.

When reviewing, I recommend enabling the hide whitespace changes option, it renders the refactor much easier to read.

Fixed:
- #805


rfuzzo (approved):
looks good to me

PR #948, opened 2022-10-06
Add extra nullchecks for the RenderResourceBlob

njibhu:
# Add extra nullchecks for the RenderResourceBlob

This nullcheck already exists in a few places, but it seems some of the assertions were forgotten.
Wolvenkit doesn't really support extracting non-rendered meshes. There's a few of them in the base folder of the archives, and right now they trigger NullPointerExceptions.

If we want to support these later, we'll probably want to have different extracting functions altogether, or it would need at least some bigger refactoring.


rfuzzo (approved):
great, thanks!

PR #951, opened 2022-10-08
Add Arrow Key Support for Asset Browser Navigation

WRodewald:
Add Arrow Key Support for Asset Browser Navigation

Implemented:
- Add left and right arrow key support for Asset Browser navigation (left column). The keys will collapse / expand nodes and jump up and down the hierarchy.

Notes:
The tooltip hover behaves really wonky. I don't know if my changes resulted in the behavior of if they just "spotlight" them. Also, I had to register to KeyUp since the SfTreeGrid eats left and right key presses as it would normally use them to move between columns. Not sure how to solve this.


rfuzzo (approved):
lgtm

PR #955, opened 2022-10-10
Update "Open in Asset Browser" to full expand Browser to file path

dragonzkiller:
Implemented:
- Asset Browser will open the directory tree to the Depo Path and scroll it into view
- Right side File View will show all files in directory, but select desired file
- Notification system will now warn if the file could not be found in the Asset Browser

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
@@ -547,7 +547,7 @@ namespace WolvenKit.ViewModels.Tools
         private void MoveToFolder(RedDirectoryViewModel dir) => LeftSelectedItem = dir.GetModel();
 
         /// <summary>
-        /// Navigates the right-side of the browser to the existing file
+        /// Navigates the Asset Browser to the existing file.
         /// </summary>
         /// <param name="file"></param>
         public void ShowFile(FileModel file)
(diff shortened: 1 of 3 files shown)


WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
```diff
@@ -561,8 +561,30 @@ public void ShowFile(FileModel file)
...
 
-            RightItems.Clear();
-            RightItems.AddRange(list);
+            if (list.Count > 0)
+            {
+                var fullPath = "";
+                var parentDir = LeftItems.ElementAt(0);
+                parentDir.IsExpanded = true;
+
+                foreach (var dir in file.RelativePath.Split('\\').SkipLast(1))
```
seberoth, line 570: Replace `\\` with `Path.DirectorySeparatorChar`

WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
```diff
@@ -561,8 +561,30 @@ public void ShowFile(FileModel file)
...
+                {
+                    fullPath += dir;
+                    parentDir = parentDir.Directories
+                        .Where(x => x.Key == fullPath)
+                        .First()
+                        .Value;
+                    parentDir.IsExpanded = true;
+                    fullPath += '\\';
```
seberoth, line 578: Replace `\\` with `Path.DirectorySeparatorChar`

seberoth (changes requested):

PR #964, opened 2022-10-13
Feature/node editor 2 update

E1337Kat:
# some small updates to the node editor branch

Implemented:
-  A command line argument to enable the node editor if so desired.

Fixed:
- Bumped nodify version to 3.0.0 and made changes necessary to minimally support it.

### DIFF
--- a/WolvenKit.App/ViewModels/Shell/AppViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/AppViewModel.cs
@@ -207,6 +207,14 @@ namespace WolvenKit.ViewModels.Shell
                 return false;
             }
 
+            if (args[1] == "--node_editor_enabled")
+            {
+                var message = $"Node Editor Enabled";
+                _loggerService.Success(message);
+                MessageBox.Show(message);
+                return false;
+            }
+
             if (!File.Exists(args[1]))
             {
                 var message = $"Sorry, '{args[1]}' could not be found";
--- a/WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
+++ b/WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
@@ -30,17 +30,16 @@ namespace WolvenKit.Views.Editors
 
             if (CurrentMousePosition != PreviousMousePosition)
             {
-
-                //MouseLocation = TransformPosition(CurrentMousePosition);
-                // ^ `TransformPosition` was removed from NodifyEditor. Rather than trying to figure out the "correct" way to fix it,
-                // I hard coded it to what the definition of the function _was_ just to get things to build correctly on Nodify 3.0.0
-                MouseLocation = new Point((ViewportLocation.X + CurrentMousePosition.X) / ViewportZoom, (ViewportLocation.Y + CurrentMousePosition.Y) / ViewportZoom);
+                // All coordinates that the editor works with should be inside the ItemsHost. Translating coordinates is not needed if,
+                // for example, you want to move the viewport to a certain location inside the editor's space. (like focusing a node)
+                MouseLocation = e.GetPosition(ItemsHost);
                 if (IsMouseCaptured)
                 {
                     // Panning
                     if (e.MiddleButton == MouseButtonState.Pressed)
                     {
-                        ViewportLocation -= CurrentMousePosition - PreviousMousePosition;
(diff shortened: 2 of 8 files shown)


dragonzkiller:
How does this work with the changes and suggestions by @miroiu in #927 ?

E1337Kat:
The biggest thing is bumping the version of nodify to 3.0.0 from 1.7.0, including fixing name changes for nodify methods. Ideally I want to further explore the nodify changes to not hard code now deleted code. This along with the main node-editor_2 branch provide a more solid development base to work with, imo. 

mostly I wanted to have the node editor be soft disabled so that it can still be used somewhat.

rfuzzo (approved):
This looks good to me, we can work in the rest of the comments on the other branch

WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
```diff
@@ -30,14 +30,17 @@ protected override void OnMouseMove(MouseEventArgs e)
 
             if (CurrentMousePosition != PreviousMousePosition)
             {
-                MouseLocation = TransformPosition(CurrentMousePosition);
 
+                //MouseLocation = TransformPosition(CurrentMousePosition);
+                // ^ `TransformPosition` was removed from NodifyEditor. Rather than trying to figure out the "correct" way to fix it,
+                // I hard coded it to what the definition of the function _was_ just to get things to build correctly on Nodify 3.0.0
+                MouseLocation = new Point((ViewportLocation.X + CurrentMousePosition.X) / ViewportZoom, (ViewportLocation.Y + CurrentMousePosition.Y) / ViewportZoom);
```
miroiu, line 37: All coordinates that the editor works with should be inside the `ItemsHost`. Translating coordinates is not needed if, for example, you want to move the viewport to a certain location inside the editor's space. (like focusing a node)

`MouseLocation =  e.GetPosition(ItemsHost);` should do the trick.

WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
```diff
@@ -30,14 +30,17 @@ protected override void OnMouseMove(MouseEventArgs e)
...
                 if (IsMouseCaptured)
                 {
                     // Panning
                     if (e.MiddleButton == MouseButtonState.Pressed)
                     {
-                        Offset -= CurrentMousePosition - PreviousMousePosition;
+                        ViewportLocation -= CurrentMousePosition - PreviousMousePosition;
```
miroiu, line 43: `CurrentMousePosition` and `PreviousMousePosition` cannot be relative to the `ItemsHost` because that control changes its transform based on the mouse delta, hence we need to scale it by `ViewportZoom`.

`ViewportLocation -= (CurrentMousePosition - PreviousMousePosition) / ViewportZoom;`

> I am thinking of simplifying this in the future.

E1337Kat: Ah, that makes sense. When testing it to make sure it worked, i noticed that zoomed out was annoyingly slow to pan.

WolvenKit/Views/Editors/AutomaticNodifyEditor.cs
E1337Kat: :naruhodo: I will add that in. Thanks!

WolvenKit.App/ViewModels/Shell/AppViewModel.cs
```diff
@@ -207,6 +207,14 @@ private bool TryLoadingArguments()
                 return false;
             }
 
+            if (args[1] == "--node_editor_enabled")
```
rfuzzo, line 210: Btw, this is a good idea, but in the long run I think it would make sense to move such checks to an appsettings or ini file or something - because the node editor will at some point just be globally enabled. The only reason I disabled it was to get it merged to mainline but hide it from users until we enable it on the nightly.

rfuzzo (approved):

WolvenKit.App/ViewModels/Shell/AppViewModel.cs
E1337Kat: I figured this was the reason, but wanted to give an option just in case. I could try and make such a change if you wanted? also perhaps handle the other args in the same way as this so that their order doesn't matter so much... but maybe in a different branch.

PR #975, opened 2022-10-22
Sector Editing Tweaks

justarandomguyintheinternet:
- Adds a `QuestPrefabRefHash` to all nodes that get added by importing into a sector from `.json`. The hash is needed to make objects interactable and persistent
- Log a clicked mesh's name to the log in sector preview, makes finding the right node a lot easier

### DIFF
--- a/WolvenKit.App/ViewModels/Documents/MeshPreviewBlock.cs
+++ b/WolvenKit.App/ViewModels/Documents/MeshPreviewBlock.cs
@@ -322,10 +322,8 @@ namespace WolvenKit.ViewModels.Documents
                     SelectedItem = group;
                     e.Handled = true;
                 }
-                if (e.HitTestResult.ModelHit is SubmeshComponent lod)
+                if (e.HitTestResult.ModelHit is SubmeshComponent { Parent: MeshComponent { Parent: MeshComponent mesh } })
                 {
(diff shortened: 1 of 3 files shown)


WolvenKit.App/ViewModels/Documents/MeshPreviewBlock.cs
```diff
@@ -322,6 +322,12 @@ public void OnMouseDown3DHandler(object sender, MouseDown3DEventArgs e)
                     SelectedItem = group;
                     e.Handled = true;
                 }
+                if (e.HitTestResult.ModelHit is SubmeshComponent lod)
```
seberoth, line 325: Could use pattern matching for that
`e.HitTestResult.ModelHit is SubmeshComponent { Parent: MeshComponent { Parent: MeshComponent mesh } }`

seberoth (changes requested):

WolvenKit.App/ViewModels/Documents/MeshPreviewBlock.cs
justarandomguyintheinternet: Good to know, added that. Thanks a lot :D

seberoth (approved):

PR #978, opened 2022-10-25
Faster archive loading

seberoth:
Faster archive loading

Implemented:
- Faster archive loading

Fixed:
- Faster archive loading

Faster archive loading


rfuzzo (approved):
Faster

PR #1000, opened 2022-11-08
ImportExportViewModel Refactoring: consolidating Process Selected and Process All paths

kwekmaster:
Title says it all. Referencing methods will not feel any difference.

**Implemented:**
- Congruent code paths moved into a new Process Bulk method

**Testing Done:**
- Ensure that Process All and Process Selected works just like before from the UI.


rfuzzo (approved):
👍

Issue #926, opened 2022-10-03
Script deployment support


keless:
"The point is to have an all in one tool for creating mods for the games made with the engine."  https://redmodding.org/ description of WolvenKit

Currently if I have a /scripts folder in my WolvenKit Project, when I "pack + install" my project, it completely ignores the scripts folder.

I realize there are a number of different scripting solutions for Cyberpunk right now, and each requires a different configuration for deploying scripts. As a bare minimum it would be nice to have a configuration option per project to specify a folder to copy files from the project's scripts folder into.

For instance; when using RedScript, files and folders under < projectfolder >/scripts/*  should be copied into < gamefolder >/r6/scripts/< projectname >/*

It could be nice if a project could let you specify the type of scripting solution you intend to use (CET lua, RedScript .reds, REDMod .script, etc) and automatically know where to deploy based on that (and possibly check that the appropriate prerequisite mod framework CET/SCC/etc tools/mods are installed properly). 

Alternatively, a simple "copy <project>/scripts/ to: <path>" setting would also work.

This behavior would then be triggered whenever "Pack & Install" or "Install & Run" are executed.

kwekmaster:
@rfuzzo -- forgot to mention. Taking this, please assign to me. Thanks.

PR #1004, opened 2022-11-09
Enhancement--Script deployment support. Issue 926

kwekmaster:
Addresses https://github.com/WolvenKit/WolvenKit/issues/926

- Specification discussion was done earlier thru discord and can be ref'd if needed.
- Supports all present and future scripting elements--assuming they are deployed under the game directory.  
- The sample testbed I used contain CET, redscript, and red4ext dlls for deployment.

**Implemented**

_Assumptions for the preliminary approach: the user is responsible for structuring their scripts to match the target deployment directory structure._

- The ability to deploy scripting elements user puts under `resource` into `packed`. Please see the testbed below for a sample structure.
- Simple as-is copying, skipping the special `yaml` and `xl` files.

**Notes**
Hot deployment seems to follow a separate path and is not touched by this PR.

**Testing Done**
- Using everything in the test bed below
- With only yaml/xl files
- With blank `resources`
- Both as regular and REDmod 

Testing meets the requirements.

**Test Bed**
Tests are done the following `resources` structure. `.xl` and `.yaml` files are added thru clicking `(+)New File`
image

### DIFF
--- a/WolvenKit.App/Controllers/RED4Controller.cs
+++ b/WolvenKit.App/Controllers/RED4Controller.cs
@@ -322,11 +322,11 @@ namespace WolvenKit.Functionality.Controllers
             _loggerService.Success($"{cp77Proj.Name} archiveXL files packed into {cp77Proj.GetPackedArchiveDirectory(options.IsRedmod)}");
 
             // pack 'script' files
-            if (!PackScriptFiles(cp77Proj))
+            if (!PackResources(cp77Proj))
             {
                 _progressService.IsIndeterminate = false;
-                _loggerService.Error("Packing script files failed, aborting.");
-                _notificationService.Error("Packing script files failed, aborting.");
+                _loggerService.Error("Packing other resource files failed, aborting.");
+                _notificationService.Error("Packing other resource files failed, aborting.");
                 return false;
             }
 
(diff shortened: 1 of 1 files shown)


WolvenKit.App/Controllers/RED4Controller.cs
```diff
@@ -463,6 +475,33 @@ private bool PackArchives(Cp77Project cp77Proj, LaunchProfile options)
 
             return true;
         }
+
+
+        private static bool PackScriptFiles(Cp77Project cp77Proj)
+        {
+            //preliminary approach -- everything under resources, except for *.yaml, *.xl, is a 'script' element
```
rfuzzo, line 482: hmm, this doesn't seem right? Scripts should only be ".reds" files... I know it's preliminary, but I think this is confusing. You could change the name of the function to "PackResources" I suppose.

rfuzzo (changes requested):

WolvenKit.App/Controllers/RED4Controller.cs
kwekmaster: Although I have concerns about Scripts == `*.reds` exclusively, I do agree that this preliminary approach packs resources. I will change the method to "PackResources" so we have a path for users (like me) to deploy their scripts :)

The change will be made.

kwekmaster: Changes completed!  Please review.

rfuzzo (approved):
thank you!

PR #1013, opened 2022-11-20
Add the DepotPath subdirs to serialized convert

solaarTW:
# Pull request template

Add the DepotPath subdirs to serialized convert

Implemented:
- If DepotPath structure exists in input file then apply structure to output

Fixed:
- NA

I don't know how to code execution arguments like --help. It would be helpful to have one that's included in the if-condition to write the serialized files into a single directory or create the DepotPath directory structure.

### DIFF
--- a/WolvenKit.Modkit/RED4/RedConverter.cs
+++ b/WolvenKit.Modkit/RED4/RedConverter.cs
@@ -65,7 +65,7 @@ namespace WolvenKit.Modkit.RED4
                 // Would be nice to make this a runtime option like --structure "[none|depotpath]" or something
                 if (destination.IndexOf(@"\\source\\archive\\") > 0)
                 {
-                    destination = Regex.Split(destination, @"\\source\\archive\\")[1];
+                    destination = Regex.Split(destination, "\\source\\archive\\")[1];
                     destination = Path.Combine(outputDirInfo.FullName, destination);
                     Directory.CreateDirectory(destination);
                 }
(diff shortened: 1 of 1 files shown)


solaarTW (commented):
Removed @literal from IndexOf() and recommitted.
Changing write destination causes false error with mismatch on outputDirInfo

WolvenKit.Modkit/RED4/RedConverter.cs
```diff
@@ -58,7 +60,16 @@ public bool ConvertToAndWrite(ETextConvertFormat format, string infile, Director
...
+                var destination = Path.GetDirectoryName(infile);
+                // If DepotPath structure exists then add it to output directory
+                // Would be nice to make this a runtime option like --structure "[none|depotpath]" or something
+                if (destination.IndexOf(@"\\source\\archive\\") > 0)
+                {
+                    destination = Regex.Split(destination, "\\source\\archive\\")[1];
+                    destination = Path.Combine(outputDirInfo.FullName, destination);
+                    Directory.CreateDirectory(destination);
+                }
+                var outpath = Path.Combine(destination, $"{Path.GetFileName(infile)}.{format}");
```
dragonzkiller, line 72: This entire thing can be done with:

```csharp
var outDir = Path.GetDirectoryName(infile);
var idx = outDir.IndexOf(@"\source\archive");
if (idx != -1)
{
    outDir = Path.Combine(outputDirInfo.FullName, outDir.Substring(idx + 16));
    Directory.CreateDirectory(outDir);
}
var outpath = Path.Combine(outDir, $"{Path.GetFileName(infile)}.{format}");
```
Again, you'd have to check for the actual directory separator, and not `\` though. so some formatting for that would have to be added

dragonzkiller (changes requested):

PR #1020, opened 2022-11-28
Add the apps list of the mesh to the material.json

Simarilius-uk:
Add the apps list of mesh items to the material.json when exporting withMaterial

Implemented:
- Add the apps list of mesh items to the material.json when exporting withMaterial

This will allow the Blender plugin to be modified to only bring in a selected material, currently theres not enough information to go from entity>appearance>mesh and select the right material as the appearance names and material names can be different in the mesh. 
I've done some exports and cant find any problems, and have imported the resulting files to blender with the current plugin without issue.


Simarilius-uk:
Also has the fix to allow scenerid anims to be exported when merged into anim files.

PR #1024, opened 2022-12-01
Clean all button--issue 1022. 

kwekmaster:
Thought this would be a simple UI addition, but it turns out not so:

Prior to this proposed change, the only way for the user to cleanup changes/removals/renames to "other" (e.g. `r6/scripts/myscript.reds` => `r6/scripts/thecorrectnameformyscript.reds`) already-packed files under `resources` is to **manually delete** them from the `packed` directory themselves.

I also added `LaunchProfile` entry and missing code path to `Cleanup()` to completely clean as an option. This allows it to clean misc items added or removed in resources dir, such as scripts. 

Implemented and Added:
- The `Clean All` button
image

- Added `LaunchProfile` options to do so.
image

- Added optional code path to `Cleanup()` to handle the rest of the `resources/` dir not covered by it. Options are done thru `LaunchProfile`s. The user now can click `Clean All` or integrate it in their profile.

Testing:
- Ensure `Clean All` does it's purpose, without any UI desync or oddness observed using the current removal method.
- Ensure default behavior persists with the `LaunchProfile` option unchecked
- Ensure intended behavior works with the `LaunchProfile` option checked
- Integration with Install to Game: good integration. It also removes the logged files not found in the current `packed/` directory.

### DIFF
--- a/WolvenKit.App/Controllers/IGameController.cs
+++ b/WolvenKit.App/Controllers/IGameController.cs
@@ -18,5 +18,6 @@ namespace WolvenKit.Functionality.Controllers
 
         Task<bool> LaunchProject(LaunchProfile profile);
         bool PackProjectHot();
+        bool CleanAll();
     }
 }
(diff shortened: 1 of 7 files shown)


kwekmaster:
Since I am currently traveling, I will have sporadic internet access

However, when I do, I shall address any concerns you may have and make any changes necessary.

Issue #1035, opened 2022-12-17
Mesh Preview Exception (#submeshes > appearance/#CNames)


manavortex:
I've decided to split the netrunner suit's thighs and calves into submeshes for purposes of chunkmask hiding. After importing the mesh, I'm running into a number of odd behaviours when switching to the Mesh Preview tab (as detailed below).

### To Reproduce
1. Add T-Bug's Netrunner suit to a project. 
2. Import the attached 5_submeshes.glb
3. Open the mesh in WKit
4. Switch to the Mesh Preview tab

[5_submeshes.zip](https://github.com/WolvenKit/WolvenKit/files/10252238/5_submeshes.zip)

Nothing is rendering. I'm seeing the following exception: 
```
[17/12/2022 17:59:15] Unhandled exception in WolvenKit v8.8.1.0
[17/12/2022 17:59:15] Message: Index was outside the bounds of the array.
Source: WolvenKit.Modkit
StackTrace:    at WolvenKit.Modkit.RED4.Tools.MeshTools.ContainRawMesh(MemoryStream gfs, MeshesInfo info, Boolean lodFilter, UInt64 chunkMask, Boolean mergeMeshes) in D:\a\WolvenKit\WolvenKit\WolvenKit.Modkit\RED4\Tools\MeshTools.cs:line 456
   at WolvenKit.ViewModels.Documents.RDTMeshViewModel.MakeMesh(CMesh cMesh, UInt64 chunkMask, Int32 appearanceIndex) in D:\a\WolvenKit\WolvenKit\WolvenKit.App\Helpers\MeshHelpers.cs:line 78
   at WolvenKit.ViewModels.Documents.RDTMeshViewModel.RenderMesh() in D:\a\WolvenKit\WolvenKit\WolvenKit.App\ViewModels\Documents\RDTMeshViewModel.cs:line 346
   at WolvenKit.ViewModels.Documents.RDTMeshViewModel.<.ctor>b__90_0(CompositeDisposable disposables) in D:\a\WolvenKit\WolvenKit\WolvenKit.App\ViewModels\Documents\RDTMeshViewModel.cs:line 241
   at ReactiveUI.ViewForMixins.<>c__DisplayClass4_0.<WhenActivated>b__0() in /_/src/ReactiveUI/Activation/ViewForMixins.cs:line 101
[...]

PR #1042, opened 2022-12-22
issue 1035: prevent exception (does not write back yet)

manavortex:
# Fix mesh preview exception on material

Fixed:
- https://github.com/WolvenKit/WolvenKit/issues/1035

Doesn't write back yet, but at least it renders?

### DIFF
--- a/WolvenKit.App/Helpers/MeshHelpers.cs
+++ b/WolvenKit.App/Helpers/MeshHelpers.cs
@@ -437,7 +437,7 @@ namespace WolvenKit.ViewModels.Documents
         public ICommand LoadMaterialsCommand { get; set; }
         public void LoadMaterials()
         {
-            if (null == SelectedAppearance)
+            if (SelectedAppearance == null)
             {
                 Locator.Current.GetService<ILoggerService>().Warning($"No material selected!");
                 return;
--- a/WolvenKit.App/Helpers/MeshHelpers.cs
+++ b/WolvenKit.App/Helpers/MeshHelpers.cs
@@ -437,6 +437,11 @@ namespace WolvenKit.ViewModels.Documents
         public ICommand LoadMaterialsCommand { get; set; }
         public void LoadMaterials()
         {
+            if (null == SelectedAppearance)
+            {
+                Locator.Current.GetService<ILoggerService>().Warning($"No material selected!");
+                return;
+            }
             IsLoadingMaterials = true;
             Parallel.ForEachAsync(from entry in SelectedAppearance.RawMaterials orderby entry.Key ascending select entry, (material, cancellationToken) => LoadMaterial(material.Value)).ContinueWith((result) =>
             {
--- a/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
@@ -652,6 +652,8 @@ namespace WolvenKit.Modkit.RED4.Tools
                 var apps = info.appearances.Keys.ToList();
                 for (var e = 0; e < apps.Count; e++)
                 {
+                    // Null-proof materials
+                    info.appearances[apps[e]] = InitializeDefaultMaterial(info.meshCount, info.appearances[apps[e]]);
                     meshContainer.materialNames[e] = info.appearances[apps[e]][index];
                 }
 
@@ -661,6 +663,43 @@ namespace WolvenKit.Modkit.RED4.Tools
             RemoveDoubleFaces(ref expMeshes);
             return expMeshes;
         }
+
+        private static string FindNextInSequence(IReadOnlyList<string> sequence, string current){
+            for (var i = 0; i < sequence.Count - 1; i++)
+            {
+                if (sequence[i] != current || sequence[i + 1] == current)
+                {
+                    continue;
+                }
+                return sequence[i + 1];
+            }
+            return null;
+        }
+
+        private static string[] InitializeDefaultMaterial(int numTotalSubmeshes, string[] meshAppearances)
+        {
+            List<string> ret = meshAppearances.ToList();;
+            // If the array is empty: initialize it with "default" material
+            if (ret.Count == 0)
+            {
+                ret.Add("default");
+            }
+
+            // append all missing appearance names
+            for (var e = ret.Count; e < numTotalSubmeshes; e++)
+            {
+                // get last item from array, which is never falsy because of the above statement 
+                var lastUsedString = ret[^1];
+
+                var nextUsedString = FindNextInSequence(ret, lastUsedString) ?? lastUsedString;
+                
+                ret.Add(nextUsedString);
+            }
+
+            return ret.ToArray();
+
+        }
+        
         public static void UpdateMeshJoints(ref List<RawMeshContainer> meshes, RawArmature newRig, RawArmature oldRig)
         {
             // updating mesh bone indices
--- a/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
@@ -664,7 +664,8 @@ namespace WolvenKit.Modkit.RED4.Tools
             return expMeshes;
         }
 
-        private static string FindNextInSequence(IReadOnlyList<string> sequence, string current){
+        private static string FindNextInSequence(IReadOnlyList<string> sequence, string current)
+        {
             for (var i = 0; i < sequence.Count - 1; i++)
             {
                 if (sequence[i] != current || sequence[i + 1] == current)
--- a/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
@@ -650,6 +650,7 @@ namespace WolvenKit.Modkit.RED4.Tools
 
                 meshContainer.materialNames = new string[info.appearances.Count];
                 var apps = info.appearances.Keys.ToList();
+                
                 for (var e = 0; e < apps.Count; e++)
                 {
                     // Null-proof materials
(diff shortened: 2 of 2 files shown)


WolvenKit.Modkit/RED4/Tools/MeshTools.cs
```diff
@@ -661,6 +663,43 @@ public static List<RawMeshContainer> ContainRawMesh(MemoryStream gfs, MeshesInfo
             RemoveDoubleFaces(ref expMeshes);
             return expMeshes;
         }
+
+        private static string FindNextInSequence(IReadOnlyList<string> sequence, string current){
```
WopsS, line 667: Formatting, `{` on new line.

WolvenKit.Modkit/RED4/Tools/MeshTools.cs
```diff
@@ -661,6 +663,43 @@ public static List<RawMeshContainer> ContainRawMesh(MemoryStream gfs, MeshesInfo
...
+                    continue;
+                }
+                return sequence[i + 1];
+            }
+            return null;
+        }
+
+        private static string[] InitializeDefaultMaterial(int numTotalSubmeshes, string[] meshAppearances)
+        {
+            List<string> ret = meshAppearances.ToList();;
```
WopsS, line 681: Why converting this to a list instead of doing it on the array? Below you will convert it back to array.
If you want the array's length, you can do `meshApparences.Length`.

WolvenKit.Modkit/RED4/Tools/MeshTools.cs
```diff
@@ -661,6 +663,43 @@ public static List<RawMeshContainer> ContainRawMesh(MemoryStream gfs, MeshesInfo
...
+                ret.Add("default");
+            }
+
+            // append all missing appearance names
+            for (var e = ret.Count; e < numTotalSubmeshes; e++)
+            {
+                // get last item from array, which is never falsy because of the above statement 
+                var lastUsedString = ret[^1];
+
+                var nextUsedString = FindNextInSequence(ret, lastUsedString) ?? lastUsedString;
```
WopsS, line 694: Does it need to reuse appearance names to fill in the array?

manavortex: It could also randomly assign something (e.g. default), but I've given an example of how those arrays usually look in the linked [issue](https://github.com/WolvenKit/WolvenKit/issues/1035), and this takes care of all of these cases.

WolvenKit.Modkit/RED4/Tools/MeshTools.cs
manavortex: This avoids out of bound errors in case of arrays with a static length (which the compiler hit me with when I tried to open an ancient mesh). I figured better safe than sorry :D

WolvenKit.App/Helpers/MeshHelpers.cs
```diff
@@ -437,6 +437,11 @@ public List<Material> GetMaterialsForAppearance(CMesh mesh, CName appearance)
         public ICommand LoadMaterialsCommand { get; set; }
         public void LoadMaterials()
         {
+            if (null == SelectedAppearance)
```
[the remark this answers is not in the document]
manavortex: Out of curiosity, why not? Will be a difficult habit to break, I've spent too much time in Java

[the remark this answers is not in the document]
WopsS: Coding style of WolvenKit.
Besides this, C# doesn't let you assign variables inside `if`s, so no reason to make sure about that.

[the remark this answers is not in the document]
manavortex: Fair, I'll try to keep it in mind. It's a global habit I've developed, so... yeah, it'll be hard to break.

WolvenKit.Modkit/RED4/Tools/MeshTools.cs
WopsS: Then I would say to just use `e` with condition on delta (i.e. `numTotalSubmeshes - ret.Count`) instead of finding the next in sequence, it should be the same behavior.

manavortex: I'm not familiar enough with c# for that, can you sketch it out for me?

manavortex: Wait, I think that does something different?

WopsS: The way I understand it now is that you do the following:

```
numTotalSubmeshes: 7
meshAppearances: [ mesh1, mesh2, mesh3, mesh4, , mesh5 ]

FindNextInSequence(numTotalSubmeshes, meshAppearances): [ mesh1, mesh2, mesh3, mesh4, , mesh5, mesh1, mesh2 ]
```

This way I do not see a reason why there should be 2 for-loops, your entire `InitializeDefaultMaterial` is having a worst-case complexity of `O(n^2)`, assume this:

```
numTotalSubmeshes: 7
meshAppearances: [ mesh1 ]
```

When you call `InitializeDefaultMaterial` your algorithm will start doing more comparisons with each element you add to `ret`, first it will do 1, then 2, then 3, ... then 6.

Another example of output is:

```
numTotalSubmeshes: 7
meshAppearances: [ mesh1, mesh2, mesh3 ]

FindNextInSequence(numTotalSubmeshes, meshAppearances): [ mesh1, mesh2, mesh3, mesh1,  mesh2, mesh1, mesh2 ]
```

Note that `mesh3` is not used anymore, which I assume is not expected behavior.

So what I was proposing was something like this:

```
for (int i = 0; i < (numTotalSubmeshes - ret.Count); i++)
{
    var newItem = ret[i];
    ret.Add(newItem);
}
```
I should do the same thing you are doing now.

Or am I not fully understanding what you want to do here?

WopsS: Ops, my bad the second statement (about `mesh3`) is not valid.

manavortex: I get what you mean now, but that would always start at the beginning. Means if I'd extend a list to 5 entries, your solution would lead to
```
[ mat1, mat2, mat3, mat1 ] => [ mat1, mat2, mat3, mat1, mat1 ]
```
and mine should be
```
[ mat1, mat2, mat3, mat1 ] => [ mat1, mat2, mat3, mat1, mat2 ]
```
... since it checks the last used item in the sequence.

I'm sure you could achieve the same sorcery with lambdas and streams somehow, but I don't know how ':)

WopsS: Ops, my bad the second statement (about `mesh3`) is not correct, should be another output, that is `[ mesh1, mesh2, mesh3, mesh3,  mesh3, mesh3, mesh3 ]`.

manavortex: Yeah, my idea was since the original meshes are set up to always repeat the same sequence, the autocomplete should for reasons of maximum convenience do the same.
It might probably be better to move all this to the mesh import section so it can get written to disk? But I couldn't find that. :D Maybe I'll do that next.

WopsS: > I get what you mean now, but that would always start at the beginning. Means if I'd extend a list to 5 entries, your solution would lead to
> 
> ```
> [ mat1, mat2, mat3, mat1 ] => [ mat1, mat2, mat3, mat1, mat1 ]
> ```
> 
> and mine should be
> 
> ```
> [ mat1, mat2, mat3, mat1 ] => [ mat1, mat2, mat3, mat1, mat2 ]
> ```
> 
> ... since it checks the last used item in the sequence.
> 
> I'm sure you could achieve the same sorcery with lambdas and streams somehow, but I don't know how ':)

Ah, I see now. But then I would a for-loop before appending to the list to get the last index of the mesh, e.g.:

```csharp
var lastIndex = ret.TakeWhile(s => s == ret[^1]).Count();

// TODO: Do all the checks for out of bounds.

for (var e = ret.Count; e < numTotalSubmeshes; e++)
{
    ret.Add(ret[lastIndex++]);
}
```

manavortex: That won't update, though, will it? Like... it might be overkill to recalculate the next sequence item each time. Maybe we could figure out the sequence length, and use that as an offset? but it'd require lots of out-of-bounds checking, and might not improve read- and maintainability... I'm uncertain.

WopsS: Not sure what you mean with `won't update`, but it won't be an overkill, you can find the index of the last entry with `IndexOf`, then assume everything before it is already correct.

There is two cases that comes to my mind now: 
1. `startIndex + 1 == list.Count -> startIndex = 0`, if this happen you can assume this is the last entry in the list and can safely start from 0.
3. An input like `[ mesh1, mesh2, mesh3, mesh3 ]` will cause issues with `IndexOf`. But maybe just assume this is wrong / intended and just append `mesh3` (penultimate value).

Anyway, my point here is that the speed of the algorithm can be improved.

manavortex: You're absolutely right. I have an idea how I can get that out of the loop. I'll sleep on it and hopefully implement it tomorrow :)

WopsS: Please do not sleep on it, it is fragile.

manavortex: If it was C...

manavortex: Pushed update, works. Please resolve and proceed at your discretion :))

manavortex:
Done with WopsS's review changes, could somebody trigger the pipeline maybe? 👀

PR #1046, opened 2022-12-30
Restore ExportFiles for Scripting and Add Export Settings

dragonzkiller:
Fixed:
- Refactored `wkit.ExportFiles` for scripting

Currently just a draft PR to track adding in export options. PR could be merged now to restore functionality without the additional feature

### DIFF
--- a/WolvenKit.App/Helpers/WKitUIScripting.cs
+++ b/WolvenKit.App/Helpers/WKitUIScripting.cs
@@ -1,18 +1,18 @@
 ﻿using System;
 using System.Collections;
-using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reactive;
 using MoreLinq;
 using Splat;
+using WolvenKit.App.ViewModels.Exporters;
 using WolvenKit.Common;
 using WolvenKit.Core.Interfaces;
 using WolvenKit.Functionality.Services;
 using WolvenKit.Modkit.Scripting;
+using WolvenKit.RED4.Archive;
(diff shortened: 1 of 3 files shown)


dragonzkiller:
Added in export settings with the option to select the exposed property name in the script via an attribute. By default, the attribute will keep the property name, but setting it to something else will change the name in the script. Only parameters that are built-in types or enums have been tagged since currently no type conversions exist for more advanced types.

Export settings for the following types (not all attributes) are now supported:
- xbm
- mesh
- wem
- opus
- mlmask
- anims
- morphtarget

For example: In the example below, `ImageType` is the overridden name for scripts that actually points to `MaterialUncookExtension` parameter for `Mesh` and `UncookExtension` parameter for `Xbm`. However, `LodFilter` points to the `LodFilter` parameter, since the attribute has no script name defined.

```js
wkit.ExportFiles(['base\\environment\\architecture\\common\\int\\int_nkt_apartment_a\\int_nkt_apartment_a_deco_pillar_l60_h400_aa.mesh',
    'base\\gameplay\\gui\\fullscreen\\loading\\wat_lch_dc_4k.xbm'],
    {
        Mesh: { ExportType: "WithMaterials", ImageType: "png", LodFilter: true, Binary: false},
        Xbm: { ImageType: "dds", Flip: false }
    }
);
```
See `ExportArgs.cs` for more information on how it was done.

dragonzkiller:
Oh I also forgot to mention that this also enables scripts to use both relative paths and absolute paths since the code checks for that now

rfuzzo (approved):

PR #1051, opened 2023-01-06
Fixed multiply assignment KeyUp event

### LINKED ISSUE #977: Project Explorer: key actions fire twice
**Describe the bug**
In the **Project Explorer**, calling a context menu option via keyboard shortcut will trigger the option twice. In case of delete, this will lead to error messages.

**To Reproduce**
1. Select a file in the process explorer
2. Press the "F2" key to rename
3. Two dialogues will be opened

or
1. Select a file in the process explorer
2. Press the "delete" key
4. Confirm the dialogue prompt
5. Console: An error will be thrown that the file couldn't be deleted because it doesn't exist

**Expected behavior**
The action should only fire once. First suspect: onkeyup and onkeydown are both firing, but I haven't looked into the source yet, so it can be anything.

**Desktop (please complete the following information):**
 - OS: Windows 10 x64
 - Version WolvenKit-8.7.0-nightly.2022-10-23

Onichan410:
closes #977

Add condition to add KeyUp event

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
@@ -363,6 +363,8 @@ namespace WolvenKit.ViewModels.Tools
             }
         }
 
+        public bool IsKeyUpEventAssigned { get; set; }
+
         /// <summary>
         /// Renames selected node.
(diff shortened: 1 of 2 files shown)


rfuzzo (approved):
looks good to me, thanks!

PR #1066, opened 2023-01-11
Implemented mesh name to MultiMesh missing bone error message

### LINKED ISSUE #902: Multi-mesh export with missing rig: error message should be more helpful
**Describe the bug**
When exporting a multi-mesh and encountering a missing rig in one of the additional meshes, the error message refers to the primary mesh. This is confusing to the user, because they may assume that they selected the wrong rig for their primary mesh or there is something wrong with it.

**To Reproduce**
Begin export of a multi-mesh rig. So, for example, export `t0_000_pma_base__full` (player body), with additional meshes `h0_001_ma_c__player` (player head) and `a0_001_pma_base_hq__full` (player arms). 

If you add only rig `man_base_deformations`, it will give you the following error: `base\characters\common\player_base_bodies\player_man_average\t0_000_pma_base__full.mesh - Bone: l_J_eye_lid_up_rowA_0_JNT not present in export Rig(s)/Import Mesh`

While perhaps obvious in hindsight (the reference to "eye_lid"), at a glance it is not clear what part of the multi-mesh export it is referring to, and it might send the user off on a chase for the wrong rig file.

**Expected behavior**
The error message should display what mesh is causing the issue - do not reference the primary mesh at all or make it clear it is an additional mesh. Re-writing the above error, it might appear such as:

`base\characters\common\player_base_bodies\player_man_average\t0_000_pma_base__full.mesh - In additional mesh h0_001_ma_c__player - Bone: l_J_eye_lid_up_rowA_0_JNT not present in export Rig(s)/Import Mesh`.

Syrrr:
Implemented mesh name to MultiMesh missing bone error message
Closes #902

Implemented:
- Multimesh file name in error message if rig is missing bones

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
@@ -261,7 +261,7 @@ namespace WolvenKit.Modkit.RED4.Tools
 
             return true;
         }
-        public bool ExportMultiMeshWithRig(List<Stream> meshStreamS, List<Stream> rigStreamS, FileInfo outfile, bool lodFilter = true, bool isGLBinary = true, ValidationMode vmode = ValidationMode.TryFix)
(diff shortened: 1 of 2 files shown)


rfuzzo (approved):
thanks, looks good to me

PR #1070, opened 2023-01-11
Implemented saveAs for wscript files (Issue #1028)

### LINKED ISSUE #1028: Save as on wscript just saves
**Describe the bug**
if you modify a wscript then do save as with the intent of saving your changes to a new script it just saves to the existing one. 

**To Reproduce**
create a script, edit it , Save it. edit it some more, save as and it will just save.

**Expected behavior**
should let you pick a new name and save to that so the original doesnt get overwritten.

**Desktop (please complete the following information):**
 - OS: Win 10
 - Version current main

Syrrr:
# Fix/wscript-saveas

Fixed:
- SaveAs acting as regular save

It now opens a proper SaveFileDialog and gives the opportunity to save the script with a new name and theoretically in a different folder (might not be good?). I also reopen the saved file on saving to make sure that the UI updates accordingly.

closes #1028

### DIFF
--- a/WolvenKit.App/ViewModels/Documents/WScriptDocumentViewModel.cs
+++ b/WolvenKit.App/ViewModels/Documents/WScriptDocumentViewModel.cs
@@ -123,6 +123,7 @@ public partial class WScriptDocumentViewModel : DocumentViewModel
 
         ContentId = path;
         FilePath = path;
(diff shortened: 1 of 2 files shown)


Simarilius-uk:
Thanks for fixing this :)

rfuzzo (approved):
thanks

PR #1081, opened 2023-01-14
Fix submesh order being determined by mesh name (MeshData in Blender) instead of node name (Object name in Blender)

### LINKED ISSUE #1036: Imported mesh's submeshes getting reordered
**Describe the bug**
If you merge submeshes in blender and then separate them either via selection or via material, even tho you rename the submesh names correctly, upon import the order of the submeshes will get reordered randomly so material will not be linked correctly. I also tried the "keep submesh order" option in import, but I got an exception when trying to do mesh preview which may be related to this bug: https://github.com/WolvenKit/WolvenKit/issues/1035

**To Reproduce**
1. I exported this mesh "t2_119_pma_jacket__suit" to blender
2. Either assign individual materials to the submeshes, merge them and then separate by material
Or, merge them all and separate by selection
3. Rename the submeshes to their correct names(so the order is correct) and export
4. Import the mesh
5. submeshes will be reordered (can check via mesh preview)

**Expected behavior**
It should keep the correct submesh order so that material can be mapped correctly. I'd like to merge submeshes because I can use sculpt mode to edit. 

Tested on nightly 2022-12-15

Syrrr:
# fix/submesh-order

This fixes the submesh order not being determined by object name in Blender, but rather the mesh name that is connected to the object. Those occasionally get mixed up when joining and separating.

Closes #1036

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
@@ -361,8 +361,10 @@ namespace WolvenKit.Modkit.RED4
             var Meshes = new List<RawMeshContainer>();
             foreach (var node in model.LogicalNodes)
             {
+                _loggerService.Info(node.Name);
                 if (node.Mesh != null)
                 {
+                    _loggerService.Info("belongs to " + node.Mesh.Name);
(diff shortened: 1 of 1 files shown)


seberoth (approved):
lgtm

PR #1115, opened 2023-01-25
Issue 1114 - allow non 1:1 mlmasks. 

kwekmaster:
Allow non 1:1 mlmasks to be imported thru WKit

Fixed:
- removed code based on the incorrect assumption that all mlmasks are 1:1. It is not.
- since the original code assumes that it is 1:1, need to also change the logic to determine power-of-two ness of the dimensions

Tested:
- Maintained previous behavior for 1:1 mlmasks
- Works with non 1:1 mlmasks
- Enforcement of the power of two constraint


rfuzzo (approved):
Looks good

PR #1121, opened 2023-01-28
fix: Sanitize RedRef Depot Path Hashes

### LINKED ISSUE #1062: [BOT] Uppercase handling
[Issue create with https://github.com/rfuzzo/LizzyFuzzy] 
 Author: Lizzy Fuzzy 
 Add some sort of uppercase handling to wkit: easiest is a warning popup on pack that some files/folders have uppercase letters

dragonzkiller:
Implemented:
- `RedRefEditor` (via the embedded `RedCNameEditor`) now sanitizes the hashes of the resource paths before hashing

Fixed:
- Fixes #1062, but more "correctly" than just handling uppercase letters as the entire path is sanitized

This adds the option `SanitizeHash="True"` to the `RedCNameEditor` in the XAML which will remove all leading and trailing characters that are one of `" ' \ /` as well as replacing any repeated slashes (forwards and backwards) to be a single backslash. Final path is then made all lowercase before being fed into the hashing algorithm.

### DIFF
--- a/WolvenKit.Common/RED4/CR2W/JSON/PrimitiveConverter.cs
+++ b/WolvenKit.Common/RED4/CR2W/JSON/PrimitiveConverter.cs
@@ -366,10 +366,10 @@ public class ResourceReferenceConverter : JsonConverter<IRedRef>, ICustomRedConv
                 case "DepotPath":
                 {
                     var converter = options.GetConverter(typeof(CName));
-                    if (converter is ICustomRedConverter conv)
+                    if (converter is CNameConverter conv)
                     {
                         reader.Read();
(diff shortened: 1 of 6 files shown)


dragonzkiller:
Modified `ResourceReferenceConverter` to use a modified `CNameConverter` to sanitize the paths upon importing a JSON file. It will perform the full sanitization like the game will, so if a modder uses a messy path, it'll be the "correct" on upon import

dragonzkiller:
After a discussion on Discord with Seb, I'm going to close this for now and change it over to a new type so we don't have to rely on `CName`

Issue #1029, opened 2022-12-08
Expose on Wolvenkit.CLI the parameter MaterialUncookExtension for the export WithMaterial


Neurolinked:
**Actual State**
Actually using the Wolvenkit.CLI executable for the extraction of the meshes with the option WithMaterials does not have a way to let you choose the texture formats for normals, masks layer and so on

**Compatibility with the CLI**
Actually Wolvenkit in the export option panel let you specify with **WithMaterial Setting** the format for the textures [**MaterialUncookExtension**]

**Adding the parameter**
If the parameter it's exposable, that will help others have the format they found confortable work with.
MLSB historically used DDS, now it's also support PNG, but if i can't specify an export format for the repository generation , i have to force the PNG one as default.

Neurolinked:
I'm trying just now the edit, but still the command I'm using wont work as intended, it was supposed to export mesh withmaterials with textures with dds format, but i still get only png files

The -r option it's just to try only a folder from the basegame4_appearance.archive

`WolvenKit.CLI.exe uncook -p "C:\Program Files (x86)\GOG Galaxy\Games\Cyberpunk 2077\archive\pc\content\basegame_4_appearance.archive" -r "^base.vehicles.+.mesh$" -o C:\Export\ --mesh-export-type WithMaterials --uext dds`

Neurolinked:
the last edit for the console version of wolvenkitten didn't obtain the results expected:

I've used this command

`.\WolvenKit.CLI.exe uncook -p "C:\Program Files (x86)\GOG Galaxy\Games\Cyberpunk 2077\archive\pc\content\basegame_4_appearance.archive" -r "^base.vehicles.+.mesh" -o C:\CPArk\181\ --mesh-export-type WithMaterials --uext dds`

Even after specifying the parameter `--mesh-export-material-repo <mesh-export-material-repo>`     Location of the material repo, if not specified, it uses the outpath.
Does not extract glb and it's materials telling me that i have to specify the Depot path

[https://ibb.co/zbysNDf](Uncook fail I)
[https://ibb.co/m9KK8gY](Uncook fail II)
[https://ibb.co/QDKqPdw](Uncook command help)

seberoth:
@Neurolinked sry for the late reply. Use `MeshOnly` instead of `WithMaterials`

PR #1124, opened 2023-01-30
fix: Apply meshes uext arg to uncook command

seberoth:
Apply meshes uext arg to uncook command

Implemented:
- Applied #1029 also to uncook


rfuzzo (approved):
Thanks

PR #1138, opened 2023-02-03
Initial conversion from CName resource paths to new ResourcePath type

dragonzkiller:
Fixed:
- Paths are now fully sanitized to match how the game 

NOTE: This will show up to the end user that all the paths are being forced to be lowercase and without leading slashes, etc.


dragonzkiller:
Removing and resetting this to handle latest changes on `main`

Issue #1135, opened 2023-02-02
Error Importing Files


XilaMonstrr:
Cannot import a wide variety of texture files. I get the following errors.
They are particularly prolific in: environment/decorations/containers

[error (1).txt](https://github.com/WolvenKit/WolvenKit/files/10571901/error.1.txt)
[error.txt](https://github.com/WolvenKit/WolvenKit/files/10571902/error.txt)

rfuzzo:
thanks, I made a fix so that it imports, but both your images are TexGroup Data... So, I'm not sure what to expect

PR #1140, opened 2023-02-03
Texture Import fixes

rfuzzo:
# Texture Import fixes

Fixed:
- don't compress images when compression is "none"

<Additional notes>
#1135


seberoth (approved):
lgtm

PR #1141, opened 2023-02-04
Mesh preview robustness

manavortex:
# Mesh preview robustness

## (Hopefully) Changed: 
Clicking the "Generate materials for appearance" button will now recreate the materials. This was previously done only once (abort condition: rendered), and the user would be required to re-open the file if they changed anything.

Fixed: Exception breaking mesh preview

If more materialEntries than material definitions are present, there would be exceptions, and the mesh preview would break. 
New behaviour: it will now stop parsing when it has reached the end of the list. Appearances with missing materials won't be listed. (Potentially it might be better to assign a default material? Maybe screaming pink with "UNDEFINED" on it? I don't know how, though.)

Will create a follow-up ticket.


rfuzzo (approved):
Thank you! This looks all good.

PR #1147, opened 2023-02-08
use naudio implementation from naudio github

rfuzzo:
# use naudio implementation from naudio github

Implemented:
- proper audio implementation from scratch using Naudio samples
- three different waveform visualizers

Fixed:
- fixed various audio player issues
- proper mvvm implementation

<Additional notes>


seberoth (approved):
lgtm

PR #1151, opened 2023-02-10
✨ `a[rchive]:<string>` Refinement Matches on Archive Path

effs:
Implemented:
- Case-insensitive string match on archive path for Asset Browser search

It's a refinement, so can be used in any valid refinement position (but first probs usually makes sense):

`a:auska > .mesh > gold` == `.mesh > a:auska > gold`


rfuzzo (approved):
"^th(?:ank)?:(?<s>.*)$"

Issue #1107, opened 2023-01-22
WolvenKit NEXT Roadmap


rfuzzo:
Roadmap to v9 of WolvenKit

## Major issue tracking
- Make all projects nullable (https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references)
  - #1072 
- Clean up the use of Dependency Injection across the whole solution
  - #823 
  - #1100 
- Remove all WPF dependencies from WolvenKit.app.csproj
  - #1102 
  - #1065
- Use WINUI 3 for the UI
- [ ]  Move almost all logic out of WolvenKit.csproj  into Wolvenkit.app.csproj
- [ ]  Create a prototype WINUI project

  
  
## Stuff that's not necessarily dependent on WINUI but nice to have, or that solves itself
- Add in-app tooltips
  - #1010
  - #808 
- Clean up all binding errors
  - #1067

rfuzzo:
probably not gonna happen

PR #1157, opened 2023-02-11
wopss: remove all locator-pattern DI

rfuzzo:
# wopss: remove all locator-pattern DI

Fixed:
- remove all locator-pattern DI and replace with factories

<Additional notes>
#1107


WolvenKit.App/Factories/ChunkViewmodelFactory.cs
```diff
@@ -0,0 +1,101 @@
...
+        _projectManager = projectManager;
+        _gameController = gameController;
+        _archiveManager = archiveManager;
+        _appViewModel = appViewModel;
+        _tweakDbService = tweakDbService;
+        _locKeyService = locKeyService;
+        _parserService = parserService;
+    }
+
+    public ChunkViewModel ChunkViewModel(IRedType rootChunk, ReferenceSocket socket)
```
WopsS, line 52: I would call these methods `Create`.

WolvenKit.App/Factories/IChunkViewmodelFactory.cs
```diff
@@ -0,0 +1,13 @@
+﻿using WolvenKit.App.Models.Nodify;
+using WolvenKit.App.ViewModels.Shell;
+using WolvenKit.RED4.Types;
+using WolvenKit.App.ViewModels.Documents;
+
+namespace WolvenKit.App.Factories;
+
+public interface IChunkViewmodelFactory
```
WopsS, line 8: I would make a generic class named `IFactory<T> where T : class` and then implement it in the concrete class above.

This `IFactory<T>` can use `ActivatorUtilities.CreateInstance` in a `Create(params object[] args)` method to pass the service provider and extra params.

But as a questions, are these params necessary for the object to be constructed or can these be added later?

WopsS (changes requested):
Only looked at the first few files for now, will continue tomorrow.

PR #1161, opened 2023-02-16
wopss: document factories

rfuzzo:
# document factories 2.0

Implemented:
- remove Locator pattern from wkit.app

<Additional notes>
#1107


WolvenKit.App/Factories/ChunkViewmodelFactory.cs
```diff
@@ -0,0 +1,106 @@
...
+using WolvenKit.Common;
+using WolvenKit.Core.Interfaces;
+using WolvenKit.RED4.CR2W;
+using WolvenKit.RED4.Types;
+using WolvenKit.App.ViewModels.Documents;
+using WolvenKit.App.Services;
+
+namespace WolvenKit.App.Factories;
+
+public interface IFactory<T> where T : class
```
WopsS, line 14: What I meant previously was that this should be something like https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger?view=dotnet-plat-ext-7.0. With a default method for creation, then where service is registered you can have a generic concrete class that creates an instance of `T`.

Issue #1167, opened 2023-02-24
Handle import/export of topology from/to morphtarget


manavortex:
For custom body mods without UV framework, it seems to be necessary to change the morphtarget's 3d model along with the body model. Currently, Wolvenkit does not allow importing over a morphtarget, which leaves Noesis as the only working alternative.

There are instances (as with the female body mesh) where the morphtarget is in the same folder as the .mesh and has the same name, with the file extension as the only difference. These cases need to be considered and automated somehow (suffix for the morphtarget glb?)
![image](https://user-images.githubusercontent.com/965785/221164821-fd9e31dc-e9a2-42e3-b1c1-87be4af3a779.png)

## Import
Wolvenkit needs to allow importing a morphtarget's glb export back into the same file. 

## Export
**UV Maps**
When exporting morphtargets, the UV maps are Y-flipped compared to the mesh export. This should be the same.

### Additional information
Link to Discord conversation: https://discord.com/channels/717692382849663036/808086068023918603/1078583164320100382

effs:
This is fixed, right? Or was there still some weirdness remaining?

manavortex:
<img width="753" alt="image" src="https://user-images.githubusercontent.com/965785/230669347-69851d0c-2452-49bf-9d41-d6723c6f7c1a.png">

PR #1170, opened 2023-02-25
⚓ Import/Export Bimsnes

effs:
🐞 .morphtarget export/import work and roundtrip, because...

🐞 `FileSystemArchive` fixed so that it can be used to access current project assets
✨ `Cp77Project.AsArchive()` can be used to directly get an archive of the currently loaded project

📣 Try to produce some easier-to-read output from failed exports

Addresses #1167


effs:
Or, well, they don't _fail_ in an obvious way anymore, at any rate.

rfuzzo (approved):
thanks

PR #1192, opened 2023-03-04
1146: prevent adding folders to themselves by accident

### LINKED ISSUE #1146: Project Explorer: Drag and drop improvements
# Duplicate file in same folder

I would like to drag and drop a file to its own folder with the Ctrl key held down to create a duplicate (filename_001.extension or something). Right now, when I do this, nothing happens - presumably, the tool checks if target and destination are the same.

# Prevent self-drag-and-drop for folders

By contrast, dragging and dropping a folder onto itself will create a full recursive copy inside the folder, even without the Ctrl key being held down. _That has happened to me countless times and I'm annoyed enough to write a ticket now, but I'm probably a target group of one here._

If a folder is dragged onto/into itself, then nothing should happen - drag and drop move or copy should only take effect on a higher level than its own root path.

manavortex:
# 1146: prevent adding folders to themselves by accident

Implemented:
- Dragging-and-dropping a folder on itself will no longer copy the folder to the folder [(1146)](https://github.com/WolvenKit/WolvenKit/issues/1146)

_yo dawg, we heard you liked folders_

closes #1146

### DIFF
--- a/WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
+++ b/WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
@@ -496,6 +496,12 @@ namespace WolvenKit.Views.Tools
                                 files.Add(sourceFile.FullName);
                             }
                         }
+                        // 1146: addresses "prevent self-drag-and-drop" 
+                        if (files.Count > 0 && files[0] == targetDirectory)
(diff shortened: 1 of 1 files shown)


rfuzzo (approved):
lgtm

PR #1196, opened 2023-03-06
📛 Export/Import Type Tagged `.morphtarget` GLTF

effs:
Minimal, unstructured scaffolding to import/export type-tagged GLTFs. Since this is more involved than is ideal, export is just supported for `.morphtarget` (so that it is distinct from a `.mesh` export.) 

You can, however, use file naming to avoid having to select the import format for GLTFs. 

# BUT

Rather than try to surgically fix things here and there, I think the best option is to:

1. Get rid of the separate Mesh, WithRig and Multimesh exports and just treat everything as cases of Multimesh in a single code path (which is what import does on the return trip.) This also needs a better UI, but should work transparently anyway. -> Doing this in #1197 

2. Try to separate import/export data handling from the filesystem ops, either by returning data streams from the exporters, or by passing in a fs context that _cannot_ be changed downstream.

3. _Then_ maybe doing something more.

### DIFF
--- a/WolvenKit.App/ViewModels/Exporters/TextureExportViewModel.cs
+++ b/WolvenKit.App/ViewModels/Exporters/TextureExportViewModel.cs
@@ -201,6 +201,7 @@ public partial class TextureExportViewModel : ExportViewModel
 
                 meshExportArgs.Archives.Insert(0, proj.AsArchive());
 
+                // Should check for depo here instead of dtl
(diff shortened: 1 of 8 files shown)


rfuzzo (approved):
Thanks, looks good (can't comment on the actual mesh logic though). 

In the future, it would be great if commits did not contain any emojis please!

PR #1198, opened 2023-03-08
⚙️ Settings Migration Refactor for V3 Plus Experimental Setting

effs:
Made the migration logic a bit more uniform.

The experimental setting isn't used yet, need to thread it through.

Related to #1197

### DIFF
--- a/WolvenKit.App/Services/SettingsDto.cs
+++ b/WolvenKit.App/Services/SettingsDto.cs
@@ -6,6 +6,9 @@ namespace WolvenKit.App.Services;
 
 public class SettingsDto : ISettingsDto
 {
+    private const int _CurrentSettingsVersion = 2;
+
+
     // Deserialize
     public SettingsDto()
     {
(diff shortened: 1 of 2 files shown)


seberoth:
Why bump to v3 if you just add an option?
Just add it to the Dto and set the default value to true.
Also cuurently it is only set to true on migration and not for new users

rfuzzo (approved):
Thanks, we also have appsettings.json for feature flags btw

rfuzzo (changes requested):
changing to request changes to address Seb's points. I like the more general version migration though. 
but: 
1. Add true to the default value
2. Don't bump the version

effs:
> Thanks, we also have appsettings.json for feature flags btw

Oh, I had missed that, thanks! This was meant more for user-togglable features. Alternative in this case might be to persist the MeshArgs selection instead, but that'll still reset between restarts…

effs:
Can live without this

PR #1202, opened 2023-03-08
Add a save image item to the right click menu on the image preview.

Simarilius-uk:
Implemented:
- Added a right click option to Save Image to the image preview pane to allow embedded images to be saved

Fixed:
- Having to tell people to screengrab the Interface to get the images.

<Additional notes>
I don't consider this a solution to the larger 'how to handle embeddeds' that we've been discussing for a while, its just a sticking plaster to deal with my immediate main issue. Will  be fine with it being ripped out if a proper solution appears.

### DIFF
--- a/WolvenKit/Views/Documents/RDTTextureView.xaml.cs
+++ b/WolvenKit/Views/Documents/RDTTextureView.xaml.cs
@@ -7,9 +7,11 @@ using System.Windows;
 using System.Windows.Input;
 using System.Windows.Media;
 using System.Windows.Media.Imaging;
+using System.Windows.Shapes;
 using Microsoft.Win32;
 using ReactiveUI;
 using Splat;
+using WolvenKit.App.Services;
 using WolvenKit.App.ViewModels.Documents;
 using WolvenKit.Common.Model.Arguments;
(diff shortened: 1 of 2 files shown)


WolvenKit/Views/Documents/RDTTextureView.xaml.cs
```diff
@@ -222,6 +246,11 @@ public void ResetZoomPan(object sender, RoutedEventArgs e)
             pan.Y = 0;
             end.X = 0;
             end.Y = 0;
+            var temp_filename = Path.Combine(ISettingsManager.GetTemp_OBJPath(), "embedded_preview.png");
```
seberoth, line 249: What is this?

seberoth (changes requested):

WolvenKit/Views/Documents/RDTTextureView.xaml.cs
Simarilius-uk: something I should have deleted. I made the reset zoom menu item do the save first then added the other one, forgot to delete that code.

seberoth (approved):

PR #1204, opened 2023-03-09
Fix: Simplify Project UI Layout Loading, Which May Avoid File Lock?

effs:
Dunno, completely resistant to debugging.

And also life is gray and boring.

### DIFF
--- a/WolvenKit/Views/Shell/DockingAdapter.xaml.cs
+++ b/WolvenKit/Views/Shell/DockingAdapter.xaml.cs
@@ -42,7 +42,6 @@ namespace WolvenKit.Views.Shell
         private bool _stateChanged;
 
         private bool _usingProjectLayout = false;
-        private bool _hadLoadedProject = false;
         private readonly bool _debuggingLayouts = false;
         private readonly bool _useAppdataStorage = true;
 
--- a/WolvenKit/Views/Shell/DockingAdapter.xaml.cs
+++ b/WolvenKit/Views/Shell/DockingAdapter.xaml.cs
@@ -474,22 +474,23 @@ namespace WolvenKit.Views.Shell
                 if (!_hadLoadedProject && ItemsSource is ObservableCollection<IDockElement> oc)
                 {
                     _hadLoadedProject = true;
-                    //oc.Clear();
                 }
                 var layoutPath = Path.Combine(_viewModel.ActiveProject.ProjectDirectory, "layout.xml");
                 if (File.Exists(layoutPath))
                 {
-                    var reader = XmlReader.Create(layoutPath);
-                    var Debugging_A = PART_DockingManager.LoadDockState(reader);
-                    Trace.WriteLine(Debugging_A);
-                    reader.Close();
+                    Locator.Current.GetService<ILoggerService>().Info($"Trying to load project layout from {layoutPath}...");
+
+                    var loadedProjectLayout = PART_DockingManager.LoadDockState(layoutPath);
+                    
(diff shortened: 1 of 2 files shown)


WolvenKit/Views/Shell/DockingAdapter.xaml.cs
```diff
@@ -474,22 +474,23 @@ public void OnActiveProjectChanged()
...
-                    //oc.Clear();
                 }
                 var layoutPath = Path.Combine(_viewModel.ActiveProject.ProjectDirectory, "layout.xml");
                 if (File.Exists(layoutPath))
                 {
-                    var reader = XmlReader.Create(layoutPath);
-                    var Debugging_A = PART_DockingManager.LoadDockState(reader);
-                    Trace.WriteLine(Debugging_A);
-                    reader.Close();
+                    Locator.Current.GetService<ILoggerService>().Info($"Trying to load project layout from {layoutPath}...");
```
seberoth, line 481: Why not store the logger in a local var and resuse this?

WolvenKit/Views/Shell/DockingAdapter.xaml.cs
```diff
@@ -474,22 +474,23 @@ public void OnActiveProjectChanged()
...
+
+                    var loadedProjectLayout = PART_DockingManager.LoadDockState(layoutPath);
+                    
+                    Locator.Current.GetService<ILoggerService>().Info($"...Project layout load returned {loadedProjectLayout}, who knows what that means? Did it work?");
```
seberoth, line 485: > who knows what that means
It means, that the action was successfull...

seberoth (changes requested):

WolvenKit/Views/Shell/DockingAdapter.xaml.cs
```diff
@@ -468,28 +468,31 @@ public void OnActiveProjectChanged()
                 return;
             }
 
+            var logger = Locator.Current.GetService<ILoggerService>();
+
             try
             {
                 // need to also handle if files have been modified (probably elsewhere, though)
                 if (!_hadLoadedProject && ItemsSource is ObservableCollection<IDockElement> oc)
```
rfuzzo, line 476: unrelated to the PR, but this does nothing

WolvenKit/Views/Shell/DockingAdapter.xaml.cs
effs: Fun fact: it doesn't

WolvenKit/Views/Shell/DockingAdapter.xaml.cs
```diff
@@ -147,35 +149,38 @@ private void LoadLayout()
...
+                logger.Debug($"...Project layout load returned {isSuccessful}");
+
                 reader.Close();
-                Locator.Current.GetService<ILoggerService>().Info($"Loaded default layout");
             }
-            catch (Exception ex)
+            catch (Exception e)
             {
-                Trace.WriteLine(ex.Message);
+                logger.Error($"Project layout load error: {e.Message}");
```
effs, line 183: This is not **project** layout

WolvenKit/Views/Shell/DockingAdapter.xaml.cs
seberoth: Yes, it does. Just keep in mind, that it is only loading the state. 
So if the current "docked" windows are different then the previous saved, it will also return false.

Issue #1208, opened 2023-03-11
Scene files don't show up in graph editor


yna2jd:
**Describe the bug**
I don't know anything about the graph editor, but if you click on a scene file and try to look at the graph editor tab it's just totally blank

**To Reproduce**
Open up a scene file
click graph editor

**Expected behavior**
Editing graphs?

**Version (please complete the following information):**
 - Version 8.8.1

**Screenshots**
![Untit12led](https://user-images.githubusercontent.com/123516015/224462784-a33034bc-b349-4ef7-b534-09b5a9e2cced.png)

yna2jd:
oops ignore the white in the screenshot

seberoth:
Yeah, that feature was never finished for scene files and the dev who started working on it doesn't work on wkit anymore.

So we should probably hide this tab in future

rfuzzo:
It's hidden now, but we can keep this open since it would be cool to have a scene graph viewer. related to: #575

seberoth:
Done iwth b850ba2

PR #1211, opened 2023-03-11
General fixes and stuff

### LINKED ISSUE #1206: some buffers not being serialised to JSON but written as bytes
**Describe the bug**
Have come across a couple of structures that wkit shows the internal structure of, and which can be written to JSON as individual nodes, but when the whole file is converted to json they are written as a string of bytes.
Behaviour observed on base\environment\decoration\small_shops\stands\mannequin_real\generated\mannequin_real_man_a_doggy.app
trying to access the appearance[0] parametersBuffer
and
base\worlds\03_night_city\sectors\_generated\collisions\03_night_city.geometry_cache
trying to access the bufferTableSectors

**To Reproduce**
1. Add either of the above files to a project
2. Right click and convert to json in the project tree
3. look at the JSON and the mentioned sections are just Byte arrays.
4. Open the file and export the section to JSON and you get the individual sub sections/properties etc.

**Expected behavior**
JSON with the properties visible rather than the byte strings. 

**Version (please complete the following information):**
 - checked with main.

### LINKED ISSUE #1207: LocKeys
TweakDB Browser can't show a single LocKey. When I click on a string with a LocKey on it, the string becomes empty.
LockKey Browser shows two columns. When I type a number in there, nothing happens too.
I don't understand what is wrong. I have the latest .Net framework runtime installed. I believe this have started when Cyberpunk got an update.

seberoth:
**!!! This bumps JSON to v0.0.4 !!!**

Implemented:
- Switch LocKey Browser Primaray column from Text to Numeric. Partially fixes #1207 
- Buffer serialization for `SerializationDeferredDataBuffer` and `SharedDataBuffer`. Fixes #1206 

Fixed:
- NodeRef hash calculation for paths with aliases

Disabled unused graph viewer for scene files. See #1208

### DIFF
--- a/WolvenKit.RED4/Types/Pools/NodeRefPool.cs
+++ b/WolvenKit.RED4/Types/Pools/NodeRefPool.cs
@@ -1,13 +1,12 @@
 ﻿using System.Collections.Concurrent;
 using System.Text;
 using WolvenKit.Common.FNV1A;
-using WolvenKit.Core.Interfaces;
 
 namespace WolvenKit.RED4.Types;
 
 public static class NodeRefPool
 {
-    public static ILoggerService? Logger;
+    //public static ILoggerService? Logger;
 
     private static readonly ConcurrentDictionary<string, ulong> s_pool = new();
     private static readonly ConcurrentDictionary<ulong, string> s_poolReverse = new();
--- a/WolvenKit.RED4/Types/Pools/NodeRefPool.cs
+++ b/WolvenKit.RED4/Types/Pools/NodeRefPool.cs
@@ -1,13 +1,18 @@
 ﻿using System.Collections.Concurrent;
 using System.Text;
 using WolvenKit.Common.FNV1A;
+using WolvenKit.Core.Interfaces;
 
 namespace WolvenKit.RED4.Types;
 
 public static class NodeRefPool
 {
+    public static ILoggerService? Logger;
+
     private static readonly ConcurrentDictionary<string, ulong> s_pool = new();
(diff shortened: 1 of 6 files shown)


Simarilius-uk:
confirm its fixed 1206 for the app files, but it seems to still not have fixed the BufferTables in the geometry_cache

seberoth:
> confirm its fixed 1206 for the app files, but it seems to still not have fixed the BufferTables in the geometry_cache

Yep, thats "intended" since we have no writer for it yet. So even if you change it in wkit, the changes will not be saved.

WolvenKit.RED4/Types/Pools/NodeRefPool.cs
```diff
@@ -1,14 +1,19 @@
 ﻿using System.Collections.Concurrent;
 using System.Text;
 using WolvenKit.Common.FNV1A;
+using WolvenKit.Core.Interfaces;
 
 namespace WolvenKit.RED4.Types;
 
 public static class NodeRefPool
 {
+    public static ILoggerService? Logger;
```
rfuzzo, line 10: is this assigned anywhere?

rfuzzo (changes requested):
I trust you on the json converters, the rest looks good. one comment

WolvenKit.RED4/Types/Pools/NodeRefPool.cs
seberoth: No, not yet. Should have commented that out for now...

rfuzzo (approved):
danke

PR #1214, opened 2023-03-11
Garment Support fixes

engres01:
Garment Support attribute check and type fix

Implemented:
- Removed unnecessary check for color attribute before writing to garment support parameters, since importing garment support is currently driven by a checkbox in import settings

Fixed:
- Fixed an incorrect integer type being written to a parameter in the off chance that the cap color attribute has a value


manavortex (commented):
LGTM. I have no idea what it does, but this won't break anything… I think.

seberoth (approved):
Mana said it all...

PR #1216, opened 2023-03-12
fix: Mesh preview not updating after switching tabs

seberoth:
Fixed:
- Mesh preview not updating after switching tabs. Thanks @manavortex for the help


manavortex (commented):
Confirmed working :)

rfuzzo (approved):

PR #1217, opened 2023-03-13
Remove WindowsAPICodePack

dragonzkiller:
Removes WindowsAPICodePack and moves `CommonFileDialog` to new `FolderPicker` and `OpenFileDialog`. Also removes the pre-compiled DLLs as they are no longer needed

Implemented:
- FolderPicker from https://stackoverflow.com/questions/11624298/how-do-i-use-openfiledialog-to-select-a-folder/66187224#66187224


rfuzzo (approved):
very nice! thank you!

PR #1235, opened 2023-03-20
Add the selected mesh to the export list for multimesh 

Simarilius-uk:
Implemented:
Currently the Multimesh has a list labelled 'additional meshes' that you have to include the selected mesh into to get it to export. Minor change that adds the selected mesh to the export list if its not in there. Added to both old and new export flows, but cant test the old one as its not working for other reasons.

Fixed:
- never got round to filing a bug, but the fact the additional list wasn't additional at all but the whole export list, and the selected file wasn't automatically added.


rfuzzo (approved):
Looks good

PR #1273, opened 2023-04-15
JSON fixes

seberoth:
JSON fixes

Fixed:
- Deserialization of CArrayFixedSize / CStatic
- Deserialize CEnum without values
- Deserialize CVariant with uninitialized handles
- Serialize/Deserialize SharedDataBuffer


Tests/WolvenKit.Utility/DumpInfo.cs
```diff
@@ -336,23 +337,34 @@ private void MergeStrings(bool factOnly)
                         ext = hashService.GetGuessedExtension(dc.Hash);
                     }
 
-                    if (ext == ".questphase")
+                    if (ext == ".questphase" || ext == ".scene")
```
rfuzzo, line 340: We could use the enum.tostring here (actually, we should make a pass and refactor all hardcoded extensions to enums)

rfuzzo (approved):

PR #1279, opened 2023-04-18
fix: Calculate Morphtarget Bounds Cumulatively on Import

effs:
The influences can be combined. Whether they're treated as cumulative or mutually exclusive is completely external semantics. We must assume cumulative in order to create a big enough bounding box and quantization space.


effs:
It's quite possible I'm overlooking something here (since I already overlooked _this_) so more eyes would be nice.

seberoth (approved):
lgtm

PR #1285, opened 2023-04-20
Fixes #1284, LxrsFooter performance improvements and naming scheme enforcing

### LINKED ISSUE #1284: Enforce consistent naming scheme and performance improvements to LxrsFooter class
**Describe the bug**
The `LxrsFooter.cs` class in the `WolvenKit.RED4` solution currently has some inconsistent naming an duplicate code.
Also `FileInfos` filed can be made an auto property as the backing field would still be immutable but its contents can still be modified from extern like in the current implementation. 
Consider exposing FileInfos as immutable object to avoid any unintended side effects by exposing the list directly. This could be achieved thought following modifications with a backing field again:
- `IReadOnlyList<string>`
- `.AsReadOnly()`

paulober:
### LxrsFooter performance improvements and naming scheme enforcing

Implemented:

Some minor performance improvements to `LxrsFooter.cs` by: 
- Splitting duplicate code
- Use a single encoding instance instead of retrieving it multiple times

Other changes:
- Renamed some variables
- Changed `FileInfos` property to an auto-property for a more idiomatic C# way

Fixed:
- Fixes #1284

### DIFF
--- a/WolvenKit.RED4/Archive/Base/LxrsFooter.cs
+++ b/WolvenKit.RED4/Archive/Base/LxrsFooter.cs
@@ -11,16 +11,11 @@ public class LxrsFooter
 
     public const uint s_magic = 0x4C585253; //1397905484;
     private const uint s_version = 1;
+    
+    public List<string> FileInfos { init; get; }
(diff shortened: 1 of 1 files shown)


seberoth (approved):
Looks good to me. Thanks for the PR

PR #1290, opened 2023-04-22
Fix #1289, Possibly null reference

### LINKED ISSUE #1289: Dereference of a possibly null reference
### Description:
`Export.cs` contains a dereference of a possibly null reference in following line:
https://github.com/WolvenKit/WolvenKit/blob/2a361c5021315bb603d47232f6d6ccac1057c7f3/WolvenKit.Modkit/RED4/Export.cs#L63

paulober:
### Possibly null reference

Implemented:
- `?` to avoid possibly null reference
- Naming scheme for some variables
- Added missing param tag in XML comment

Fixed:
- Fixes #1289


seberoth (approved):
lgtm

PR #1295, opened 2023-05-21
reset layout: trying to laod DockStates.xml from appdata if it exists

manavortex:
# Reset Layout will consider DockStates.xml in AppData

Fixed:
- reset layout will now use DockStates.xml in AppData if file exists and the setting is true.
Confirmed working on my end. Won't fix the project layout randomly breaking, but at least you can now fall back to a clean state without having to remove the file by hand.

Maybe there's something wrong with `DockStatesDefault.xml`?


manavortex:
Seb did it better :P

PR #1297, opened 2023-05-24
more file validation in WScript

manavortex:
file validation in WScript

Implemented:
- Added file validation for app, ent, workspot and anims
- changed file structure of WScript folder: Wolvenkit subfolder will always overwrite, everything else only if not present
- moved Logger to subdirectory
- made Logger a bit more null-proof

The original logger won't be deleted, so this shouldn't break existing scripts


seberoth (approved):
lgtm

PR #1303, opened 2023-06-02
Improved JSON number handling

seberoth:
Implemented:
- Saving int64/uint64 as string
- Reading int64/uint64 as string or number (to keep it backward compatible)


rfuzzo (approved):
Thanks! Lgtm

PR #1305, opened 2023-06-13
Feature/fbx export import

rfuzzo:
# fbx export import with redmod

Implemented:
- fbx export with redmod 
- fbx import with redmod 

Fixed:
- some small cleanup and renaming of vms

### DIFF
--- a/WolvenKit.App/Models/MaterialXmlModel.cs
+++ b/WolvenKit.App/Models/MaterialXmlModel.cs
@@ -1,243 +1,167 @@
-﻿using System.Xml.Serialization;
+﻿using System;
+using System.ComponentModel;
+using System.Xml.Serialization;
 
 namespace WolvenKit.App.Models;
 
 // NOTE: Generated code may require at least .NET Framework 4.5 or .NET Core/Standard 2.0.
 /// <remarks/>
-[System.SerializableAttribute()]
-[System.ComponentModel.DesignerCategoryAttribute("code")]
-[System.Xml.Serialization.XmlTypeAttribute(AnonymousType = true)]
-[System.Xml.Serialization.XmlRootAttribute(Namespace = "", IsNullable = false, ElementName = "mesh")]
+[Serializable()]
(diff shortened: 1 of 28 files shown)


WolvenKit.Common/Model/Arguments/ExportArgs.cs
```diff
@@ -194,13 +194,13 @@ public class MeshExportArgs : ExportArgs
...
+        /// Choose exporter
         /// </summary>
         [Category("Experimental")]
-        [Display(Name = "Use New Mesh Exporter")]
-        [Description("Use New Mesh Exporter Logic. Turn off if you run into issues.")]
+        [Display(Name = "Choose Exporter")]
+        [Description("Choose the exporter type. Use default if you run into issues.")]
         [WkitScriptAccess("ExperimentalNewMeshExporter")]
-        public bool ExperimentUseNewMeshExporter { get; set; } = false; // TODO: Remove this when the new exporter is stable.
+        public MeshExporterType MeshExporter { get; set; }
```
seberoth, line 203: Maybe change category to `General` or something like that?
Also please change the ScriptAttribute (or delete the name param from it)

WolvenKit.App/Models/MaterialXmlModel.cs
```diff
@@ -0,0 +1,243 @@
+﻿using System.Xml.Serialization;
+
+namespace WolvenKit.App.Models;
+
+// NOTE: Generated code may require at least .NET Framework 4.5 or .NET Core/Standard 2.0.
```
seberoth, line 5: Not sure about this, but since it probably doesn't change that often (won't be regenerated) maybe we could/should apply naming conventions?

WolvenKit.Common/Model/Arguments/ExportArgs.cs
rfuzzo: oh right sorry I forgot

PR #1325, opened 2023-07-06
RTTI improvements

seberoth:
Implemented:
- Made abstract RTTI classes abstract
- Updated some default values


rfuzzo (approved):
yeah totally checked all 4539 files. approving 👍

PR #1326, opened 2023-07-07
wscript updates

Simarilius-uk:
Updated the scripts to all have the add from project, and to load project files over base game if available. Added a seperate list for types to be added/exported if embedded so you can export mlmasks/xbms only if embedded.

Added a script to allow you to export the embedded from a file without having to export an ent/sector.


seberoth (approved):
lgtm

PR #1346, opened 2023-08-18
FIX: Provide Root Explicitly for Single-Joint Rig Export

effs:
Fixed:
- Single-joint models (like many hats) are exported with correct rigging

SharpGLTF's ancestor finder assumes a single node is the root node, so it calculates the inverse bind matrix incorrectly.

https://github.com/vpenades/SharpGLTF/blob/b23b5ce225637f38a00be67f4641c991ed36435d/src/SharpGLTF.Core/Schema2/gltf.Skin.cs#L238


effs:
Should probably test this some :thinkies: and I'm not really sure it's a great solution.. but there's a lot of cruft on this path so it may be better left cleaned up in the grand unified exporter theory.

rfuzzo (approved):
looks fine, but I'm not qualified to review mesh stuff :D

PR #1349, opened 2023-08-22
file validation: don't warn about empty depotPaths for componentOverr…

manavortex:
- removed warning for empty depot path for pathOverrides (they'll target the player ent in this case)


rfuzzo (approved):
approved

PR #1354, opened 2023-09-03
treeview: recursive folding/unfolding with shift key held down

manavortex:
in tree view, opening/closing a node with shift key will also affect its children. 

Implemented extra checks for some objects in mesh/app, as these are the ones where not having it annoyed me the most. 
Feedback welcome, I have actually no idea what the correct way is of getting this helper class into the model.

### DIFF
--- a/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
@@ -25,6 +25,7 @@ using WolvenKit.App.Models.Nodify;
 using WolvenKit.App.Services;
 using WolvenKit.App.ViewModels.Dialogs;
 using WolvenKit.App.ViewModels.Documents;
+using WolvenKit.App.ViewModels.Tools;
 using WolvenKit.Common;
 using WolvenKit.Common.Services;
 using WolvenKit.Core.Extensions;
@@ -49,6 +50,7 @@ namespace WolvenKit.App.ViewModels.Shell;
 public partial class ChunkViewModel : ObservableObject, ISelectableTreeViewItemModel, INode<ReferenceSocket>
 {
     private readonly IChunkViewmodelFactory _chunkViewmodelFactory;
+    private readonly ChunkViewModelTools _chunkViewModelTools = new ChunkViewModelTools();
     private readonly IDocumentTabViewmodelFactory _tabViewmodelFactory;
     private readonly ILoggerService _loggerService;
     private readonly ISettingsManager _settingsManager;
@@ -93,7 +95,9 @@ public partial class ChunkViewModel : ObservableObject, ISelectableTreeViewItemM
         ITweakDBService tweakDbService,
         ILocKeyService locKeyService,
         Red4ParserService parserService,
-        ChunkViewModel? parent = null, bool isReadOnly = false)
+        ChunkViewModel? parent = null,
+        bool isReadOnly = false
+    )
     {
         _chunkViewmodelFactory = chunkViewmodelFactory;
         _tabViewmodelFactory = tabViewmodelFactory;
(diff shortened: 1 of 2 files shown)


WolvenKit.App/ViewModels/Tools/ChunkViewModelTools.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using DynamicData.Binding;
+using WolvenKit.App.ViewModels.Shell;
+using WolvenKit.RED4.Types;
+
+namespace WolvenKit.App.ViewModels.Tools;
+
+public class ChunkViewModelTools
```
rfuzzo, line 10: public static class ChunkViewModelExtensions

WolvenKit.App/ViewModels/Tools/ChunkViewModelTools.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    public void SetChildExpansionStates(ChunkViewModel? node, bool isExpanded)
```
rfuzzo, line 12: public static void SetChildExpansionStates(this ChunkViewModel? node, bool isExpanded)

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -49,6 +50,7 @@ namespace WolvenKit.App.ViewModels.Shell;
 public partial class ChunkViewModel : ObservableObject, ISelectableTreeViewItemModel, INode<ReferenceSocket>
 {
     private readonly IChunkViewmodelFactory _chunkViewmodelFactory;
+    private readonly ChunkViewModelTools _chunkViewModelTools = new ChunkViewModelTools();
```
rfuzzo, line 53: delet

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -196,6 +201,11 @@ public partial class ChunkViewModel : ObservableObject, ISelectableTreeViewItemM
     {
         if (IsExpanded && !_propertiesLoaded)
             CalculateProperties();
+
+        if (IsShiftBeingHeld)
+        {
+            _chunkViewModelTools.SetChildExpansionStates(this, IsExpanded);
```
rfuzzo, line 207: this.SetChildExpansionStates(IsExpanded);

rfuzzo (changes requested):
could do it like this I guess, with extension methods.

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
manavortex: Is it enough to have <ClassName>Extension.cs to have it immediately available?

rfuzzo: doesn't depend on the name, it is available when you have a static method with this as first argument for that class

manavortex: Ahh, yeah - I wanted to make it a static class, but that would make HotReload not work for it, so I instantiated it. Any advice about that?

rfuzzo: huh, I didn't know that... no, no idea. that's weird but okay we can leave it nonstatic then I guess

manavortex: I don't want to instantiate this for every single tree view node that I'm touching >.< I'll check later if I can make a factory class or something

rfuzzo (approved):

PR #1401, opened 2023-10-22
feat: copy tweakxl to clipboard

rfuzzo:
# copy tweakxl to clipboard

Implemented:
- copy tweakxl to clipboard menu option


seberoth (approved):
lgtm

PR #1408, opened 2023-10-30
fix: do not clear pasted chunk clipboard

### LINKED ISSUE #1286: Copy/Paste chunks: Improve user experience by not clearing the clipboard after pasting components
**Bug Description**
When a component object (for example entSkinnedMeshComponent) is copied and pasted, the **clipboard is cleared**, which **produces the neccessity to copy an object again when there's a need to paste it multiple times**. In my opinion and drawing on my own experience with using WolvenKit as a modder, not clearing the clipboard would provide more friendly user experience, especially in situations, when for example an object needs to be pasted in multiple locations (like different appearances of a character).

**To Reproduce**
1. Go to Asset Browser and **add any appearance file to the project** (for example search for 'panam.app')
2. Expand *appearances > (any appearance) > components*, right-click any componend and **click 'Copy From Array/Buffer'**
3. Right click any components array and **click 'Paste Into Array/Buffer'**
4. Right click any components array again
5. Notice, that **pasting options are now unavailable**

**Expected behavior**
The clipboard is not cleared and pasting options are still available. This behavior is especially useful, when an object needs to be pasted in multiple locations, like multiple components arrays.

**Version:**
 - 8.9.1

rfuzzo:
# do not clear pasted chunk clipboard

please check if that's a problem

Fixed:
- do not clear pasted chunk clipboard

<Additional notes>
closes #1286

### DIFF
--- a/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
@@ -1515,19 +1515,29 @@ public partial class ChunkViewModel : ObservableObject, ISelectableTreeViewItemM
             {
                 return;
             }
-
-            if (PropertyType.IsAssignableTo(typeof(IRedArray)))
+            if (RedDocumentTabViewModel.CopiedChunk is IRedCloneable irc)
             {
-                if (!CreateArray())
+                if (PropertyType.IsAssignableTo(typeof(IRedArray)))
                 {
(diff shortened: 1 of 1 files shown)


seberoth (changes requested):
Please paste a (deep)copy (if applicable) then. Else every copy will be the same reference.

rfuzzo:
ah thanks, that makes sense

rfuzzo:
Updated branch

seberoth (approved):

PR #1436, opened 2023-11-19
holding shift: copy absolute instead of relative path

manavortex:
Could maybe someone who knows how to do that hook up the GUI so that the descriptor changes? It's currently too much for my brane.

Otherwise, I'm fine with having an un-documented feature :D


rfuzzo (approved):
neat! thanks

Issue #1447, opened 2023-11-25
Anims: Raw SIMD Animation Reimport Converted Incorrectly


effs:
Currently at least non-quantized-rotation SIMD animbuffer is causing issues where the player's transform in the world becomes invalid somehow. Quantized-rotation SIMD buffers *do* seem to work.

With current export setup, the animation does look correct in Blender.

It's possible that this is rather an issue with e.g. events, tracks, or fallback frames being incorrect when converting to a compressedbuffer than an actual transform problem.

Example: `base\animations\weapon\firearms\shotgun\constitutional_tactician\pwa_w_shotgun__constitutional_arms_tactician.anims`, `first_equip` animation.

PR #1448, opened 2023-11-25
Anims: Experimental SIMD Animation Support & Blender Fixes

effs:
## SIMD (Experimental)

SIMD animations are heavyweight and detailed. There aren't that many of them, but they tend to be the showy ones. The buffer format is distinctly optimized for SIMD semantics (obvs.)

- Export with as much data as possible. This means transforms etc. for every joint for every frame, so it's a lot.
- Import converts to CompressedBuffer instead. This works for *most* SIMD buffers (see #1447).

- [ ] Need to implement SIMD buffer construction still (may also solve #1447)

## Blender Enhancements

- Additive animations are exported with local transform included so they show up more manageable in Blender until there's support for loose anims on that side. Import strips the local again.
- Misc. fixes

## GLTF

- Extra data schema V3 improves (?) V2
- Migration mechanism for versions
- Migration from V2 to V3 on import

## Internals

- Create new data structures instead of replacing them on import. This should also enable partial animset replacements (with non-updated either kept or deleted) and adding new animations.


kmirea:
Can i comment here ? :P

I was poking at one of the additive weapon anims, noticed that a particular sequence, was followed by another two of same name (affixed with _0 and _1) respectively and same buffer length,  is it possible these should be combined into one import. once i actually able to parse them and see transforms might have an answer,,but just a thought (for reference; pwa_procedural_weapon.anims => add_body_hit_270_upperbody, add_body_hit_270_upperbody_01, add_body_hit_270_upperbody_02)

effs:
>  is it possible these should be combined into one import

Couldn't say, they don't look obviously connected (but they can ofc be additive on top of one another.) Other kinds with number suffixes seem to be variants of the same base animation.

PR #1449, opened 2023-11-26
Multi mesh materials

Simarilius-uk:
Multimesh Material JSON Appearance fix
Implemented:
- added the mesh names to the submeshes rather than just the numerical prefix (only on multimesh)
- added the mesh names to the appearances in the material json so that if multimesh import has 2 meshes with the same named appearances both still get imported.

Fixed:
- Materials being available for multimesh output

<Additional notes>
Still need to do a bit of work on the Blender side to make appearance selection work, as they dont have the same name the app file has for them (but I'm not sure this has worked previously either)

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
@@ -213,7 +213,7 @@ namespace WolvenKit.Modkit.RED4.Tools
             return true;
         }
 
-        public static MeshesInfo GetMeshesinfo(rendRenderMeshBlob rendMeshBlob, CMesh? cMesh = null)
+        public static MeshesInfo GetMeshesinfo(rendRenderMeshBlob rendMeshBlob, CMesh? cMesh = null, string meshname="")
         {
             var meshesInfo = new MeshesInfo(rendMeshBlob.Header.RenderChunkInfos.Count);
             var redQuantScale = rendMeshBlob.Header.QuantizationScale;
@@ -349,7 +349,7 @@ namespace WolvenKit.Modkit.RED4.Tools
                         ArgumentNullException.ThrowIfNull(m);
                         materialNames[e] = m;
                     }
-                    meshesInfo.appearances.Add(app.Name + $"{i}", materialNames);
+                    meshesInfo.appearances.Add(Path.GetFileNameWithoutExtension(meshname) + '_' + app.Name + $"{i}", materialNames);
                 }
             }
 
@@ -568,7 +568,7 @@ namespace WolvenKit.Modkit.RED4.Tools
                 meshContainer.materialNames = new string[info.appearances.Count];
 
                 
-                var Mesh_apps = info.appearances.Keys.Select(key => Path.GetFileNameWithoutExtension(meshname)+'_' + key).ToList();
+                //var Mesh_apps = info.appearances.Keys.Select(key => Path.GetFileNameWithoutExtension(meshname)+'_' + key).ToList();
 
                 var apps = info.appearances.Keys.ToList();
 
@@ -576,7 +576,7 @@ namespace WolvenKit.Modkit.RED4.Tools
                 {
                     // Null-proof materials
                     var d = info.appearances[apps[e]];
-                    info.appearances[Mesh_apps[e]] = InitializeDefaultMaterial(info.meshCount, d.NotNull());
+                    info.appearances[apps[e]] = InitializeDefaultMaterial(info.meshCount, d.NotNull());
                     meshContainer.materialNames[e] = d[index];
                 }
 
--- a/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
@@ -349,7 +349,14 @@ namespace WolvenKit.Modkit.RED4.Tools
                         ArgumentNullException.ThrowIfNull(m);
                         materialNames[e] = m;
                     }
-                    meshesInfo.appearances.Add(Path.GetFileNameWithoutExtension(meshname) + '_' + app.Name + $"{i}", materialNames);
+                    if (meshname.Length > 0)
+                    {
+                        meshesInfo.appearances.Add(Path.GetFileNameWithoutExtension(meshname) + '_' + app.Name + $"{i}", materialNames);
+                    }
+                    else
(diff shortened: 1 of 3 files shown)


WolvenKit.Modkit/RED4/Uncook.cs
```diff
@@ -785,26 +785,26 @@ public bool ExportMultiMeshWithRig(Dictionary<Stream, string> meshStreamS, List<
 
             var expMeshes = new List<RawMeshContainer>();
             var matData = new List<MatData>();
-            foreach (var meshStream in meshStreamS.Keys)
+            foreach (var meshStream in meshStreamS)
```
dragonzkiller, line 788: what if this were like this?

```csharp
foreach (var (meshStream, meshName) in meshStreamS)
```

We could then leave the rest of the code untouched for all the `.Key` and instead of `.Value` it would be easier to understand that it's actually the name of the mesh.

WolvenKit.Modkit/RED4/Uncook.cs
```diff
@@ -785,26 +785,26 @@ public bool ExportMultiMeshWithRig(Dictionary<Stream, string> meshStreamS, List<
...
-                MeshTools.UpdateSkinningParamCloth(ref Meshes, meshStream, cr2w);
+                var Meshes = MeshTools.ContainRawMesh(ms, meshesinfo, meshExportArgs.LodFilter,  ulong.MaxValue,  false, meshStream.Value );
+                MeshTools.UpdateSkinningParamCloth(ref Meshes, meshStream.Key, cr2w);
 
                 MeshTools.WriteGarmentParametersToMesh(ref Meshes, cMesh, meshExportArgs.ExportGarmentSupport);
 
                 var meshRig = MeshTools.GetOrphanRig(cMesh);
 
-                MeshTools.UpdateMeshJoints(ref Meshes, expRig, meshRig, meshStreamS[meshStream]);
+                MeshTools.UpdateMeshJoints(ref Meshes, expRig, meshRig, meshStreamS[meshStream.Key]);
```
dragonzkiller, line 807: If the previous were implemented, this would become:

```csharp
MeshTools.UpdateMeshJoints(ref Meshes, expRig, meshRig, meshName);
```

dragonzkiller (commented):
Not requesting any particular changes, but some suggestions. Seb can weigh in as well.

WolvenKit.Modkit/RED4/Tools/MeshTools.cs
```diff
@@ -550,12 +557,26 @@ public static List<RawMeshContainer> ContainRawMesh(MemoryStream gfs, MeshesInfo
                     meshContainer.indices[i] = gbr.ReadUInt16();
                 }
 
+                if (meshname.Length>0)
+                {
+                    meshContainer.name = mergeMeshes ?
+                                        info.appearances.Keys.FirstOrDefault("default") :
+                                        Path.GetFileNameWithoutExtension(meshname)+"_submesh_" + Convert.ToString(index).PadLeft(2, '0') + "_LOD_" + info.LODLvl[index];
```
effs, line 564: Can you pull this in from the meshinfo? Or maybe throw the naming convention in a function? Might be over-defensive but could prevent accidentally changing one but not the other.

WolvenKit.Modkit/RED4/Tools/MeshTools.cs
```diff
@@ -550,12 +557,26 @@ public static List<RawMeshContainer> ContainRawMesh(MemoryStream gfs, MeshesInfo
...
-                    info.appearances.Keys.FirstOrDefault("default") :
-                    "submesh_" + Convert.ToString(index).PadLeft(2, '0') + "_LOD_" + info.LODLvl[index];
+                }
+                else
+                {
+                    meshContainer.name = mergeMeshes ?
+                                        info.appearances.Keys.FirstOrDefault("default") :
+                                        "submesh_" + Convert.ToString(index).PadLeft(2, '0') + "_LOD_" + info.LODLvl[index];
+                }
+
```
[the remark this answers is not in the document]
Simarilius-uk: As far as I can tell its an arbitrary naming convention that wkit decided to use for the submeshes in an object at some point. the whatever is just the mesh index of the submeshes.

WolvenKit.Modkit/RED4/Uncook.cs
Simarilius-uk: sounds good, I felt ugly when I was doing it. Was just trying to get it to work. Will make this change.

WolvenKit.Modkit/RED4/Tools/MeshTools.cs
Simarilius-uk: The mesh info doesnt have the parent object name as part of it, thats part of the issue. Could change the meshinfo, but tbh it was gonna start effecting a lot more of the code and I was already borderline out of my depth. Am aware this only fixes the default export path at the minute, your experimental one goes a different route right?

WolvenKit.Modkit/RED4/Uncook.cs
dragonzkiller: > that would need to be (meshStream.Key, meshname) wouldnt it?

no. since `meshStreamS` (bad name btw lol) is a dictionary, doing a foreach loop does this:
```csharp
foreach (var (key, value) in dict)
```
so the first var becomes the key value which we we were previously iterating over

WolvenKit.Modkit/RED4/Tools/MeshTools.cs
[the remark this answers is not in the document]
dragonzkiller: My concern with this area is just ensuring we're being as agnostic as possible with our mesh naming conventions. Blender/glTF is what we support the most at the moment, but there's been talk about trying to bring back other file formats and I'd like to avoid writing specifically for one program/format over another.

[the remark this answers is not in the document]
Simarilius-uk: its no less agnostic than it was before this patch, its just more useable with multimesh.

WolvenKit.Modkit/RED4/Tools/MeshTools.cs
effs: I don't even remember what it does tbh, but now that you mention it, it's probably better to try to make the change in a more holistic overhaul yeah.

PR #1469, opened 2023-12-06
always flip xbms on export

rfuzzo:
remove flip args

Fixed:
- always flip xbms on export
- remove manual flip options

<Additional notes>


seberoth (approved):
Seems to work :D

PR #1489, opened 2023-12-30
better guessing of default import settings from xbm's filename

manavortex:
# .xbm import settings

Minor refactoring for better guessing of default texture import settings from file name


rfuzzo (approved):
lgtm

PR #1500, opened 2024-01-07
Quest/Scene editor

seberoth:
Quest/Scene editor

Implemented:
- Quest/Scene editor

Should be "stable" enough to be in nightly.
Also can start working on node editing then...


rfuzzo (approved):
manually approved (I totally reviewed all 116 files yeah)

PR #1515, opened 2024-01-14
guess premultiplyAlpha from filename: consider overlay/alpha

manavortex:
# PremultiplyAlpha guessing

**Implemented:**
- Added "overlay" and "alpha" to file name partials that should default to transparency on

### DIFF
--- a/WolvenKit.Common/RED4/CommonFunctions.cs
+++ b/WolvenKit.Common/RED4/CommonFunctions.cs
@@ -269,10 +269,10 @@ public static class CommonFunctions
         }
 
         // what else could we possibly check for?
-        var keywordsToCheck = new List<string> { "decal", "icon", "overlay", "alpha" };
+        var partialsToCheck = new List<string> { "decal", "icon", "icons", "overlay", "alpha" };
 
         // check if we have a match (filename has already been toLower)
-        return keywordsToCheck.Any(keyword => fileName.Contains(keyword));
+        return partialsToCheck.Any(partial => fileName.Contains($"_{partial}") || fileName.Contains($"{partial}_"));
     }
 
 }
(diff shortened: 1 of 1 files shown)


WolvenKit.Common/RED4/CommonFunctions.cs
```diff
@@ -268,7 +269,10 @@ public static bool ShouldPremultiplyAlpha(string fileName)
         }
 
         // what else could we possibly check for?
-        return fileName.Contains("decal") || fileName.Contains("icon");
+        var keywordsToCheck = new List<string> { "decal", "icon", "overlay", "alpha" };
+
+        // check if we have a match (filename has already been toLower)
+        return keywordsToCheck.Any(keyword => fileName.Contains(keyword));
```
rfuzzo, line 275: Is that check too broad? it would guess that a file called "electricbatonalpha_mk2_d.xbm" has premultiply alpha, even though "alpha" is just part of the item name. 

Maybe it needs to check for `_decal` or `_alpha`? That wouldn't really solve the problem obviously ("electric_baton_alpha_mk2_d.xbm"). 

Not sure. Maybe move the full file name check before the  ends_with check?
here:       // corresponds to greyscale textures - roughness and maps

manavortex: Good point - users could still set it by hand since it'll only suggest the default value, but yeah, this check is too broad.
Which do you prefer? 

1. contains $"_{keyword}"
2. endsWith(keyword)

rfuzzo: hmm, "endsWith" I think. Is that how it usually works? Would I, as a texture modder, call my file "electric_baton_alpha_mk2_decal.xbm" or "electric_baton_alpha_mk2_overlay.xbm"?

If yes, then let's go with an endsWith check yeah. 👍

manavortex:
Whenever someone says "do people normally", I think of this: 
image

I'll go with contains

rfuzzo (approved):
lgmt

PR #1520, opened 2024-01-18
Option to Override Mesh Name with Node Name when Importing for Compatibility

### LINKED ISSUE #1519: Regression: submesh order not preserved on import
When importing the mesh with the Nightly from 15-01, the submesh order is not preserved. Importing the same .glb file into different mesh files will import them persistently incorrect (different for each mesh).

This does not happen with the Nightly from 09-01

[import_order_bug_source.zip](https://github.com/WolvenKit/WolvenKit/files/13954821/import_order_bug_source.zip)

effs:
To support older files without modders having to fix their mesh naming in Blender, new import option that forces using the node name (object name in Blender) instead of the actual mesh name.

Screenshot 2024-01-18 082336

There's a warning printed if the names don't match either way.

Screenshot 2024-01-18 083313

## Breaking

- Error if a mesh is referenced by multiple nodes, unless someone comes up with a valid use case (in which case the entire import logic needs to be reviewed.)

closes #1519

### DIFF
--- a/WolvenKit.Common/Model/Arguments/ImportArgs.cs
+++ b/WolvenKit.Common/Model/Arguments/ImportArgs.cs
@@ -99,6 +99,22 @@ namespace WolvenKit.Common.Model.Arguments
         private GltfImportAsFormat _importFormat = GltfImportAsFormat.Mesh;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


seberoth (approved):
lgtm

PR #1521, opened 2024-01-18
setting previous import behaviour as default

manavortex:
# setting previous import behaviour as default

maintaining backwards compatibility by setting the flag for mesh import behaviour to true as default

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
@@ -551,7 +551,7 @@ namespace WolvenKit.Modkit.RED4
 
             if (mesh.Name != node.Name)
             {
-                _loggerService.Info($"Mesh name `{mesh.Name}` does not match parent node name `{node.Name}`. Using {(args.OverrideMeshNameWithNodeName ? "NODE" : "MESH")} name.");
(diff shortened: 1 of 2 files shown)


WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
```diff
@@ -551,7 +551,7 @@ private RawMeshContainer GltfMeshToRawContainer(Mesh mesh, GltfImportArgs args)
 
             if (mesh.Name != node.Name)
             {
-                _loggerService.Warning($"Mesh name `{mesh.Name}` does not match parent node name `{node.Name}`. Using {(args.OverrideMeshNameWithNodeName ? "NODE" : "MESH")} name.");
+                _loggerService.Info($"Mesh name `{mesh.Name}` does not match parent node name `{node.Name}`. Using {(args.OverrideMeshNameWithNodeName ? "NODE" : "MESH")} name.");
```
effs, line 554: This should remain a `Warning`, it's for a reason.

effs (approved):

WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
manavortex: OK, I'll put it back :)

rfuzzo (approved):
sure

PR #1527, opened 2024-01-21
Feature/new type selector

seberoth:
# Feature/new type selector

**Implemented:**
- new type selector


WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -1052,18 +1057,39 @@ private async Task AddItemToArray()
...
+                            new("Color", "Color description", typeof(CColor)),
+                            new("CpuNameU64", "", typeof(CName)),
+                            new("Cube", "Reference to cube xbm", typeof(CResourceReference<ITexture>)),
+                            new("DynamicTexture", "Reference to dtex file", typeof(CResourceReference<ITexture>)),
+                            new("FoliageParameters", "Reference to fp file", typeof(CResourceReference<CFoliageProfile>)),
+                            new("Gradient", "Reference to gradient file", typeof(CResourceReference<CGradient>)),
+                            new("HairParameters", "Reference to hp file", typeof(CResourceReference<CHairProfile>)),
+                            new("MultilayerMask", "Reference to mlmask file", typeof(CResourceReference<Multilayer_Mask>)),
+                            new("MultilayerSetup", "Reference to mlsetup file", typeof(CResourceReference<Multilayer_Setup>)),
+                            new("Scalar", "", typeof(CFloat)),
```
manavortex, line 1083: this could be "a decimal number". Also, maybe `Integer` is missing here? Or can we use `Scalar` interchangeably and the engine will cope?

manavortex (approved):
LGTM! (Cursory review and a manual text - it seems to work… no doubt someone will find edge cases)

rfuzzo (approved):

PR #1533, opened 2024-01-27
Mesh material name index fix

manavortex:
# Mesh material name index fix

Mesh materials will now actually check if the material definition's index property is matching before just assuming that everything is fine if the array index is fine


rfuzzo (approved):
mesh

PR #1539, opened 2024-01-28
New default layout suggestion

manavortex:
# New default layout suggestion

Asset browser is now pinned by default
Import and export tool are visible by default

What do y'all think?

image


seberoth:
Not sure if we should assume 2560x1440 as default resolution.
I think, many users still have FullHD (like me) and then the actual file area is rather small.
grafik

manavortex:
It could probably benefit from some re-scaling for smaller displays, yeah.

effs:
AB benefits more from height than width, Editor from both. And since you're less likely to use Export and Import at the same time, they can be tabbed to save space. So I'd still put AB in a column on the right stacked with either E/I or Preview below. Gives a bit more flexibility on width.

manavortex:
Most hardcore users whose layouts I've seen use Import/Export simultaneously, and I don't think we benefit much from height because you'd probably use the filter by <15 items anyway, wouldn't you? 

Open to suggestions though, push one?

effs:
But default is not for hardcore users or those who know how to search ultra-efficiently, default is for My First Mod :)

manavortex:
Fair point :)

rfuzzo (approved):
why not

PR #1548, opened 2024-02-01
Animgraph ragdoll nodenames

manavortex:
# Node information for animgraph ragdoll

This has been requested and tested by sketchcritic - there were no exceptions that they reported :)

### DIFF
--- a/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
@@ -16,7 +16,6 @@ using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.DependencyInjection;
 using CommunityToolkit.Mvvm.Input;
 using DynamicData.Binding;
-using Microsoft.ClearScript.JavaScript;
 using Microsoft.Win32;
 using WolvenKit.App.Controllers;
 using WolvenKit.App.Extensions;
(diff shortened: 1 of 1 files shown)


WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -10,14 +11,15 @@
 using System.Runtime.Serialization;
 using System.Text.Json;
 using System.Threading.Tasks;
-using System.Windows;
+using System.Windows.Forms;
```
rfuzzo, line 14: why do we need a dependency on Winforms here?

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -42,9 +44,12 @@
 using static WolvenKit.App.ViewModels.Dialogs.DialogViewModel;
 using static WolvenKit.RED4.Types.RedReflection;
 using CKeyValuePair = WolvenKit.RED4.Types.CKeyValuePair;
+using Clipboard = System.Windows.Clipboard;
 using IRedString = WolvenKit.RED4.Types.IRedString;
 using Mat4 = System.Numerics.Matrix4x4;
+using OpenFileDialog = Microsoft.Win32.OpenFileDialog;
```
rfuzzo, line 50: this seems to be not needed

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -42,9 +44,12 @@
...
 using Quat = System.Numerics.Quaternion;
+using SaveFileDialog = Microsoft.Win32.SaveFileDialog;
```
rfuzzo, line 52: this seems to be not needed

rfuzzo (changes requested):
Seems to be fine, just some usings have snuck in?

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
manavortex: Rider auto-corrected the existing `System.Windows` to `System.Windows.Forms`, narrowing down the dependency. Awfully nice of it :D

rfuzzo (approved):
seems fine?

PR #1549, opened 2024-02-01
Draft: make importExportArgs adjust themselves from file path

manavortex:
# make importExportArgs adjust themselves from file path

Not sure if this is a good or a terrible idea

### DIFF
--- a/WolvenKit.Common/Model/Arguments/ExportArgs.cs
+++ b/WolvenKit.Common/Model/Arguments/ExportArgs.cs
@@ -25,6 +25,7 @@ namespace WolvenKit.Common.Model.Arguments
     /// </summary>
     public abstract class ExportArgs : ImportExportArgs
     {
+        
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


WolvenKit.Common/Model/Arguments/ExportArgs.cs
```diff
@@ -301,6 +303,17 @@ public class MeshExportArgs : ExportArgs
         /// </summary>
         /// <returns>String</returns>
         public override string ToString() => $"GLTF/GLB | Export Type : {meshExportType} | Lod filter : {LodFilter} | Is Binary : {isGLBinary}";
+
+        public override ImportExportArgs AdjustDefaultFromFilepath(string baseFile)
+        {
+            // We don't want LOD filters to be active by default for weapon meshes 
+            if (baseFile.Contains("weapon", StringComparison.CurrentCultureIgnoreCase))
```
rfuzzo, line 310: I think this is too dangerous and it's also hidden. 

If we do this it should 
1. be behind a toggle switch in the UI
2. needs some description in the UI
3. needs a warning in the UI that this is active

rfuzzo (changes requested):

WolvenKit.Common/Model/Arguments/ExportArgs.cs
manavortex: Could absolutely implement it as abstract but I didn't want to change 28 inheriting classes without touching base first :D That would make it not-hidden, at least.

And yeah, I noticed in hindsight that this doesn't only affect the default settings (or does it? This should only become active once, when the object is created…)

rfuzzo:
This can be merged if we show on the UI that the export settings have been set based on some factors.

E.g. lower left-hand corner on the export tool: show an info png ℹ️ with a tooltip "guessing best settings for your file..." if we set the settings based on the filename

manavortex:
If you implement the GUI, I can do the rest.

PR #1550, opened 2024-02-01
inkWidget Improvements

jackhumbert:
# inkWidget Improvements

* Allows custom dynamic classes inheriting from `inkIWidgetController` & `inkWidgetLogicController`
  * Class must exist in mod's redscripts or otherwise registered with the RTTI
  * Modifies `Handle` creation type selector to allow entering the custom class name (base type is dependent on the handle)
  * Adds context menu option for editing class name
  * Allows adding, removing, and renaming of properties (only `inkWidgetReference` types for the time being, but any should be supported) of dynamic classes
* Allows editing of `inkWidgetReference`
  *  Allows copy/pasting from/to regular `Handle<inkWidget>`
* Adds a hook that copies `inkWidgetLibraryItem.package` to `inkWidgetLibraryItem.packageData` on saving `.inkwidget` files, and hides `packageData` from the UI
  * The only difference between `package` & `packageData` in the game is `package` is used for synchronous spawning, and `packageData` is (generally) used for asynchronous spawning (@psiberx, 2022)
  * `inkWidget.backendData` is often only found in `packageData`, and is not created upon copy - this doesn't seem to affect anything that I've seen

# Misc Fixes

* Fixes TweakXL context menu visibility in TreeView


seberoth (approved):
lgtm

PR #1558, opened 2024-02-04
#1557 overwrite with gamefile

### LINKED ISSUE #1557: Replace with original not working for multiselect in flat view
When trying to overwrite multiple files in flat view, only the first file will get exported.

manavortex:
# Fix "replace with original"

**Additional notes:**
closes #1557

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
@@ -405,7 +405,7 @@ public partial class AssetBrowserViewModel : ToolViewModel
     /// Browse the left side folder tree to the folder containing the selected item. (e.g. for after searching)
     /// </summary>
     private bool CanBrowseToFolder() => RightSelectedItem is RedFileViewModel &&
-                                        _archiveManager.GetGameFile(RightSelectedItem!.FullName) is not null;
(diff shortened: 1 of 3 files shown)


WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
```diff
@@ -404,29 +404,33 @@ public void UpdateSearchInArchives()
     /// <summary>
     /// Browse the left side folder tree to the folder containing the selected item. (e.g. for after searching)
     /// </summary>
-    private bool CanBrowseToFolder() => RightSelectedItem is RedFileViewModel;
+    private bool CanBrowseToFolder() => RightSelectedItem is RedFileViewModel &&
+                                        _archiveManager.GetGameFile(RightSelectedItem!.FullName) is not null;
```
seberoth, line 408: Not sure what this should fix.
`RightSelectedItem` is either set via the AB itself and so should be in the Archives or it is set via `ShowFile` which already check this itself.
Also if you are using _archiveManager.GetGameFile() you should probably exclude the project check.

manavortex: You're right, I fixed it in the next commit.

The fix itself is L433, which occasionally threw exceptions while I was browsing through ad textures

PR #1576, opened 2024-02-20
Setting default compression mode for normal map

manavortex:
# SWIZZLE EVERYTHING

Setting default compression mode for normal map
resolved a few todos


rfuzzo (approved):
manually approved. let's see if it breaks anything

PR #1579, opened 2024-02-25
renumber child index properties

manavortex:
# Renumber child index properties for definition arrays

Super useful, I will be using the shit out of this. Display is strictly conditional and will only display for the right node type.
https://github.com/WolvenKit/WolvenKit/assets/965785/653913fe-4263-4371-915d-dd366dd4542f


rfuzzo (approved):
renumbered approval properties

PR #1581, opened 2024-02-28
mesh material: more fallback upon export failures, better logging

manavortex:
I hope this didn't break the singleton

### DIFF
--- a/WolvenKit.Core/Services/SerilogWrapper.cs
+++ b/WolvenKit.Core/Services/SerilogWrapper.cs
@@ -6,14 +6,9 @@ namespace WolvenKit
 {
     public class SerilogWrapper : ILoggerService
     {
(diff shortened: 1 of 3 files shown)


seberoth (changes requested):
Pls don't do that. Pass it down like everywhere else.
Would create a second instance. Also won't work with CLI

manavortex:
> Pls don't do that. Pass it down like everywhere else. Would create a second instance. Also won't work with CLI

Alright, that was what I didn't know :) How _do_ I pass it down? I followed the chain two or three hops, and didn't find any loggers

seberoth:
`ModTools` has `_loggerService`
Add it to `MaterialExtractor` (also as `_loggerService`) and pass it into its constructor

seberoth (approved):
lgtm

PR #1595, opened 2024-03-10
mesh import: better error message

manavortex:
# mesh import: better error message

This is the #1 import error that makes people ask on Discord, and it's so easy to circumvent :D 

image


rfuzzo (approved):
Better approval message

PR #1596, opened 2024-03-10
split chunkviewmodel desc+value

manavortex:
# split chunkviewmodel desc+value

put them into their own partial classes because I don't want @seberoth to have merge conflicts all the time <3 

Aura is currently testing this, and I've been giving it a spin myself. So far, nothing broke.


rfuzzo (approved):
seems to have no conflicts with Seb's cvm branch, so we can merge it probably

PR #1603, opened 2024-03-17
Import export tools bugfixes

manavortex:
# Import/Export tools

- Morphtargets will now have a checkbox (disabled by default) to export their texture files
- Pasting import/export settings will now refresh the property display of changed items
- Fixed the annoying multiselect bug where the active item would get deselected
- .anims will now export to .anims.glb (and automatically import to Animation)
- minor refactoring in import tool to allow better type switching from file name

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/ImportExportItemViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/ImportExportItemViewModel.cs
@@ -21,10 +21,8 @@ public abstract partial class ImportExportItemViewModel : ObservableObject, ISel
         Properties.PropertyChanged += Properties_PropertyChanged;
     }
 
-    private void Properties_PropertyChanged(object? sender, System.ComponentModel.PropertyChangedEventArgs e)
-    {
+    private void Properties_PropertyChanged(object? sender, System.ComponentModel.PropertyChangedEventArgs e) =>
         PropertiesDisplay = Properties.ToString();
-    }
 
     public string BaseFile { get; set; }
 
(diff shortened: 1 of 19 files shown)


WolvenKit.App/ViewModels/Tools/ImportExportItemViewModel.cs
```diff
@@ -45,11 +43,7 @@ private void Properties_PropertyChanged(object? sender, System.ComponentModel.Pr
 
     public void SetProperties(ImportExportArgs args)
     {
-        Properties.PropertyChanged -= Properties_PropertyChanged;
```
seberoth, line 48: Not sure why you removed that.
Added that last week (e66d87e) to reflect the changes in the file info column and preserve the binding.

manavortex: Are those changes merged back? Because it didn't work for me, and putting the listener back makes it work O_O

seberoth: Yeah, its in main. And it worked for me.

rfuzzo (changes requested):
Moved comment by @Simarilius-uk here for visibility: This will break the plugin and the wscript.

ref: 1 comment on commit cb330c0 (animations will export to)

WolvenKit.App/ViewModels/Tools/ImportExportItemViewModel.cs
manavortex: Maybe I broke it in another spot because the way it is is working for me right now O_O Let's talk on Discord

rfuzzo (approved):
approved after changes in bleder plugin have been made

PR #1608, opened 2024-03-23
implemented "search and replace" in editor view. seems to work???

manavortex:
# Search&Replace

Fixed an exception on startup by null-checking, while I was at it


manavortex:
This seems to have broken the mesh preview somehow (or maybe it's my mesh?), not sure how and why and wtf

rfuzzo (approved):
thanks!

PR #1611, opened 2024-03-25
fixed export error for appearances without chunks

manavortex:
# fixed export error for appearances without chunks

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
@@ -582,8 +582,8 @@ namespace WolvenKit.Modkit.RED4.Tools
                 for (var e = 0; e < apps.Count; e++)
                 {
                     // Null-proof materials
-                    var d = info.appearances[apps[e]];
-                    info.appearances[apps[e]] = InitializeDefaultMaterial(info.meshCount, d.NotNull());
+                    var d = InitializeDefaultMaterial(info.meshCount, info.appearances[apps[e]]);
+                    info.appearances[apps[e]] = d;
                     meshContainer.materialNames[e] = d[index];
                 }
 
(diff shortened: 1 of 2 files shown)


WolvenKit.Modkit/RED4/Tools/MeshTools.cs
```diff
@@ -581,9 +581,10 @@ public static List<RawMeshContainer> ContainRawMesh(MemoryStream gfs, MeshesInfo
...
                 for (var e = 0; e < apps.Count; e++)
                 {
-                    // Null-proof materials
-                    var d = info.appearances[apps[e]];
-                    info.appearances[apps[e]] = InitializeDefaultMaterial(info.meshCount, d.NotNull());
+                    var appearance = apps[e];
+                    // Null-proof materials by making sure we have one material instance per submesh
+                    var d = info.appearances[appearance];
+                    info.appearances[apps[e]] = InitializeDefaultMaterial(info.meshCount, ref d);
                     meshContainer.materialNames[e] = d[index];
```
seberoth, line 588: Why does `InitializeDefaultMaterial` now both return the value AND the same data as ref?
Isn't it the same as `meshContainer.materialNames[e] = info.appearances[apps[e]][index];` now?

manavortex: Now that you say it. I'll fix it

manavortex: Updated

seberoth (approved):

manavortex:
The array doesn't get overwritten anymore, no passing the ref anymore

seberoth:
> The array doesn't get overwritten anymore, no passing the ref anymore

Is this the answer to my comment on the other PR? Also why is it 2 now?
One will conflict with the other

manavortex:
The other one had the wrong title for some reason???

PR #1614, opened 2024-03-26
Dynamic class improvements

seberoth:
# Dynamic class improvements

**Implemented:**
- Better handling of dynamic classes in handles/weakhandled

**Fixed:**
- Small issues with creating custom types in the Type Selector


seberoth:
Could also look into adding custom types for other handles. Currently it is only visible for Jacks inkWidget stuff

PR #1616, opened 2024-03-27
unify archive lists in archive manager

rfuzzo:
unified the two caches in the archive manager into one
- added a method to only get the mod archives
- should support lookup and search as well


WolvenKit.Modkit/Managers/ArchiveManager.cs
```diff
@@ -566,89 +571,50 @@ public override void LoadAdditionalModArchives(string archiveBasePath, bool anal
         /// <inheritdoc />
         public override Optional<IGameFile> Lookup(ulong hash, ArchiveManagerScope searchScope)
```
seberoth, line 572: This looks wrong to me...
1. searchScope is not used
2. Shouldn't be mods first if requested? (Like load order)
3. Would place project at start. Though not related to this PR, it maybe should be selectable via searchScope)

rfuzzo: Done, thanks

seberoth (approved):

PR #1618, opened 2024-03-29
projectFileBrowser: hopefully more robust copy

manavortex:
# projectFileBrowser: hopefully more robust copy

Project browser's drag&drop move often failed silently - this is hopefully more robust. 
Additionally, copying without overwriting will now create `- Copy` items, the same as Windows Explorer.

### DIFF
--- a/WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
+++ b/WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
@@ -580,7 +580,7 @@ namespace WolvenKit.Views.Tools
 
             // Abort if a directory is dragged on itself or its parent
             if (!isCopy && sourceFiles.Count == 1 &&
-                (sourceFiles[0] == targetDirectory || !isCopy && Path.GetDirectoryName(sourceFiles[0]) == targetDirectory))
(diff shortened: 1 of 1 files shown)


WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
```diff
@@ -568,49 +570,132 @@ private void RowDragDropController_Dropped(object sender, TreeGridRowDroppedEven
...
+        /// </summary>
+        private async Task ProcessFileAction(IReadOnlyList<string> sourceFiles, string targetDirectory)
         {
-            var controlKeyDown = ViewModel.ModifierViewStateService.GetModifierState(ModifierKeys.Control);
-            
-            foreach (var sourceFile in sourceFiles)
+            var isCopy = ViewModel?.ModifierViewStateService.GetModifierState(ModifierKeys.Control) == true;
+
+            // Abort if a directory is dragged on itself or its parent
+            if (!isCopy && sourceFiles.Count == 1 &&
```
rfuzzo, line 582: Somethings odd here, why does it check for !isCopy twice? Or is that a typo?

manavortex: That was from "CopyOnly" being passed as parameter, good catch!

PR #1624, opened 2024-03-31
Material context menu entries

manavortex:
# Mesh context menu: "Rename material" and "Add new material definition"

image
image

Will work for `externalMaterials` and `localMaterialBuffer.materials`, creating a local or external material (or renaming one)


rfuzzo (approved):
sure

PR #1625, opened 2024-03-31
scene graph load fix when NotablePoints and LocStore are missing

JakubMarecek:
Fixed issue with scene graph loading when NotablePoints and LocStore are missing. The issue was in choice node which tried to get data from NotablePoints and LocStore, but if those are missing, it raises exception.

### DIFF
--- a/WolvenKit.App/ViewModels/GraphEditor/Nodes/Scene/scnChoiceNodeWrapper.cs
+++ b/WolvenKit.App/ViewModels/GraphEditor/Nodes/Scene/scnChoiceNodeWrapper.cs
@@ -19,9 +19,12 @@ public class scnChoiceNodeWrapper : BaseSceneViewModel<scnChoiceNode>
 
         var notable = _sceneResource
             .NotablePoints
-            .First(x => x.NodeId.Id == scnSceneGraphNode.NodeId.Id);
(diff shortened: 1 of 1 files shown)


WolvenKit.App/ViewModels/GraphEditor/Nodes/Scene/scnChoiceNodeWrapper.cs
```diff
@@ -57,14 +60,20 @@ private void GetChoices()
...
+            if (vdEntry != null)
+            {
+                var vpEntry = _sceneResource
+                    .LocStore
+                    .VpEntries
+                    .FirstOrDefault(x => x.VariantId.Ruid == vdEntry.VariantId.Ruid);
 
-            Options.Add($"[{vdEntry.LocaleId.ToEnumString()}] {vpEntry.Content}");
+                if (vpEntry != null)
+                    Options.Add($"[{vdEntry.LocaleId.ToEnumString()}] {vpEntry.Content}");
```
seberoth, line 73: Please use curly braces also on single lines. Rest looks good

seberoth (changes requested):

seberoth (approved):

PR #1633, opened 2024-04-06
descriptor read-only, "simple mode" editor

manavortex:
# ChunkViewModel: Descriptors and values

- Refactored both to use switch statements rather than endless if/then/else blocks
- CVM will now check if it should be read only 
- Stub implementation for ~~noob filter~~simple editor mode (I didn't manage to hide the rows in the treeview, or I'd have left it in)

From the first commit, only `WolvenKit.App/ViewModels/Tools/ChunkViewModel.UserInteractionStates.cs` is relevant for the review (read-only)


manavortex:
https://github.com/WolvenKit/WolvenKit/assets/965785/10df25fb-b5f2-4237-abcf-cf130458194a


Settings default value isn't applied yet somehow, I probably need to... hook some other event?

PR #1639, opened 2024-04-13
hiding nodes in .app and .ent

manavortex:
# Easy editor mode: Added some more fields across types


rfuzzo (approved):
I think that's okay, but we should refactor the whole descriptor thing a bit in the future: we already have the "ClassesExt" framework, where the red4 types are partially defined from the RTTI and we have custom extensions on them in Wkit. Blocked properties or descriptors can be handles for each class in their wkit extension, so we don't need a massive switch or hardcoded property names in one file - instead we can hardcode the names in the class extension

PR #1642, opened 2024-04-15
move hidden in ui property logic to partial classes

rfuzzo:
just a refactor to move the hidden fieldnames into the classes


rfuzzo:
On hold until I find a good place to put this in the WolvenKit.App project

manavortex (commented):
I love this, it's much better than my approach. 
Can we do this for descriptor and value too?

rfuzzo:
Thanks and yeah we can. But we'll need to find a way to do it with extensions maybe, not sure yet... in rust we could use a ui trait for that, but in c# I need to think a bit 😆

manavortex:
While we're playing Wunschkonzert, I care about the following properties for the GUI stuff: 

- Description
- Value
- properties included by noobfilter
- properties marked as read-only (these will also be included in the noobfilter)

Optional, something that I would _really_ love (but I don't know how to make it work), but I thought I'd throw it in here: 
image
image
- Display (as) child property
- GetDisplayTypeName

_I suggested automating this for anything with just one property, but @seberoth was against it because then it doesn't accurately reflect the CR2W structure anymore. Now that I found out how to add things to the settings, I can implement a toggle, though :))_

PR #1643, opened 2024-04-16
Add Ukrainian to the LocKey settings

FedorenkoCodes:
# Add Ukrainian to the LocKey settings

**Implemented:**
- As of now there is no way to access Ukrainian localization. This change adds a way for Ukrainian to be set in the settings for LocKey.

image


seberoth (approved):
Looks good to me. Thanks

PR #1645, opened 2024-04-19
Graph nodes properties

JakubMarecek:
Added node properties to graph editor for scenes and quests, currently not all, but some most common


rfuzzo (approved):
Details.AddReview("approved")

PR #1646, opened 2024-04-19
ArchiveXL dynamic paths, fixes mesh_ent preview

### LINKED ISSUE #1563: .ent preview: Support ArchiveXL dynamic variants
Currently, using dynamic variants completely breaks the .ent preview in Wolvenkit. File validation can already deal with the substitution, so let's implement this for Wkit as well.

In case of multiple variants, the preview should 
- take the first valid file path per substitution combo 
- offer a valid "default" appearance that will render the first appearance from each mesh involved 
- not offer any other appearances because IDK how to deal with them

Maybe we could implement material switching per submesh, offering the lists from the meshes on that end?

manavortex:
# ArchiveXL dynamic paths

**Fixed:**
- mesh entity preview for meshes with dynamic path subsitutions
- currently, only one appearance is added, but at least you can see _something_

**Additional notes:**
Partially resolves #1563 (or maybe does resolve it, but root entity file previews don't work yet. See the TODOs in `RDTMeshViewModel:RenderEntity`)

### DIFF
--- a/WolvenKit.App/Helpers/ArchiveXlHelper.cs
+++ b/WolvenKit.App/Helpers/ArchiveXlHelper.cs
@@ -1,0 +1,118 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Splat;
+using WolvenKit.App.Services;
+using WolvenKit.Core.Interfaces;
+
+namespace WolvenKit.App.Helpers;
+
+/// <summary>
+/// This class will resolve ArchiveXL dynamic variant depot paths.
+/// They start with an asterisk (*) and contain substitution wildcards (<see cref="s_keysAndValues"/>).  
+/// </summary>
+internal class ArchiveXlHelper
+{
+    private static readonly Dictionary<string, string[]> s_keysAndValues = new()
+    {
+        { "{camera}", ["fpp", "tpp"] },
+        { "{feet}", ["lifted", "flat", "high_heels", "flat_shoes"] },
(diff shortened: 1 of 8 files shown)


WolvenKit.App/Helpers/ArchiveXlHelper.cs
```diff
@@ -0,0 +1,118 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Splat;
```
dragonzkiller, line 5: What is Splat doing here? I would prefer to avoid adding more libraries than necessary--even though it's used in the main csproj

WolvenKit.App/WolvenKit.App.csproj
```diff
@@ -63,6 +63,7 @@
 	<PackageReference Include="HelixToolkit.SharpDX.Core.Wpf" />
     <PackageReference Include="AutomaticGraphLayout" />
     <PackageReference Include="Nodify" />
+    <PackageReference Include="Splat"/>
```
dragonzkiller, line 66: see other splat comment

dragonzkiller (commented):
most of this involves areas i'm not too familiar with, so i wouldn't want to approve. just the comments above

WolvenKit.App/Helpers/ArchiveXlHelper.cs
manavortex: it's required to get the service

rfuzzo (approved):

PR #1653, opened 2024-04-23
Alternative TrimmingTextBox

manavortex:
# better than original TrimmingTextBox


seberoth (approved):
Don't know what to say :P

PR #1662, opened 2024-04-27
Refactoring/archive manager replace with original

manavortex:
# Project browser: "Replace with original" 

Can now take a search scope as parameter (will default to currently-active scope). Context menu entry will search game and mods, but consider game first.

### DIFF
--- a/WolvenKit.App/Controllers/IGameController.cs
+++ b/WolvenKit.App/Controllers/IGameController.cs
@@ -1,15 +1,71 @@
 using System.Threading.Tasks;
 using WolvenKit.App.Models;
+using WolvenKit.Common;
 using WolvenKit.Core.Interfaces;
 
 namespace WolvenKit.App.Controllers;
 
 public interface IGameController
 {
+    /// <summary>
+    /// Finds a file in the currently active scope and adds it to the mod
+    /// </summary>
+    /// <param name="hash">Unique hash of the file</param>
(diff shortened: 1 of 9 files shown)


seberoth:
If you want to refactor game file loading, maybe you should work together with rfuzzo on #1616 
Also don't quite understand why you added an new constructor to `WolvenKitException` and marked it obsolete directly.

manavortex:
> Also don't quite understand why you added an new constructor to `WolvenKitException` and marked it obsolete directly.

Autocomplete/doc reasons. Expectation is that I do new Exception(string message) - I get better feedback from the UI this way

seberoth:
But you won't get any feedback for `throw new Exception(string message)`. Then you would need to mark `Exception` as obsolete, not `WolvenKitException`.

manavortex:
Right, but that case will also not trigger your LogWrapper, so I consider that a win :D

seberoth:
Tbh, I consider this as a lose. Now it allows negative error codes but they all get ignored/filtered out later.

manavortex:
Sure, then let's nuke it :)

manavortex:
Nuked it. Not sure about collaborating with Fuzzo, since my changes (hopefully) touch a whole different part of the file loading..ß

rfuzzo (approved):
👍

PR #1663, opened 2024-04-27
Feature/import archive

### LINKED ISSUE #1583: Convenience: Open .archive from file menu
Wolvenkit would become a lot more accessible to people who can't read if it had `File -> open .archive`

**Target group**: the Stupidest Assumable User

**Precondition**: 
An extra .archive folder must be configured. If that's not the case, WKit will show an annoying popup, telling the user to configure that now. (Bonus points if the annoying popup has a button that takes you directly to the settings page)

After clicking this entry, WKit should do the following:
1. Let the user select an .archive file via file picker
2. Copy the selected .archive to the extra folder (if the same file already exists, a warning should be displayed)
3. Switch the Asset Browser to the Mod Browser
3.1 If the Mod Browser is already opened, it should refresh (or toggle back to asset browser and back to mod browser again)
4. Pre-populate the search bar with `archive:<filename_without_extension>.archive` and apply the filter

manavortex:
# Import .archive

image

Doesn't check any prerequisites, but throws a lot of user-friendly error messages for Seb's new logviewer.

**Additional notes:**
closes #1583

### DIFF
--- a/WolvenKit.App/Services/SettingsManager.cs
+++ b/WolvenKit.App/Services/SettingsManager.cs
@@ -10,6 +10,7 @@ using DynamicData.Binding;
 using WolvenKit.App.Models;
 using WolvenKit.Common;
 using WolvenKit.Core;
+using WolvenKit.Core.Exceptions;
 using WolvenKit.Core.Extensions;
 
 namespace WolvenKit.App.Services;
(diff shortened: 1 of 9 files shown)


seberoth:
Why does it contain stuff from #1662?

manavortex:
because git hates me, that's why :)

seberoth (approved):

PR #1666, opened 2024-04-27
skipping dynamic path resolution (and check)  for non-mesh-components

manavortex:
# Fixing missing meshes in sector preview

Turns out that the check was a little too thorough


Simarilius-uk:
Fixes the issue here.

seberoth (approved):

PR #1671, opened 2024-04-28
fix(GUI): Fix crash while removing TreeView items

seberoth:
fix(GUI): Fix crash while removing TreeView items

Fixed:
- Crash while removing TreeView items


manavortex (commented):
LGTM. I mean, considered that Syncfusion. Thank you for the comment, let's never touch this again until you refactor it

manavortex (approved):

PR #1672, opened 2024-04-28
First actor test isnt triggering most of the time, this seems to.

Simarilius-uk:
**Fixed:**
The actor bit of the log message wasnt triggering in my testing, added a second test that does. Didn't take the first out as I wasn't sure if there is another case I'm just missing.


rfuzzo (approved):
lgtm, no idea what it does tho

Issue #1668, opened 2024-04-28
WScript function to create a new instance of a datatype in a given format


justarandomguyintheinternet:
**Describe the solution you'd like**
`wkit.CreateInstance(className, type)`
- `className`: The name of the class of which a new instance is to be created
- `type`: The format in which the instance should be returned, `"json"` or `"cr2w"`
- e.g. calling `wkit.CreateInstance("worldStaticMeshNode", "json")` would return the following JS object (Values are the default values when manually creating a new `worldStaticMeshNode` in WKit):
```json
{
  "$type": "worldStaticMeshNode",
  "castLocalShadows": "Default",
  "castRayTracedGlobalShadows": "Default",
  "castRayTracedLocalShadows": "Default",
  "castShadows": "Default",
  "debugName": {
    "$type": "CName",
    "$storage": "string",
    "$value": "None"
  },
  "forceAutoHideDistance": 0,
  "isHostOnly": 0,
  "isVisibleInGame": 1,
  "lodLevelScales": 4294967295,
  "mesh": {
    "DepotPath": {
      "$type": "ResourcePath",
      "$storage": "string",
      "$value": "engine\\meshes\\editor\\box.w2mesh"
    },
    "Flags": "Default"
  },
  "meshAppearance": {
    "$type": "CName",
    "$storage": "string",
    "$value": "default"
  },
  "occluderAutohideDistanceScale": 255,
  "occluderType": "Default",
  "proxyScale": null,
  "removeFromRainMap": 0,
  "renderSceneLayerMask": "Default",
  "sourcePrefabHash": "0",
  "tag": "None",
  "tagExt": "None",
  "version": 0,
  "windImpulseEnabled": 1
}
```

PR #1677, opened 2024-04-30
Added WScript CreateInstanceAsJSON function

justarandomguyintheinternet:
# WScript `CreateInstanceAsJSON` function

**Implemented:**
- `CreateInstanceAsJSON(className)`
- Takes the name of the class of which an instance is to be created, and returns it as a JSON formated string
- e.g. `wkit.CreateInstanceAsJSON("worldStaticMeshNode")`

**Additional notes:**
- Implements #1668, except for always returning JSON, as cr2w does not make any sense in this use-case


manavortex (approved):
LGTM

PR #1683, opened 2024-05-03
fix(ArchiveManager): Load order again?

seberoth:
Is it right now? I'm confussed -.-


manavortex (approved):
LGTM, but if you can nuke the reverse in L.507 as well, it'd look even better

dragonzkiller (approved):
LGTM

PR #1620, opened 2024-03-30
Improved CR2W handling

seberoth:
# Improved CR2W handling

Improved CR2W handling

PR #1696, opened 2024-05-05
fix(CR2W): Try to read the CR2W defined type before the RTTI one

seberoth:
Solves an issue introduced with #1620 for files from old game versions.


manavortex (approved):
LGTM

PR #1699, opened 2024-05-07
Feature/replace path after rename

manavortex:
refactor/rename like

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
@@ -26,6 +26,7 @@ using WolvenKit.App.Models.ProjectManagement.Project;
 using WolvenKit.App.Services;
 using WolvenKit.App.ViewModels.Shell;
 using WolvenKit.Common;
+using WolvenKit.Common.Extensions;
 using WolvenKit.Common.FNV1A;
 using WolvenKit.Common.Interfaces;
 using WolvenKit.Core.Extensions;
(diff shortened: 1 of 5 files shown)


WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -948,6 +960,98 @@ private void OpenInMlsb()
...
+        using (var cr = new CR2WReader(fs))
+        {
+            if (cr.ReadFile(out cr2w) != EFileReadErrorCodes.NoError)
+            {
+                return;
+            }
+
+            foreach (var result in cr2w!.FindType(typeof(IRedRef)))
+            {
+                if (result.Value is not IRedRef resourceReference || resourceReference.DepotPath != oldPath)
```
manavortex: I think this will not work with folders, you'd have to check StartsWith?

WolvenKit.RED4/Types/Primitives/Simples/ResourcePath.cs
```diff
@@ -10,6 +10,8 @@ namespace WolvenKit.RED4.Types;
 [DebuggerDisplay("{GetResolvedText()}", Type = "ResourcePath")]
 public readonly struct ResourcePath : IRedString, IRedPrimitive<string>, IEquatable<ResourcePath>, IComparable<ResourcePath>, IComparable
 {
+    public static readonly char DirectorySeparatorChar = '\\';
```
manavortex, line 13: Can't you get this from Path as default separator, or something like that?

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
seberoth: Not sure why you commented on a outdated file

WolvenKit.RED4/Types/Primitives/Simples/ResourcePath.cs
seberoth: For Windows it would work, yes. But on Linux/Mac not.
CR2W ALWAYS uses `\` doesn't matter which OS

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
manavortex: because git sorted them in reverse temporal order and I was so confused -.-

seberoth (approved):

PR #1709, opened 2024-05-14
Fix/dynamic material export

### LINKED ISSUE #1706: Material export writes wrong texture paths
**Describe the bug**
When exporting the mesh file in the attached file, material export writes incorrect values into two texture slots. Inside the referenced .mi file the paths exist.

**To Reproduce**
1. Download example 
[test.zip](https://github.com/WolvenKit/WolvenKit/files/15287414/test.zip)
2. Export the included he_000_pma_c__basehead.mesh with materials.
3. Observe Normal and Roughness paths mismatch from paths inside mi file.
4. Also note that the bugged paths are placing mi files into texture slots, ensuring a game crash upon load.

**Expected behavior**
Material export keeps all data that's referenced inside the mi (and chain if applicable). 

**Version (please complete the following information):**
 - WolvenKit 8.14.1-nightly-05-11

**Screenshots**
Screenshot_693

manavortex:
**Additional notes:**
closes #1706

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/Common/MaterialExtractor.cs
+++ b/WolvenKit.Modkit/RED4/Tools/Common/MaterialExtractor.cs
@@ -8,6 +8,7 @@ using WolvenKit.Common.FNV1A;
 using WolvenKit.Common.Model.Arguments;
 using WolvenKit.Core.Extensions;
 using WolvenKit.Core.Interfaces;
+using WolvenKit.Modkit.Resources;
 using WolvenKit.RED4.Archive;
 using WolvenKit.RED4.Archive.CR2W;
 using WolvenKit.RED4.Archive.IO;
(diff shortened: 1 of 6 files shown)


rfuzzo:
I have no idea about this, who can review it?
does it break the blender plugin?

WolvenKit.Modkit/RED4/Tools/Common/MaterialExtractor.cs
```diff
@@ -370,26 +503,22 @@ void ExtractToJson(CR2WFile file, List<ICR2WImport> imports, string destFileName
...
 
     private FindFileResult TryFindMaterial(CR2WFile parent, IRedRef resourceReference, out FindFileRecord result, bool excludeCustomArchives = false)
     {
         var status = TryFindFile2(parent, resourceReference, out result);
-        if (status == FindFileResult.NoCR2W)
-        {
-            throw new Exception($"Error while parsing the file: {resourceReference.DepotPath.GetResolvedText()}");
-        }
-        if (status == FindFileResult.FileNotFound)
+        if (status is FindFileResult.NoCR2W or FindFileResult.FileNotFound)
```
seberoth, line 514: Same as above

WolvenKit.Modkit/RED4/Tools/Common/MaterialExtractor.cs
```diff
@@ -259,14 +389,22 @@ private static readonly IRedRef
...
+
+            if (escapedMaterialName is not "" && resourceReference.DepotPath.GetResolvedText() is string depotPath &&
+                depotPath.StartsWith('*'))
+            {
+                resourceReference =
+                    new CResourceAsyncReference<IMaterial>(depotPath.Replace(s_materialWildcard, escapedMaterialName)[1..]);
+            }
+            
+            var status = TryFindFile2(parentFile, resourceReference, out var result);
+            if (status is FindFileResult.NoCR2W or FindFileResult.FileNotFound)
```
seberoth, line 407: Those are 2 distict error message. Why did you merge them?

seberoth (commented):
Didn't test yet. Just some general things.
Would it be possible (without major rework) to make this more abstract? (Moving the AXL stuff to its own method)
Not the biggest fan of mixing "native" archive/cr2w stuff with custom AXL logic.
Only if it doesn't is to much work though.

WolvenKit.Modkit/RED4/Tools/Common/MaterialExtractor.cs
manavortex: Because I can't read! I'll un-do this

manavortex:
Not sure if the ArchiveXL logic can be easily factored out. I can put the substitution into the static class, though. Just hand over the path and the potential parameter, and get an always-valid path back.

seberoth (approved):

PR #1715, opened 2024-05-19
usability: cvm doubleclick behaviour

manavortex:
double click on collapsed element will collapse all of its siblings

### DIFF
--- a/WolvenKit/Views/Tools/RedTreeView.xaml.cs
+++ b/WolvenKit/Views/Tools/RedTreeView.xaml.cs
@@ -142,7 +142,6 @@ namespace WolvenKit.Views.Tools
                 return;
             }
 
-            var test = Locator.Current.GetService<AppViewModel>();
             var tt = Locator.Current.GetService<AppViewModel>();
 
             if (tt.ActiveDocument is not RedDocumentViewModel { SelectedTabItemViewModel: RDTDataViewModel rdtd })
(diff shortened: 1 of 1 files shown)


WolvenKit/Views/Tools/RedTreeView.xaml.cs
```diff
@@ -385,7 +381,14 @@ private void OnDoubleClick(object sender, MouseButtonEventArgs e)
                 return;
             }
 
+            // If the current chunk is not expanded, collapse all of its siblings
+            if (!chunk.IsExpanded && chunk.Parent is ChunkViewModel parent)
```
rfuzzo, line 385: I kinda don't understand this... so if the selected item is not expanded, then we select the parent?

manavortex:
if the node is expanded and I double-click it, it'll expand/collapse its children
if the node is collapsed and I double-click it, it'll expand/collapse its siblings

rfuzzo:
> if the node is expanded and I double-click it, it'll expand/collapse its children if the node is collapsed and I double-click it, it'll expand/collapse its siblings

🤔 hmm I do think I understand the concept, but how does the code work, it basically selects the parent correct?

manavortex:
Yeah, the rest is already in. I just noticed today that I needed this after doing a lot of search&replace in an .app with 168 appearances and had to scroll up to the parent node each time to re-collapse the nodes

rfuzzo (approved):
ok

PR #1716, opened 2024-05-19
mesh materialViewer should always have a material

manavortex:
generate a default material if we don't have one


rfuzzo:
This looks good and sounds reasonable. but we change the mesh right? can that have unintended consequences? What if I delete the default material manually again? will it be created again?

manavortex:
true... but we don't mark anything as dirty, so there's a good chance the user will just overwrite? :| 
I guess we could delete the stuff again when leaving the appearance tab

rfuzzo:
not sure.. we could also just mark dirty, that works for me as well. what do you think?

manavortex:
should print a console message in that case. I dunno, I like the other one more, I currently have lots of material-less meshes because of ArchiveXL appearance patching :D

rfuzzo (approved):
thx, lgtm

PR #1719, opened 2024-05-20
feature(CR2WEditor): Added possibility to extract/save embedded files in the CR2WEditor view.

seberoth:
Added possibility to extract/save embedded files in the CR2WEditor view.


manavortex (approved):
wow, lgtm :)

PR #1720, opened 2024-05-20
pop-up: save dirty files before packing/launching

### LINKED ISSUE #1712: Warn about un-saved files when packing/installing mod
When the user has un-saved changes to their files when trying to install or pack the mod, Wolvenkit should show them a popup (Save now / pack anyway)

manavortex:
# save dirty files before packing/launching

**Additional notes:**
closes #1712

### DIFF
--- a/WolvenKit.App/ViewModels/Documents/DocumentViewModel.cs
+++ b/WolvenKit.App/ViewModels/Documents/DocumentViewModel.cs
@@ -34,7 +34,7 @@ public abstract partial class DocumentViewModel : PaneViewModel, IDocumentViewMo
     public bool IsDirty
     {
         get => _isDirty;
-        set => SetProperty(ref _isDirty, value);
+        protected set => SetProperty(ref _isDirty, value);
     }
 
     private bool _isSimpleViewEnabled;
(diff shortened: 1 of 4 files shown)


WolvenKit.App/ViewModels/Documents/DocumentViewModel.cs
```diff
@@ -34,7 +34,7 @@ public DocumentViewModel(string path) : base(Path.GetFileName(path), path)
     public bool IsDirty
     {
         get => _isDirty;
-        protected set => SetProperty(ref _isDirty, value);
+        set => SetProperty(ref _isDirty, value);
```
seberoth, line 37: Why did you make it public? Should kept private/protected.

seberoth (changes requested):

WolvenKit.App/ViewModels/Documents/DocumentViewModel.cs
manavortex: The compiler didn't let me and insisted it needed to be public. It's protected in the interface, though.

seberoth: Remove the setter from the interface

manavortex: Ah, yeah, makes sense. Fixed

seberoth (approved):

Issue #1732, opened 2024-05-24
Scenes graph view - option to hide extra details


anygoodname:
Could you please consider adding an option to hide extra details in the Scenes graph view?

Currently the abundance of details make it hard to read it all, inflate node rectangles consuming large screen space to show just few nodes out of hundreds in effect, so it's hard to fit scene parts on screen and follow the scene flow, and make the vital parts like node ids or node type hard to recognize.

Just a general option like "show node details", on/off for the whole view would suffice.

Kind regard.
Anygoodname.

PR #1733, opened 2024-05-24
Graph nodes properties

JakubMarecek:
Tweak for issue #1732 

Added true / false setting "Show Graph Editor Node Properties" to main app settings under "Display". It's enabled by default.


rfuzzo (approved):
I hereby mindlessly approve this

PR #1735, opened 2024-05-25
fixed simple mode hiding for non-empty arrays

manavortex:
simple mode was hiding non-empty arrays, it'll no longer do that


rfuzzo:
looks good to me, but I also have no idea about mesh components so I don't feel particularly confident reviewing this

manavortex:
You can try it out, open any mesh with preloadMaterialInstances (e.g. Hanako's coat). If the array isn't empty, simple mode shoudln't hide it

seberoth (approved):

PR #1740, opened 2024-05-26
rename material now works properly

manavortex:
will now mark parent tab as dirty and replace material name in chunkMaterials


rfuzzo (approved):
lgtm

PR #1747, opened 2024-05-27
Fixed cfoliage buffer layout

seberoth:
Fixed cfoliage buffer layout

**Fixed:**
- cfoliage buffer layout. Credits to @kmirea


Simarilius-uk (approved):
Tested with the plugin, format seems right, works great.

PR #1751, opened 2024-05-28
Root Motion

### LINKED ISSUE #1692: Root motion corner case validation
```
                // TODO Root motion corner case validation
```

https://github.com/WolvenKit/WolvenKit/blob/182798da55ef6d1586025443beddce87718444bd/WolvenKit.Modkit/RED4/Tools/Animation/AnimSIMD.cs#L345

effs:
## Fixes

- [x] Root Motion export logic (incl. yaw-only rotations)

## Adds

- [x] Root Motion import 
- [x] Lightweight type level support for differentiating ZupLhs (wkit) and YupRhs (glTF) coordinates so that you can enforce using the correct one in the correct place plus conversions

## Other

- [x] Ton of refactoring in AnimTools and , primarily there's now a struct for passing the data around from and to encoders so that logic is not duplicated where possible.

## Detes

- Root Motion export with toggle, import without toggle based on glTF data
- Root Motion is `Translation` and possibly `Rotation`, type dependent. No `Scale`.
- Only Root Motion *or* regular root joint transforms (if any) are exported, not both, not mixed
- glTF storage for RM is in root joint, as per above
- glTF extras for each anime schema v4 includes `RootMotionType` that tries to be clearer on intent
  - `None` if RM requested but there's none
  - `Unknown` if not requested (whether or not one exists)
  - Others if requested and found
- Previous schemas migrated with `RootMotionType.Unknown`
  - `None` uses root transforms as regular and deletes old RM
  - `Unknown` copies old RM from CR2W, if any, doesn't touch root joint, essentially equivalent to toggle
  - Others store indicated type and clear root joint transforms
 
 ## Also Done
 
 closes #1692  

# TODO separately

- [ ] #1796 
- [ ] #1788

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/Animation/AnimRootMotion.cs
+++ b/WolvenKit.Modkit/RED4/Tools/Animation/AnimRootMotion.cs
@@ -229,6 +229,141 @@ namespace WolvenKit.Modkit.RED4.Animation
 
 #endregion export
 
+#region import
+        public static (RootMotionType rootMotionType, bool rootTransformsStripped) ExtractRootMotion(out animIMotionExtraction? rootMotion, ref readonly AnimationBufferData incomingAnimData, ref readonly animAnimation? oldAnimDesc, RootMotionType incomingType, string animName, ILoggerService _loggerService)
+        {
+            var rootTranslations = incomingAnimData.Translations.GetValueOrDefault(RootJointIndex, []);
+            var rootRotations = incomingAnimData.Rotations.GetValueOrDefault(RootJointIndex, []);
+
+            if (incomingType is RootMotionType.Unknown or RootMotionType.None)
+            {
+                if (rootTranslations.Count > 0 || rootRotations.Count > 0)
+                {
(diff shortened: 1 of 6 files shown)


effs:
- Root Motion export with toggle, import without toggle based on glTF data
- Root Motion is `Translation` and possibly `Rotation`, type dependent. No `Scale`.
- Only Root Motion *or* regular root joint transforms (if any) are exported, not both, not mixed
- glTF storage for RM is in root joint, as per above
- glTF extras for each anime schema v4 includes `RootMotionType` that tries to be clearer on intent
  - `None` if RM requested but there's none
  - `Unknown` if not requested (whether or not one exists)
  - Others if requested and found
- Previous schemas migrated with `RootMotionType.Unknown`
  - `None` uses root transforms as regular and deletes old RM
  - `Unknown` copies old RM from CR2W, if any
  - Others store indicated type and clear root joint transforms

effs:
@seberoth Not sure how worthwhile it is to dig into the actual logic here but if you have the chance, I'd like your eyes on 

1. The type-level stuff and the more fp-style flow over iteration, if there's anything too jarring

2. The epsilon and other normalization both in that it makes sense and whether we need it elsewhere

PR #1760, opened 2024-06-02
log duplicated nodes ID in graph editor

JakubMarecek:
It will show in log if there any duplicated node IDs in questphase.


manavortex (changes requested):
Might be better to put this into file validation? @ me on discord if you agree, I can either hack it in real quick or show you where :)

JakubMarecek:
OK, I will make some changes and make new PR.

PR #1776, opened 2024-06-09
CVM: default expansion state on first open

manavortex:
.mi files will now automatically expand their `values` array

first commit only contains automatic refactors


rfuzzo (approved):
I think that's fine, it's not a bad idea to have custom view states for specific file types where it makes sense.

PR #1777, opened 2024-06-09
Another memory optimization patch

0x25CBFC4F:
## Another memory optimization patch

**Implemented:**
- Oodle decompressing now allocates/deallocates unmanaged memory only to prevent massive byte[] going into LOH
- Right after shell initialization application now requests LOH (and every other generation) cleanup to handle other unfixed LOH cases for now.


seberoth (approved):
LGTM

PR #1782, opened 2024-06-10
Fixes erroneous handling of non-noderef strings

0x25CBFC4F:
**Fixed:**
- Fixes erroneous handling of non-noderef type strings


manavortex (approved):
LGTM

PR #1784, opened 2024-06-10
HashTool window now uses new algo to calculate NodeRef's hash

0x25CBFC4F:
**Fixed:**
- Hash Tool window used old algorithm to calculate NodeRef's hash. That is fixed now.


manavortex (approved):
LGTM

Issue #1723, opened 2024-05-21
Launch profiles: Option to delete packed folder after zipping/installing


manavortex:
With many Wolvenkit projects (especially NPVs), `packed` folders can accumulate to a significant storage space drain. We should add an option to the launch profiles to delete it if the user installs or zips the mod.

PR #1789, opened 2024-06-11
Delete packed folder, HotReload packs everything

manavortex:
# Packing improvements

#1723 Can now check a box to remove the `packed` folder after packing:
image

"Hot Reload" will now install all files, not just the .archive

### DIFF
--- a/WolvenKit.App/Controllers/RED4Controller.cs
+++ b/WolvenKit.App/Controllers/RED4Controller.cs
@@ -207,6 +207,7 @@ public class RED4Controller : ObservableObject, IGameController
                 return false;
             }
         }
+        
         _loggerService.Success($"{_projectManager.ActiveProject.ModName} packed into {hotdirectory}");
         _notificationService.Success($"{_projectManager.ActiveProject.ModName} packed into {hotdirectory}");
 
@@ -217,7 +218,7 @@ public class RED4Controller : ObservableObject, IGameController
     /// cleans the packed directory
     /// </summary>
     /// <returns></returns>
-    public bool CleanAll()
+    public bool CleanAll(bool isPostBuild = false)
     {
         _progressService.IsIndeterminate = true;
         // checks
@@ -231,10 +232,7 @@ public class RED4Controller : ObservableObject, IGameController
         }
 
         // perform cleanup
-        if (!Cleanup(cp77Proj, new LaunchProfile()
-        {
-            CleanAll = true
-        }))
+        if (!Cleanup(cp77Proj, new LaunchProfile() { CleanAll = true }, isPostBuild))
         {
             var errorMessage = "An error occured during clean all. Some files may not be removed.";
             _progressService.IsIndeterminate = false;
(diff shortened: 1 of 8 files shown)


WolvenKit.App/Models/LaunchProfile.cs
```diff
@@ -16,9 +16,13 @@ public class LaunchProfile
     // install script files
 
     [Category("Install")]
-    [Display(Name = "Clean packed directory completely first")]
+    [Display(Name = "Clean packed directory before build?")]
     public bool CleanAll { get; set; } = true;
 
+    [Category("Install")]
+    [Display(Name = "Clean packed directory after build?")]
```
rfuzzo, line 23: what's the point of this exactly? :D why clean after a build?

WolvenKit.App/Models/LaunchProfile.cs
```diff
@@ -16,9 +16,13 @@ public class LaunchProfile
     // install script files
 
     [Category("Install")]
-    [Display(Name = "Clean packed directory completely first")]
+    [Display(Name = "Clean packed directory before build?")]
```
rfuzzo, line 19: should that really ever be allowed?

WolvenKit.App/Controllers/RED4Controller.cs
```diff
@@ -269,15 +267,20 @@ private static bool IsCDPRScript(string fileName)
         };
     }
 
-    private static IEnumerable<string> GetScriptFiles(Cp77Project cp77Proj) 
-        => Directory.EnumerateFiles(cp77Proj.ResourceScriptsDirectory, "*.*", SearchOption.AllDirectories).Where(name => IsCDPRScript(name));
+    private static List<string> GetScriptFiles(Cp77Project cp77Proj)
```
rfuzzo, line 276: why the change to List?

WolvenKit.App/Models/LaunchProfile.cs
manavortex: It's the current default behaviour :D

WolvenKit.App/Models/LaunchProfile.cs
manavortex: because Nola complained that NPV projects clutter up her disk, and tbh she has a point

WolvenKit.App/Controllers/RED4Controller.cs
manavortex: ConcurrentModification warnings when accessing

rfuzzo: ah right, you made it async 👍

WolvenKit.App/Models/LaunchProfile.cs
rfuzzo: eh okay, I guess we could remove that setting

rfuzzo (approved):

rfuzzo (changes requested):
ah, the regex thing is still unresolved

rfuzzo (approved):
lgts (looks good to seb)

PR #1790, opened 2024-06-11
rename dialog will show refactoring checkbox

### LINKED ISSUE #1688: Feature: Rename/move file and update references
# Refactor: Rename
When re-naming files or folders in Wolvenkit, it would be great if references in files of the current project could be updated as well. 

Since that'll probably wreak havoc on performance, this should probably not be the default implementation? Maybe we can do smart filtering (e.g. if I just moved a .mi, we only need to look inside of .mi, .mesh, and .streamingsector...)

manavortex:
image

Closes #1688


rfuzzo (approved):
thanks

PR #1834, opened 2024-07-07
CLI: Use guessed extension when uncooking

dragonzkiller:
Worst case: bin is still extracted and the file conversion will fail. Best case: you actually get the uncooked file.


manavortex (approved):
LGTM

PR #1862, opened 2024-07-27
feature: import entspawner json

manavortex:
# feature: import entspawner json
Will open a json file from the entspawners directory, copy it to raw and noodle it through the script. Not sure if all kinks are ironed out, will require testing.


justarandomguyintheinternet:
Works great for me (Aside from some initial issues probably due to the .wscript not having been pulled from git), but I'm not a fan of it deleting the imported file from the `export` folder

manavortex:
It shouldn't do that - it copies it over to the wkit project, then deletes _that_.

PR #1913, opened 2024-09-02
Refactor uncook CLI command to use complex type instead of a massive generic

Zhincore:
# Refactor uncook CLI command to use complex type instead of a massive generic

**Implemented:**
- The uncook command no longer uses a generic class with 18 type arguments for CLI arguments.
  - Fuzzo found out about [more complex types docs](https://github.com/dotnet/command-line-api/blob/main/docs/model-binding.md#more-complex-types)

**Fixed:**
- probably fixed bug where rawOutDir was never created?


rfuzzo (commented):
looks good, but does it break existing usage? e.g. mlsb?

Zhincore:
I haven't tested mlsb, but from my simple manual tests nothing seems to have changed, I didn't change the CLI options themselves at all

rfuzzo:
> I haven't tested mlsb, but from my simple manual tests nothing seems to have changed, I didn't change the CLI options themselves at all

kewl, ship it then

PR #1914, opened 2024-09-02
Fix uncooking OpusInfo not dumping JSON

Zhincore:
# Fix uncooking OpusInfo not dumping JSON

**Fixed:**
- ~~Seb~~ Someone forgot to dump JSON from OpusInfo when requested in Export Tool, I fixed that

**Additional notes:**
It simply serializes the OpusInfo class, just like original sacred scripts from HitmanHimself did.


manavortex (approved):
LGTM, I guess!

PR #1937, opened 2024-09-28
Scripting: Added `@description` and `@usage` script file headers.

### LINKED ISSUE #1936: Feature Request: Add Optional Description to Scripts

Not all scripts are a one step process where you just run the script and it's done, or there are caveats to how and when it should be used. But there is currently no good solution to providing additional information about the script to the user in a way that is intuitive and clear.

If you create a wiki entry for your script only the people who come across your script from the wiki will know about it
If you add the description into the code itself even less people will see it
If you send a discription or disclaimer via the console output it will clog up the console or might go unnoticed due to other output.

Just like there is an author and version field there could be a description field that is displayed in the same location.

seberoth:
# Scripting: Added `@description` and `@usage` script file headers.

**Implemented:**
- Added `@description` and `@usage` script file headers. Closes #1936 

Description and Usage blocks are terminated by either a non comment line or the next keyword.

**Example:**
```
// @description
// Line 1 of description
// Line 2 of description
// ...
// Last line of description
// 
// @usage
// Line 1 of usage
// Line 2 of usage
// ...
// Last line of usage
var test = 0;
```

### DIFF
--- a/WolvenKit.App/ViewModels/Scripting/ScriptFileViewModel.cs
+++ b/WolvenKit.App/ViewModels/Scripting/ScriptFileViewModel.cs
@@ -10,6 +10,8 @@ public class ScriptFileViewModel : ScriptViewModel
 
     public string? Version => _scriptFile.Version;
     public string? Author => _scriptFile.Author;
+    public string? Description => _scriptFile.Description;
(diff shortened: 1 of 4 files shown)


manavortex (approved):
LGTM :)

PR #1948, opened 2024-09-30
if we close these, we can all go home

manavortex:
# fix code analyser findings

### DIFF
--- a/WolvenKit.App/ViewModels/Dialogs/SoundModdingViewModel.cs
+++ b/WolvenKit.App/ViewModels/Dialogs/SoundModdingViewModel.cs
@@ -79,25 +79,27 @@ public partial class SoundModdingViewModel : DialogViewModel
         }
 
         var path = Path.Combine(Environment.CurrentDirectory, "Resources", "soundEvents.json");
-        path = Path.Combine(_projectManager.ActiveProject.ResourcesDirectory, "info.json");
-        if (File.Exists(path))
+        if (!File.Exists(path))
         {
-            try
+            return;
+        }
+
+        try
+        {
+            _info = JsonSerializer.Deserialize<ModInfo>(File.ReadAllText(path), _options).NotNull();
+            foreach (var e in _info.CustomSounds)
             {
-                _info = JsonSerializer.Deserialize<ModInfo>(File.ReadAllText(path), _options).NotNull();
-                foreach (var e in _info.CustomSounds)
-                {
-                    CustomEvents.Add(e);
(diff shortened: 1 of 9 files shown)


WolvenKit.App/ViewModels/Dialogs/SoundModdingViewModel.cs
```diff
@@ -79,26 +79,28 @@ private void LoadInfo()
         }
 
         var path = Path.Combine(Environment.CurrentDirectory, "Resources", "soundEvents.json");
-        path = Path.Combine(_projectManager.ActiveProject.ResourcesDirectory, "info.json");
```
seberoth, line 82: Now it reads the wrong file. Should be info.json, not soundEvents.json

WolvenKit.Modkit/RED4/Tasks/UncookTask.cs
```diff
@@ -38,19 +38,20 @@ public int UncookTask(FileSystemInfo[] paths, UncookTaskOptions options)
...
-        }
 
         if (options.meshExportType != null && string.IsNullOrEmpty(options.meshExportMaterialRepo) && options.outpath is null)
         {
             _loggerService.Error("When using --mesh-export-type, the --outpath or the --mesh-export-material-repo must be specified.");
             return ERROR_INVALID_COMMAND_LINE;
         }
 
-        if (options.gamepath != null && options.gamepath.Exists)
+        if (options.outpath == null)
```
seberoth, line 48: Same "problem" as before. The analyzer would still complain that there is a double null check (L42)

WolvenKit.RED4/Types/Classes/ButtonCursorStateView.cs
```diff
@@ -31,7 +31,7 @@ public CName DefaultStateName
 		public ButtonCursorStateView()
 		{
 			HoverStateName = "Hover";
-			PressStateName = "Hover";
+			PressStateName = "Press";
```
seberoth, line 34: Don't edit `WolvenKit.RED4.Types.Classes`. These are autogenerated.
If its wrong, its wrong at CDPRs end.

WolvenKit.App/ViewModels/Dialogs/SoundModdingViewModel.cs
manavortex: Oh, alright! I was guessing from context, but you actually know. Fixing :)

PR #1955, opened 2024-10-09
some (switch, quest output) node properties

JakubMarecek:


dragonzkiller:
There is a lot going on here and no description of what your changes are doing and how they're beneficial to a user of wkit

rfuzzo:
thanks, looks good 👍

PR #1956, opened 2024-10-12
feature/searchbar for CR2W editor

manavortex:
# Initial implementation of CR2W editor search bar 

image
image

### DIFF
--- a/WolvenKit.App/Helpers/StringHelper.cs
+++ b/WolvenKit.App/Helpers/StringHelper.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Text.RegularExpressions;
 using WolvenKit.App.Helpers.StringHelpers;
 using WolvenKit.RED4.Types;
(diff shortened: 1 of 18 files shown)


poirierlouis:
Could you reuse XAML from search bar present in AssetBrowser and ProjectExplorer to have the same UI?

manavortex:
Could probably steal the styles, yeah. Maybe you'd like to define a neat input box style that we can all inherit from somewhere? :D

poirierlouis:
I can't take a look right now but I think there is already one used in the views I mentioned. It also supports the erase button/icon iirc.

rfuzzo:
image
One thing I noticed: the search bar clips with the tab bar a bit

rfuzzo:
request: it should search case-insensitively (right now it is case-sensitive)

rfuzzo:
Requests:
- [ ] it should search case-insensitively (right now it is case-sensitive)
- [ ] and (maybe for a future feature): option to search for variable names (right now it searches for values only as far as I can see)

and great work! 👍 this is amazing

manavortex:
> Requests:
> 
>     * [ x]  it should search case-insensitively (right now it is case-sensitive)
`StringComparison.OrdinalIgnoreCase` exists! Who'd have thought :))

>     * [ ]  and (maybe for a future feature): option to search for variable names (right now it searches for values only as far as I can see)
Yes, it's value only (and it's not 100% reliably because it only searches through `Description`, `Value`, and whatever StringHelper.ToString spits out). 
I would love to give the RTTI-types a proper ToString implementation, but I have no idea how to achieve that. I keep forgetting which of these files are generated and which aren't :D 

> and great work! 👍 this is amazing
Thank you! I've wanted this forever and ever and ever

Issue #1965, opened 2024-10-19
Make it easier to work with inkwidget and WidgetPreview


MadradMagerman:
Add ability to turn on/off **Pixel Grid** in _WidgetPreview_ to get more accurate and precise placement results as in Photoshop

Possibility to choose one of the options:: 
Grid is created within the widget main canvas (Root)
Grid is created on the entire WidgetPreview window (outside the canvas too)

*The on/off button is in the same place as _<Clear Cache>_ and _<Regenerate inkTextWidgets>_ buttons

manavortex:
- Can you put a screenshot where you'd like the toggle button to be?
- Would it be viable to have the grid disabled while Shift or Ctrl is being pressed?

MadradMagerman:
1. ![image](https://github.com/user-attachments/assets/56b1e8dc-3849-4539-9eed-a2e17119a967)
2. Not sure, but probably via bool activation

MadradMagerman:
Better example + additional enhancement

Another enhancement came to mind... (probably more complex)

Show **Layer Edges** of each element in the Widget Preview

Implementation method:
a) By clicking the cursor on a specific element on the canvas (single)
b) Show all borders of all elements in the Preview mode (mass)

``Layer Edges`` has 3 different colors for better view:
1 Canvas
2 Text
3 Other Elements
![image](https://github.com/user-attachments/assets/d9595171-8e7c-4f45-a4df-4adaf1dc6e2d)

MadradMagerman:
Grid example
![image](https://github.com/user-attachments/assets/45ce03d9-b8cb-42da-8c34-dd01064635d2)

manavortex:
Fixed in next Nightly :)

PR #1968, opened 2024-10-19
#1965 pixel grid snapping

manavortex:
# toggle for inkwidget edge mode

This does not actually overlay any grids yet. Turns out that we talked straight past each other.


manavortex:
image

PR #1977, opened 2024-10-24
feat(graphs): Refactor RedGraph

dustinlacewell:
// This changeset is in preparation of support for `.behavior` and other resource types.

WolvenKit currently supports graph views for quest and scene related resources.  The quest and scene specific portions of this support is located primarily in partial classes of `RedGraph`. An enumeration is checked throughout `RedGraph` to pick the right subset of behavior to use. 

Additionally, a number of details about individual node types associated with a given resource type are also handled via type-checks located throughout.

Concerns:
- With each supported resource type, `RedGraph` grows more crowded.
- Challenge to get a picture of logic specific to a given resource type.
- Challenge to discern what's needed to implement support for additional resource types
- Node specific view behaviors contained within parent views behind type-checks

The aims of this PR:
- Have `RedGraph` delegate to `IGraphService` for resource-specific behavior.
- Clarify the requirements for supporting additional resource types with `IGraphService`.
- Delegate node specific behaviors to node view classes implementing various interfaces. 
- Decompose graph concerns into sub-services where appropriate (node layout, view caching, etc.)

## Strategy Pattern

`RedGraph` becomes a facade which delegates *most* of its behavior to a valid `IGraphService` implementation like `QuestGraphService`, `SceneGraphService`, etc.

This scales and provides some design-time benefits:
- `IGraphService` makes it clear what's entailed to support an additional resource-type
- `BaseGraphService` is provided to provide an abstract base of shared behavior.
- `RedGraph` controls what is invariant and what may be customized based on what is delegated to `IGraphService`.
- Only a single check against the resource type is needed during construction of `RedGraph` to pick the relevant implementation of `IGraphService`. This obviates the need for the `RedGraphType` enumeration and existing checks against it.

### State Mediator

Rather than `RedGraph` passing `this` to `IGraphService`, a mediator `IGraphData` is provided which they and any related sub-services may refer without introducing cycles.

#### Why Lazy<>?
In the DI bindings, why utilize `Lazy<>` with `AddLazyTransient` for the factories? This is an easy path so that the factory classes and their node views are free to declare whatever they want as dependencies, including things that would otherwise form a cycle with `RedGraphFactory` (like `AppViewModel`, etc.). It's functionally the same as using `Locator.GetService` without depending on Splat.

## Node Type-Checking

Within the graph related view code there are places where node related behaviors are determined based on the type of the node view. For example, when a node is right-clicked, a context menu is built. Currently code like this lives within parent views rather than within the particular node view classes themselves. Parent views check the type of the node view, and do this or that based on the results. This can make it difficult to get a complete view of how a given node behaves.

In such situations, parent views should instead check whether a node implements an interface of interest, and if so, calling some standard method. When it comes to building a node's context menu, we might check for `IHasContextMenu` and call `.BuildContextMenu()` if it's found.

# Checklist

- [x] `RedGraph` is a facade delegating to `IGraphService`
	- [x] `BaseGraphService` captures shared behavior
	- [x] `QuestGraphService` implementing `IGraphService`
	- [x] `SceneGraphService` implementing `IGraphService`
	- [x] `QuestGraphFactory` 
		- [x] instantiates `RedGraph` with `QuestGraphService`
		- [x] handles quest node view creation
	- [x] `SceneGraphFactory` 
		- [x] instantiates `RedGraph` with `SceneGraphService`
		- [x] handles scene node view creation
- [ ] `RDTGraphView2` delegates to nodes:
	- [x]  `IDynamicInputNode`
		- [x] Calls `.AddInput` 
	- [x] `IDynamicOutputNode`
		- [x] Calls `.AddOutput`
	- [ ] `IGraphProvider` for sub-graph support
		- [x] Calls `.GetGraph` to generate sub-graph
		- [ ] Calls `.GetCrumbs` to generate breadcrumbs
- [ ] `GraphEditorView` delegates to `RedGraph`
	- [x] when viewport position changes
	- [x] to arrange nodes
	- [x] cleaning type-names
	- [ ] for canvas context-menu items
- [x] `GraphEditorView` delegates to node views
	- [x] for node context-menu items
	- [x] for socket recalculation

### WIP Regressions
- [ ] subgraphs don't automatically arrange nodes


dustinlacewell:
With this changeset these are the steps for introducing support for new resource types:

- Create a new directory at `WolvenKit.App:ViewModels\GraphEditor\Graphs\`
- Implement `IGraphService` for the resource type
- Implement at least one `NodeViewModel` for your graph type
- Implement a `IGraphFactory` for your `IGraphService` and `NodeViewModel` classes.
- Register your `IGraphFactory` with the DI container
- Add your `IGraphFactory` to `RedGraphFactory`
- Add the resource type you're supporting to the check in `RedDocumentViewModel.cs`

dustinlacewell:
I'm interested in getting some help with understanding how we might make the breadcrumb view automatically reactive on history data in the view-model, rather than having the view manually rebuild the breadcrumbs.

PR #1954, opened 2024-10-06
feat: add UI scaling + improve UX

poirierlouis:
# feat: add UI scaling + improve UX

**Implemented:**
- [x] Add dynamic resources for:
  - [x] font size
  - [x] icon size
  - [x] margin/padding
- [x] Trick `SfSkinManager` to dynamically change theme settings
  - It somehow breaks window's rendering when settings are reloaded. Minimising/maximising window workarounds this issue. Either show a message box to user, that WKit needs to restart, or find a fix to prevent greyish background...
- [ ] Define scale based on screen's height at startup, when not already defined manually by the user.
> Use height only to account for wider screen ratio.
- [x] Add IconBox to provide uniform boxing/scaling.
- [x] HomePage
  - [x] `HomePageView`
  - [x] [WelcomePageView](https://github.com/user-attachments/assets/74473f7a-101b-420a-a410-d94b5c16973a)
  - [x] [SettingsPageView](https://github.com/user-attachments/assets/167b250b-6194-4bcd-8015-1b287f3d742c)
    - changed icon in favor of outline style
    - changed fields submit event to 'focus out/enter'
    - limit `Scale UI (%)` between 100% and 900%
  - [x] [PluginsToolView](https://github.com/user-attachments/assets/d9de4853-daf1-4b66-8093-5bbf567cc12f)
  - [x] [ModsView](https://github.com/user-attachments/assets/e7f9650c-c0cc-4e2f-acd5-bdea22fdc082)
    - moved buttons in header (like other Home views)
- [x] Shell
  - [x] `BackStageView`
  - [x] [DockingAdapter](https://github.com/user-attachments/assets/28f4617c-9966-4df3-b282-2a41f750fdf3)
  - [x] `MainView`
  - [x] [MenuBarView](https://github.com/user-attachments/assets/d68e9430-1b96-4c40-8dd4-e72afcb764c7) (*screenshot not up-to-date*)
    - changed icons in favor of outline style
    - uniformed colors
  - [x] [RibbonView](https://github.com/user-attachments/assets/c9cb79d3-772c-4b3c-9f2d-6514eed032e1)
    - changed icons in favor of outline style
    - uniformed colors
  - [x] `Splash`
  - [x] `StatusBarView`
- [ ] Tools
[...]

PR #1983, opened 2024-10-29
Update nugets

rfuzzo:
Update nugets


poirierlouis:
After a quick look, only UI issue I found is SOURCE and RESOURCES not showing up:
image
Should be related to how I fixed it in #1954.

PR #1984, opened 2024-10-29
Draft: feature/npv

manavortex:
# NPV: creation menu, guided dialogue


seberoth:
Just my opinion: Don't think we should mix "New file" dialogue (where I select a filetype knowing what it is for) with dialogues which creates a file (or even multiple / different file types) based on some wizard.

Maybe something like Toolbar:
 - "File" 
   - "New File"
     - "File..."
     - "Wizard"
       - "Create NPV..."

manavortex:
Sure, sounds good to me :)

PR #1985, opened 2024-10-29
feature/morphtarget with mesh materials

manavortex:
# morphtarget: use mesh materials

**Implemented:**
- Exporting a morphtarget should export the mesh file's materials

~Requires https://github.com/WolvenKit/Cyberpunk-Blender-add-on/pull/178 to be merged~

### DIFF
--- a/WolvenKit.App/Helpers/ProjectResourceHelper.cs
+++ b/WolvenKit.App/Helpers/ProjectResourceHelper.cs
@@ -51,16 +51,6 @@ public static class ProjectResourceHelper
         ).GetRed4Controller();
     }
 
-    public static string SwitchArchiveRaw(string filePath)
-    {
-        if (filePath.Contains(ArchiveSubdirWithSlashes))
-        {
-            return filePath.Replace(ArchiveSubdirWithSlashes, RawSubdirWithSlashes);
-        }
-
-        return filePath.Replace(RawSubdirWithSlashes, ArchiveSubdirWithSlashes);
-    }
-
     private static string GetUniqueSubfolderPath(IEnumerable<string> allFilePaths, string currentFilePath)
     {
         var fileName = Path.GetFileName(currentFilePath);
--- a/WolvenKit.App/Helpers/ProjectResourceHelper.cs
+++ b/WolvenKit.App/Helpers/ProjectResourceHelper.cs
@@ -25,6 +25,10 @@ namespace WolvenKit.App.Helpers;
 
 public static class ProjectResourceHelper
 {
+    public static readonly string SourceSubdirWithSlashes = $"{Path.DirectorySeparatorChar}source{Path.DirectorySeparatorChar}";
+    public static readonly string ArchiveSubdirWithSlashes = $"{Path.DirectorySeparatorChar}archive{Path.DirectorySeparatorChar}";
+    public static readonly string RawSubdirWithSlashes = $"{Path.DirectorySeparatorChar}raw{Path.DirectorySeparatorChar}";
+    
     private static IProjectManager? s_projectManager;
     private static IProjectManager? GetProjectManager() => s_projectManager ??= Locator.Current.GetService<IProjectManager>();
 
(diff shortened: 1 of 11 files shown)


seberoth:
Not sure why you added export / modkit behaviour to Wkit.App / ImportExportHelper.
Should stay in Modkit ideally

manavortex:
> Not sure why you added export / modkit behaviour to Wkit.App / ImportExportHelper. Should stay in Modkit ideally

@Seberoth: because adding the mesh file to the project is much harder from inside the modkit, unless I'm overlooking something?

seberoth:
Why do you want to add the mesh to the project?
The current modkit morphtargetexport uses the baseMesh without adding it too. What prevents us from doing the same for the textures?

If you want to fully support it like you are trying here, what about the remaining mesh export args?

If you really want to add cascading export, we should discuss this / think of a proper way instead of duplicating all modkit code win Wolvenkit.App .

manavortex:
It doesn't use the basemesh, the morphtarget has its own 3d data (as far as I know). And morphtargets may define their own materials via baseMaterial, but the game will ignore that property completely and pull the material definitions from the .mesh, that's why I want them to use these in Blender as well

manavortex:
Oh - "it" was being "the game", which was clear in my head. 

Morphtargets don't have LODs, so that's not necessary - and we don't want GarmentSupport there, since that would also end up being a shapekey, and the morphtarget magic gets imported as a shapekey.

seberoth:
Then I still would suggest adding an "Export materials" setting to morphtarget and let modkit handle it.

manavortex:
How, though? The morphtarget doesn't have the materials, since these are in the mesh. I'm not sure where the pngs come from, but the game doesn't use them

manavortex:
It has one - that exports the pngs, which are incorrect

seberoth:
Could add it here. Would also allow it to be called from CLI and you don't need to add temp file somewhere.

https://github.com/WolvenKit/WolvenKit/blob/60352ce3ed5e777d867660e81ac56b8efa7119c8/WolvenKit.Modkit/RED4/Tools/MorphTargetTools.cs#L43

Something like:
```cs
if (exportMaterials) 
{
    ExportMaterials(result.File, outfile, new MeshExportArgs { ... });
}
```

manavortex:
Is a good idea, thanks :) I'm currently too braindead for anything, but I'll try putting it there the next time I look at this.
If you feel like fixing this up, be my guest!

WolvenKit.App/Helpers/ProjectResourceHelper.cs
```diff
@@ -47,6 +51,16 @@ private static RED4Controller GetRed4Controller()
         ).GetRed4Controller();
     }
 
+    public static string SwitchArchiveRaw(string filePath)
```
seberoth, line 54: Not used anywhere.
What is this for? Can it be removed again?

WolvenKit.Modkit/RED4/Tools/MorphTargetTools.cs
```diff
@@ -76,9 +85,6 @@ public bool ExportMorphTargets(
                 }
                 expTargets.Add(ContainRawTarget(diffsBuffer, mappingBuffer, tempNumVertexDiffsInEachChunk, tempNumVertexDiffsMappingInEachChunk, targetsInfo.TargetStartsInVertexDiffs[i], targetsInfo.TargetStartsInVertexDiffsMapping[i], targetsInfo.TargetPositionDiffOffset[i], targetsInfo.TargetPositionDiffScale[i], expMeshes.Count));
             }
-
-            // Only add textures to list if the setting is checked in the export
-            var textureStreams = exportTextures ? ContainTextureStreams(blob, texBuffer) : [];
```
seberoth, line 81: Aren't these textures used for anything or why we can ignore them?

manavortex: They were only used for import in the Blender addon for morphtargets. I guess if we yeet them, they can rip out that part of their code as well

PR #1993, opened 2024-10-31
meshMeshAppearance: Show chunks inline

manavortex:
# meshMeshAppearance: Show chunks inline


poirierlouis:
I saw the screenshot, might be nice to put here...
Could it wait for #1954 to provide more editors?

poirierlouis:
Cool, yeah Vector2 existed but wasn't used in TemplateSelector... And I saw other types which could have dedicated editors too.
Should be easier to layout things using resources of scaling PR imo.

PR #2005, opened 2024-11-10
Fixed nugets not including required files

seberoth:
# Fixed nugets not including required files

Could also use `<PackagePath>lib\net8.0</PackagePath>` for dlls placed in the main dir, but won't work for other files (json, so, ...)

Also included `;content`. Only used in (really) old csprojects (before nuget 3.3 was introduced). Not sure if we need to keep it. (Also duplicates these files in the nuget)

`<NoWarn>NU5104</NoWarn>` can be removed as soon as the nuget update is done. Thats just for the prerelease version of SharpGltf we currently use.

Also bumped version so we can release them.


manavortex (approved):
LGTM

PR #2008, opened 2024-11-10
fix: Don't analyze archives when loading via cli

seberoth:
fix: Don't analyze archives when loading via cli


poirierlouis:
LGTM.

poirierlouis:
I wonder, as I'm not use to cli, no subsequent tasks/features requires archive files to be analyzed?

seberoth:
> I figure Tasks are only related/used with CLI?

Yeah. Could also expose that flag, but would do that as a new feature (to allow it for mod archives for example).
But for now just disabling it. Should be the default bahviour anyway

poirierlouis:
Alright thanks for the clarification. Yes make sense to add such feature in another PR.

PR #2009, opened 2024-11-11
fix: "Open in new tab" checking wrong property for already opened files

seberoth:
# fix: "Open in new tab" checking wrong property for already opened files

**Fixed:**
- "Open in new tab" checking wrong property for already opened files

Did check the full path before. Now checks the relative path.


poirierlouis:
LGTM

PR #2010, opened 2024-11-11
fixing regenerate visual controllers

manavortex:
# fixing regenerate visual controllers


poirierlouis:
Looks more like a refactor than a fix no?

manavortex:
I fixed it by renaming the command to what's actually called... so maybe both :D

poirierlouis:
Oh, the command used in XAML was not Visible, but Visual. Hence the fix?

manavortex:
Yeah, it's supposed to be Visual :D

PR #2020, opened 2024-11-15
build: disable NuGet audit

poirierlouis:
# build: disable NuGet audit

**Implemented:**
- temporarily disable NuGet audit to build project even with vulnerabilities in package dependencies.
- should be removed after upgrading dependencies.


poirierlouis:
It should be better to merge #1983 instead.

PR #2021, opened 2024-11-16
Added reader/writer for Razer Chroma animations

seberoth:
# Added reader/writer for Razer Chroma animations

**Implemented:**
- Added reader/writer for Razer Chroma animations

**Fixed:**
- Reverted unwanted changes

**Additional notes:**
Could add some kind of helper to CVM to work with this file.
Changing `Device` should also adjust the Frames array.


rfuzzo (approved):
👍

PR #2023, opened 2024-11-17
put WindowsSdkPackageVersion into .csproj file to fix build error

manavortex:
# put WindowsSdkPackageVersion into .csproj file to fix build error

Any reasons not to merge this back? It was the only thing that fixed it for me


rfuzzo:
Microsoft doesn't recommend this:
```
We don't recommend overriding the Windows SDK version, because the Windows SDK targeting packages are included with the .NET 5+ SDK. Instead, to reference the latest Windows SDK package, update your version of the .NET SDK.
```

To fix the error on build, update your .NET SDK.

PR #2074, opened 2024-12-10
fixing animgraph crash

manavortex:
# fixing animgraph crash

- made recursive CVM property loading more robust and pulled it out of the accessor functions into calling classes


manavortex:
Not tested yet

PR #2079, opened 2024-12-15
T pose updated

kainotoa:
# $PULL_REQUEST_TITLE

**Implemented:**
- rig importing
- mesh inverse binding importing
- animation export with T-Pose rig


manavortex:
Hitman and eagul are on this :)

eaguull:
Ran some simple tests, importing does change rig and mesh matrices, but calculates stuff wrong in both. Identical resulting values to the first version of the code I tested. Certain values are swapped places in the matrices, or negative when they shouldn't be.

PR #2085, opened 2024-12-18
convenience when copying/moving file on itself

manavortex:
# ProjectExplorer: Convenience when copying/moving a file on itself

WKit will ask for a new file name.
- if file name is empty (dialogue cancelled), operation aborts
- if file name is identical, it will create a _copy


seberoth (approved):
LGTM

PR #2090, opened 2024-12-20
export tool: disabling without files

manavortex:
# export tool: show message if no files are found

image


seberoth (approved):
LGTM.
Only thing I noticed (but not related to this change) is, if you open/enable the export tool and its a floating window, the files doesn't get loaded.
Only after clicking the window (marking it as active) this happens.
Showed an empty list before and now your message. Could confuse some people.
But again, thats a different issue.

PR #2092, opened 2024-12-21
convert to photo mode app: no longer destroys original file

manavortex:
# convert to photo mode app: no longer destroys original file


WolvenKit.Common/Services/HashService.cs
```diff
@@ -206,6 +206,10 @@ private void ReadTweakNames(Stream memoryStream)
             TweakDBIDPool.SetNative(dict);
         }
 
+        public virtual bool AddResourcePath(string resourcePath) => false;
+
+        public virtual bool AddTweakName(string tweakName) => false;
```
seberoth, line 211: Why adding that here?
Its currently not supported to add custom ResourcePath/TweakNames in anything except Wkit GUI.

WolvenKit.Modkit/RED4/Tools/DocumentTools.cs
seberoth, line 1: Why factor that out of `RedDocumentViewModel` when its the only place its used for?

WolvenKit.App/ViewModels/Documents/RedDocumentViewModel.cs
```diff
@@ -189,74 +190,15 @@ public override async Task Save(object? parameter)
...
-                {
-                    continue;
-                }
-
-                hashService.AddTweakName(tweakName);
-            }
-        }
     }
 
     public override void SaveAs(object parameter) => throw new NotImplementedException();
```
seberoth, line 204: Could handle your SaveAs here (as intended).
Currently not sure what parameter is used for, but could also add a filename arg and use the instead of `FilePath`.
Could even refactor `Save` to call `SaveAs(parameter, FilePath)` then.

WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
```diff
@@ -802,27 +844,21 @@ private void OnConvertToPhotoModeClick(object sender, RoutedEventArgs e)
                 return;
             }
 
-            if (RootChunk.Tab is null || _projectManager.ActiveProject is null || RootChunk.Tab.FilePath is not string s || !File.Exists(s))
+            if (RootChunk.Tab?.Parent.Cr2wFile is null ||
+                !DocumentTools.SaveCr2W(RootChunk.Tab.Parent.Cr2wFile, absolutePath))
```
seberoth, line 848: After before mentioned changes, could just call `RootChunk.Tab.Parent.SaveAs.SafeExecute(...);` here

WolvenKit.App/ViewModels/Documents/RedDocumentViewModel.cs
manavortex: I figured I'd put it into the interface at the top level. Didn't think much about it. Can you elaborate on your other idea?

seberoth: 1. Move the code from `Save(object parameter)` to `SaveAs(object parameter)`
2. Change `SaveAs(object parameter)` to `SaveAs(object parameter, string targetPath)`
3. Replace `FilePath` with `targetPath` in the code
4. Change `Save(object parameter)` to `Save(object parameter) => SaveAs(parameter, FilePath);`

WolvenKit.Modkit/RED4/Tools/DocumentTools.cs
manavortex: I can think of a few other cases where it'd be good to have this. I can put it back if you want?

seberoth: K, if you want to have a Save feature outside of `RedDocumentViewModel`. Sure, do it.
But at least do it so that the original function doesn't change.
Haven't tested your code yet, but at least the `OnSave` hook is missing now

WolvenKit.Common/Services/HashService.cs
manavortex: If it's not in there, I can't have the DocumentTools under Tools, where they have to be to find other dependencies... it's turtles all the way down, I'm afraid

seberoth: Oh, didn't saw that you placed `DocumentTools` in `ModKit`.
Still not convinced of the usefullness of it there.
Looks like an class which is meant to be used as part of GUI (Project context) in a namespace which shouldn't really know the GUI / has projects.

manavortex: moved it out :)

manavortex:
I'll re-do this with the new changes from psi :)

PR #2093, opened 2024-12-21
0 in vector's V will be treated like 1

manavortex:
# 0 in vector's V will be treated like 1


seberoth:
Would prefer to use Vector3 for `zUpNormalDelta` and `zUpTangentDelta` (and then `Vec3ToU32`) instead of changing Vec4ToU32.
Vec4ToU32 is meant tu be used for mesh tangents, which only allow 1 or -1 as w component.
Its my fault that this isn't yet since I changed Vec4ToU32. So feel free to asign me a ticket for it.

https://github.com/WolvenKit/WolvenKit/blob/75a122752a9575ccf3d7d8a7333cfe132bc674bc/WolvenKit.Modkit/RED4/Tools/MorphTargetImportTools.cs#L263-L276

PR #2101, opened 2024-12-28
Skip class selection for handles when only one type is allowed

### LINKED ISSUE #1978: Automatically create handle when only one type is allowed
Currently, creating a handle will show a dialog to select an existing type, or in some cases define a custom new type.
In most cases, type of the handle is already known and expected. Yet, the dialog is still shown to the user, whom can only select one type and create it.

When only one type is allowed (and no custom new type), dialog should not be shown and directly forward the unique type allowed to create the handle for.
This would speed up the process for the end user.

wkit_fix_ux_create_handle

seberoth:
# Skip class selection for handles when only one type is allowed

**Implemented:**
- Skip class selection for handles when only one type is allowed

**Additional notes:**
Fixes #1978


manavortex (approved):
LGTM!

PR #2103, opened 2024-12-28
#2062 #2002 descriptors and values for cvm

### LINKED ISSUE #2002: inkAnim: Show preview to target in main array without open each one
Show path target without opening it like in mesh

inkAnim: target
image

inkAnim: interpolators
image

v_sport1_rayfield_caliburn__ext01_hood_01.mesh
image

### LINKED ISSUE #2061: Make relative position and rotation index connection with their vectors more obvious.
So in the compiledEffectInfo of an effect handle there's a placement infos array. The entries in that array have a relativePositionIndex entry and a relativeRotationIndex entry. The value used in these correspond to the vector they will use in the relativePositions and relativeRotations arrays. So for example relativePositionIndex = 0 will use the vector at the 0 position.

WolvenKit_6aSFCIYL6x

manavortex:
# descriptors and values for cvm

closes #2061 
closes #2002

### DIFF
--- a/WolvenKit.App/Helpers/StringHelper.cs
+++ b/WolvenKit.App/Helpers/StringHelper.cs
@@ -3,6 +3,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Text.RegularExpressions;
+using System.Windows.Media;
+using DynamicData;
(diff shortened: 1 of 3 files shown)


dragonzkiller (approved):
Looks good to me

Issue #1957, opened 2024-10-12
Color Curve Editor Enhancements


CyanideX:
When editing colors, skydomeColor and skylightColor specifically, it's always a challenge to visualize the daylight curve while also adjusting the light intensity (alpha) and editing the colors efficiently.

My suggested improvements are threefold: improve the UI to show data more effectively while also moving various controls to a single interface, add key information to the element list for easier navigation, and implement the addition of new elements on the color curve editor via double-click, similar to how it works with the others.

Improve the UI for ease of use:
![image](https://github.com/user-attachments/assets/f083094d-6ee4-43e6-909b-2fae0ed6c603)

- Includes horizontal repositioning of colors, color editor, and vertical, locked-axis slider for alpha values.

Add key information to the element list:
![image](https://github.com/user-attachments/assets/304d97bd-6767-4a2d-b765-b606f9acb1b0)

Double-click to add new color points to the curve:
![image](https://github.com/user-attachments/assets/032a6fea-23af-425f-b317-911321fd2fd0)

poirierlouis:
Looks fine by me, I'm not used to this part of WKit yet. But it should be possible and I could get my hands on it after #1954.

CyanideX:
![image](https://github.com/user-attachments/assets/16a3cf97-cccf-489d-8e32-48c38bc1f283)

This new color scheme makes it very difficult to use the curve editor efficiently. Could we get a lighter background for the graph, please? 🙂

poirierlouis:
@CyanideX should the graph line be white instead to keep a dark theme and be coherent with WKit theme? Would it be good enough?

CyanideX:
Yes! If you could also reduce the brightness of the grid, similar to Resolve's curve editor below, that would be perfect.

![image](https://github.com/user-attachments/assets/eb006268-fe4c-4fc0-b21f-44cc93bd4ee1)

poirierlouis:
@CyanideX what do you think about this?

![image](https://github.com/user-attachments/assets/9b4b0054-0e09-4cb5-b756-0ff503202a73)

poirierlouis:
Edit: (B) with little tweaks...

![image](https://github.com/user-attachments/assets/8d7e3aea-6194-4154-ae0b-8764d3a13577)

CyanideX:
B is looking very nice! Thanks for making these changes.

poirierlouis:
Thanks for your feedback! I wrote a PR, should be ready in nightly this week ;)

PR #2108, opened 2025-01-02
feat(CurveEditor): improve contrast ratio

poirierlouis:
# feat(CurveEditor): improve contrast ratio

**Implemented:**
- remove point's outline
- increase curve's line width, change to white color
- add color resources with `WolvenKitGraph...` name

**Fixed:**
- see [feedback](https://github.com/WolvenKit/WolvenKit/issues/1957#issuecomment-2566686742). Issue in itself is related to the same view but for others, bigger changes (so not closing here).

**Demo:**
Screenshot at 100%. Currently line width (grid + axis + curve) are not responsive, but point's size is. Can a 4K user try it and let me know if it would benefit UI scaling?

image


seberoth (approved):
LGTM

PR #2109, opened 2025-01-02
feat(RedTree/RedType): stick buttons on the right

### LINKED ISSUE #1991: Remove button's description in RedTreeView based on Editor Difficulty Level
Currently in the RedTreeView, buttons to `Add Handle`, `Create Item in Array`, etc. are shown with an icon and a description of the action.
Icon of the button should be enough when user knows WKit. That is, hide the description to improve the UX of this view.

Hide the description based on Editor Difficulty Level (Editor Mode). Only shows an icon, like other shortcuts to `Open in new tab` or to `Add asset in project`.

Another idea would be to move such buttons to the left edge of the tree, gutter-like. This would make it easier for all users to take actions based on a row, without requiring to resize the RedTreeView. Currently the view cannot be scrolled horizontally, user needs to increase the width to see buttons at the far right.
Let me know if this idea should be put in another ticket, to discuss it.

poirierlouis:
# feat(RedTree/RedType): stick buttons on the right

**Implemented:**
- remove labels of buttons, keep tooltip.
- stick buttons on the right, on top of the row.
- `Open tab` + `Add to project` buttons in `RedTreeView`.
- `Create handle` + `Delete` buttons in `RedTypeView`.

**Additional notes:**
Closes #1991

**Demo:**
2025-01-02_WKit_Stick

### DIFF
--- a/WolvenKit/Views/Templates/RedTypeView.xaml
+++ b/WolvenKit/Views/Templates/RedTypeView.xaml
@@ -56,19 +56,12 @@
                                          Converter={StaticResource BooleanToVisibilityConverter}}"
                     Command="{Binding AddHandleCommand}"
                     ToolTip="Add Handle">
-                    <StackPanel Orientation="Horizontal">
-                        <templates:IconBox
-                            IconPack="Material"
-                            Kind="PlusCircle"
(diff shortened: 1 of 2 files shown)


manavortex (approved):
Works swimmingly

PR #2110, opened 2025-01-03
Mesh fix

seberoth:
# Mesh fix

**Fixed:**
- Disabled UV parameter generation (introduced in ccb65ce ) since its causing issues with mesh seams.

**Notes:**
Normally these buffers contains `TEXCOORD_0` of the meshes, but some values in there are not normalized.
Maybe tackle that topic again later


manavortex (approved):
Would be great if you could put a comment to why it was disabled before merging :)

seberoth:
> since its causing issues with mesh seams

Also see referenced issue

PR #2121, opened 2025-01-06
Rename: no capital letters

manavortex:
# rename dialogue: changing capitalization will no longer delete file

I also disabled capitalization because we hates it, precious


rfuzzo (changes requested):
Right click > Rename incorrectly duplicates files in the project explorer (although it is correct in windows explorer)
image
image

PR #2122, opened 2025-01-06
update project references: replace in entire project

manavortex:
# update project references: replace in entire project

- will also replace in .csv and .json cr2w files
- will also replace in raw text files under resources

### DIFF
--- a/WolvenKit.App/Helpers/ProjectResourceTools.cs
+++ b/WolvenKit.App/Helpers/ProjectResourceTools.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using DynamicData;
 using Splat;
 using WolvenKit.App.Controllers;
 using WolvenKit.App.Interaction;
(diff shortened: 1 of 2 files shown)


WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -485,6 +486,49 @@ public void ReplacePathInProject(Cp77Project? activeProject, ResourcePath oldPat
...
 
+
+        if (oldPath.GetResolvedText() is string oldPathStr && !string.IsNullOrEmpty(oldPathStr) &&
+            newPath.GetResolvedText() is string newPathStr && !string.IsNullOrEmpty(newPathStr))
+        {
+            var oldPathWithForwardSlashes = oldPathStr.Replace("\\", "/");
+            var newPathWithForwardSlashes = newPathStr.Replace("\\", "/");
+            
+            var resourceFiles =
+                Directory.GetFiles(activeProject.ResourcesDirectory, "*.*", SearchOption.AllDirectories);
```
rfuzzo, line 497: This should only get specific text files in resources with known extensions imo: .lua, .script, .tweak, .yaml, .reds and .json

As it is now, it would also replace stuff in non-text files

rfuzzo (changes requested):

PR #2123, opened 2025-01-06
Added sleep functionality to wscripts

notaspirit:
# $Added sleep functionality to wscripts via AppScriptFunctions

**Implemented:**
- sleep functionality to wscripts

**Fixed:**
- none

**Additional notes:**
not 100% sure that 0x2000 is the correct exception variable but from testing it works fine and the other methods in that same class use it as well


notaspirit:
@seberoth 
I was told you are the best person to review this

poirierlouis (commented):
I think it is better to simply return instead of throwing an exception.

seberoth (approved):
LGTM

PR #2126, opened 2025-01-07
fix: crash scene view on the unexisting connection

Sun1ive:
# Crash scene on referencing to the non existing nodes

**Implemented:**
- Added try catch block
- Added explicit error logging

**Fixed:**
- Scene crash

**Additional notes:**
This is my first contribution. Any review and tips about project structure are appreciated.

image

### DIFF
--- a/WolvenKit.App/ViewModels/GraphEditor/RedGraph.Scene.cs
+++ b/WolvenKit.App/ViewModels/GraphEditor/RedGraph.Scene.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text.RegularExpressions;
 using WolvenKit.App.ViewModels.Documents;
 using WolvenKit.App.ViewModels.GraphEditor.Nodes.Scene;
 using WolvenKit.App.ViewModels.GraphEditor.Nodes.Scene.Internal;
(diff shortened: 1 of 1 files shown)


WolvenKit.App/ViewModels/GraphEditor/RedGraph.Scene.cs
```diff
@@ -370,22 +372,32 @@ public static RedGraph GenerateSceneGraph(string title, scnSceneResource sceneRe
 
                 foreach (var destination in sceneOutputConnector.Data.Destinations)
                 {
-                    var targetNode = nodeCache[destination.NodeId.Id];
-                    if (targetNode is IDynamicInputNode dynamicInputNode)
+                    try
                     {
-                        while (dynamicInputNode.Input.Count <= destination.IsockStamp.Ordinal)
+                        var targetNode = nodeCache[destination.NodeId.Id];
```
manavortex, line 377: Could you check if nodeCache has the key (maybe with TryGet) instead of the try/catch? It would be better to prevent the error alltogether rather than pushing it under the carpet

Sun1ive: it does work fine indeed. but there is another question. how to notify the user then?
image
im not sure its gonna work well in the game

manavortex (approved):
LGTM!

PR #2130, opened 2025-01-08
CVM: descriptors and values for streamingsectors and -blocks

manavortex:
# CVM: descriptors and values for streamingsectors and -blocks

After poking around in the files, I have a better idea what properties make sense to display


rfuzzo (approved):
lgtm

PR #2134, opened 2025-01-09
dialogue: edit app components

manavortex:
# dialogue: edit app components

allows cloning and deleting components by name

### DIFF
--- a/WolvenKit.App/Helpers/DocumentTools.cs
+++ b/WolvenKit.App/Helpers/DocumentTools.cs
@@ -35,11 +35,9 @@ public class DocumentTools
 
     # region appfile
 
-    public List<string> GetAllComponentNames(List<appearanceAppearanceDefinition> appearances)
-    {
-        return appearances.SelectMany(app => app.Components)
+    public static List<string> GetAllComponentNames(List<appearanceAppearanceDefinition> appearances) => appearances
+        .SelectMany(app => app.Components)
             .Select(c => c.Name.GetResolvedText() ?? "").Where(c => !string.IsNullOrEmpty(c)).Distinct().ToList();
-    }
     
     public List<string> ConnectAppToEntFile(string absoluteAppFilePath, string absoluteEntFilePath,
         bool clearExistingEntries = false)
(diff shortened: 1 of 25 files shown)


rfuzzo:
What does this do exactly? or, how can I test this best?

WolvenKit.App/Helpers/DocumentTools.cs
```diff
@@ -34,6 +35,12 @@ public DocumentTools(ILoggerService loggerService, Cr2WTools cr2wTools)
 
     # region appfile
 
+    public List<string> GetAllComponentNames(List<appearanceAppearanceDefinition> appearances)
```
rfuzzo, line 38: this should be a static function, or moved to RedDocumentViewMenuBar.xaml.cs as private func

rfuzzo (changes requested):
Works almost great!

but this is broken:
1. I duplicate component no 30
image
2. But the copy already exists, so the name is the same 
3. It also shows the xpath as 30, but the view shows properties of 31
image

rfuzzo:
image
also, when changing variables in the right view right after creating the new component, the changes are not saved

rfuzzo:
The issue with the incorrect indices still persists

repro:
1. open any app file
2. duplicate the first component with the new dialogue
3. notice that it says `root.appearances[0].components[0]` in the top right but the name doesn't match (component 0 is the new copy)
4. notice also that editing any field in the right hand side doesn't update the values until you click away, click back into it, and change somethin in the field again

image

manavortex:
@rfuzzo: Fixed it for real now :D

rfuzzo:
So, yeah I verified it now has the proper object selected but 
- the tree view selection isn't updated
- and editing a value (e.g. the name of the component) doesn't update on the left treeview side properly 

Both of these problems are because inside Wkit the programmatic selection of items inside a SfTreeView doesn't seem to work properly.

This is unrelated to this PR, but it affects it

manavortex:
I'm now clearing the selection until I can get behind how the other thing workws

PR #2139, opened 2025-01-10
added logging about scene graph initialization crash

Sun1ive:
# added logging about scene graph initialization crash

**Implemented:**
- Added log when the sceneGraph was not created. Before that it just silently didn't show up.

**Fixed:**
- Added logging

**Additional notes:**
image

### DIFF
--- a/WolvenKit.App/ViewModels/GraphEditor/RedGraph.Scene.cs
+++ b/WolvenKit.App/ViewModels/GraphEditor/RedGraph.Scene.cs
@@ -338,25 +338,35 @@ public partial class RedGraph
         var graph = new RedGraph(title, sceneResource);
 
         var nodeCache = new Dictionary<uint, BaseSceneViewModel>();
-        foreach (var nodeHandle in sceneResource.SceneGraph.Chunk!.Graph)
+        var sceneGraph = sceneResource.SceneGraph?.Chunk!.Graph;
+        if (sceneGraph != null)
         {
-            ArgumentNullException.ThrowIfNull(nodeHandle.Chunk);
 
-            var node = nodeHandle.Chunk;
-
-            var nvm = graph.WrapSceneNode(node);
-
-            if (!nodeCache.ContainsKey(nvm.UniqueId))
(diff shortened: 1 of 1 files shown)


Sun1ive:
maybe better to fix that somehow. but idk how yet

Sun1ive:
@poirierlouis Thanks for the changes in the pr. 
There is another case about the scenes I'm looking at. 
This pr can be merged if you think it fits.

WolvenKit.App/ViewModels/GraphEditor/RedGraph.Scene.cs
```diff
@@ -338,26 +338,35 @@ public static RedGraph GenerateSceneGraph(string title, scnSceneResource sceneRe
...
+
+                    graph._currentSceneNodeId = Math.Max(graph._currentSceneNodeId, nvm.UniqueId);
+                }
+                else
+                {
+                    _loggerService?.Warning("Duplicate node ID: " + nvm.UniqueId.ToString() + ". Some nodes may be missing in graph view. File: " + title);
+                }
             }
         }
+        else
```
manavortex, line 366: Can I ask you to put the else case first? It's a strong preference of mine to have the "what happens if this is invalid/broken" case at the top, because it's usually much shorter.

WolvenKit.App/ViewModels/GraphEditor/RedGraph.Scene.cs
```diff
@@ -338,26 +338,35 @@ public static RedGraph GenerateSceneGraph(string title, scnSceneResource sceneRe
...
+        {
+            _loggerService?.Error("Scene graph failed to initialize. Check that the sceneGraph is created.");
+        }
 
         foreach (var node in graph.Nodes)
```
manavortex, line 371: Does it even make sense to initialize the graph if the sceneGraph is invalid? (Genuine question, because I don't know.)
If not, you could maybe `return graph` in the `if sceneGraph is null` case and save a bunch of `else`s

PR #2142, opened 2025-01-10
feature: remember open files

manavortex:
# feature: remember open files

Projects can remember the last X open files and reopen them on project launch

### DIFF
--- a/WolvenKit.App/Models/ProjectManagement/Project/Cp77Project.cs
+++ b/WolvenKit.App/Models/ProjectManagement/Project/Cp77Project.cs
@@ -360,7 +360,6 @@ public sealed partial class Cp77Project : IEquatable<Cp77Project>, ICloneable
 
     private const string s_projectFilesDirName = ".projectFiles";
     private const string s_projectTreeStateFileName = "fileTreeState.json";
-    private const string s_openFilesList = "openFilesList.json";
 
     /// <summary>
     /// Path to <see cref="ProjectDirectory"/>/<see cref="s_projectFilesDirName"/>, where we store temp files
@@ -402,6 +401,11 @@ public sealed partial class Cp77Project : IEquatable<Cp77Project>, ICloneable
                 {
                     File.Delete(oldPath);
                 }
+            }
+
+            if (!File.Exists(newPath))
+            {
+                File.WriteAllLines(newPath, ["{}"]);
             }
 
             return newPath;
--- a/WolvenKit.App/Models/ProjectManagement/Project/Cp77Project.cs
+++ b/WolvenKit.App/Models/ProjectManagement/Project/Cp77Project.cs
@@ -11,7 +11,6 @@ using WolvenKit.Common;
 using WolvenKit.Core.Extensions;
 using WolvenKit.Core.Interfaces;
 using WolvenKit.Core.Services;
-using WolvenKit.Modkit.RED4.Serialization;
 using WolvenKit.RED4.Archive.CR2W;
 using WolvenKit.RED4.Archive.IO;
 using WolvenKit.RED4.Types;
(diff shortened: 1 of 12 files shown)


WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -1076,19 +1081,95 @@ private void SetupToolDefaults() =>
...
+            .TakeLast(_settingsManager.NumFilesToReopen)
+            .ToList();
+
+        foreach (var path in lastFilePaths)
+        {
+            _appViewModel.RequestFileOpen(path);
+        }
+
+        // write back the last X entries
+        File.WriteAllLines(lastFilesPath, lastFilePaths);
```
seberoth, line 1142: Throws an exception here, because the file is already open.

seberoth (changes requested):

manavortex:
there _will_ be people with 617236 open files who put MAX_INTEGER as value, easier to tell them to just delete one file

rfuzzo:
Hmm, I'm not convinced this is a good reason (it's equally as easy to tell someone to change the max number). I think it is more desirable to keep all project-related settings in one file and not clutter the project folder.

manavortex:
That's why I made the subfolder ;)

I'm currently logging _all_ currently open tabs every time a tab gets opened or closed to save the effort of filtering the file. In my experience, there's a (very) low chance of corruption on every I/O operation, and I'd much rather not fuck up the project file because somebody found a way to put an unescaped backslash in a file name. This is just me being paranoid, though. If you insist, I'll put them in the cdproj file, but I'd feel :s about it

rfuzzo:
When you open an existing project now, the project explorer is empty

rfuzzo (changes requested):
Bug: When you open an existing project now, the project explorer is empty. (Old Comment moved here)

manavortex:
I can't reproduce the bug :/ after project explorer is done loading, it gets populated immediately

rfuzzo:
> I can't reproduce the bug :/ after project explorer is done loading, it gets populated immediately

Happened to me on an old project created before this feature.

manavortex:
I tried that as well, of course - on several projects. They all loaded fine.

rfuzzo:
It happens 100% of all times.
Repro steps:
1. Delete the `.projectFiles` folder in your mod
2. load it in wolvenkit
3. notice that the project explorer is empty

image

WolvenKit.App/Models/ProjectManagement/Project/Cp77Project.cs
```diff
@@ -340,6 +356,60 @@ public string PackedLegacyModDirectory
     }
 
 
+    #region project_state_files
+
+    private const string s_projectFilesDirName = ".projectFiles";
+    private const string s_projectTreeStateFileName = "fileTreeState.json";
+    private const string s_openFilesList = "openFilesList.json";
```
rfuzzo, line 363: this is unused

WolvenKit.App/Services/ISettingsManager.cs
```diff
@@ -192,5 +192,7 @@ public static List<SaveGame> GetSaveGames()
 
     bool ReopenLastProject { get; set; }
 
+    int NumFilesToReopen { get; set; }
```
rfuzzo, line 195: There should be an option "Reopen Files on launch" in addition to the number of files to reopen imo

rfuzzo (changes requested):

rfuzzo (changes requested):
bug: when closing a project with no open files, it still saved the old file - resulting in reopening of files that weren't open upon close

rfuzzo (approved):
works on my machine now XD

PR #2147, opened 2025-01-11
replace in project files: better exclusion criteria

manavortex:
# replace in project files: better exclusion criteria

Replace in project files will now filter for files that don't need to be checked for replacements (should speed things up)


rfuzzo (approved):
looks good

PR #2169, opened 2025-01-13
[main] refactor/cvm => simplifications & codestyle compliance

manavortex:
# Refactor: CVM

I recommend reviewing this by commit :)

## logic
- inverted ifs
- moved methods into correct partial class
- changed big if/else blocks to switch statements

## compiler suggestions
- removed redundant checks and casts
- merged checks into pattern

## OCD
- typos / camel casing
- added () around expressions for clarity


WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -656,20 +656,14 @@ public Type PropertyType
     {
         get
         {
-            var type = Data?.GetType() ?? typeof(RedDummy);
+            var type = Data.GetType();
```
rfuzzo, line 659: this seems to have lost the reddummy?

rfuzzo (changes requested):

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
manavortex: Since Data is never null, there's no need to fall back to RedDummy

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -471,12 +471,12 @@ partial void OnDataChanged(IRedType value)
 
     #region Properties
 
-    public ObservableCollectionExtended<ChunkViewModel> SelfList { get; set; } = new();
+    public ObservableCollectionExtended<ChunkViewModel> SelfList { get; set; }
```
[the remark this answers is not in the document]
manavortex: The compiler was complaining that the assigned value wasn't considered in any paths. I can put it back?

rfuzzo (commented):
Same here, the breaking out of the methods into the mesh functions is a great idea, but the unrelated logic changes are harder to review and will take time.

PR #2171, opened 2025-01-13
un-dynamify mesh materials

manavortex:
# un-dynamify mesh materials

**Implemented:**
- can convert dynamic mesh materials into regular materials

before:
image
image

after:
image


rfuzzo (approved):
seems to work and is a standalone feature.

Issue #2179, opened 2025-01-14
Fix compiler warnings in build job


manavortex:
To improve the code base's quality, we should get a grip on those pesky warnings with every build 

⚠️ Attention:
> If you want to set a default value, do it via (Post)Constructor 
> ~ @seberoth ([discord](https://discord.com/channels/717692382849663036/717720094196760760/1315237977886162974))

To clarify, this is about the long list of these:
```log
Warning: D:\a\WolvenKit\WolvenKit\WolvenKit.RED4\Types\Primitives\CArrayBase.cs(167,16): warning CS8767: Nullability of reference types in type of parameter 'item' of 'int CArrayBase<T>.Add(object item)' doesn't match implicitly implemented member 'int IList.Add(object? value)' (possibly because of nullability attributes). [D:\a\WolvenKit\WolvenKit\WolvenKit.RED4\WolvenKit.RED4.csproj]
Warning: D:\a\WolvenKit\WolvenKit\WolvenKit.RED4\Types\Primitives\CArrayBase.cs(225,17): warning CS8767: Nullability of reference types in type of parameter 'value' of 'bool CArrayBase<T>.Contains(object value)' doesn't match implicitly implemented member 'bool IList.Contains(object? value)' (possibly because of nullability attributes). [D:\a\WolvenKit\WolvenKit\WolvenKit.RED4\WolvenKit.RED4.csproj]
```

PR #2180, opened 2025-01-14
refactor: fix compiler warnings (part 1)

manavortex:
# refactor: fix compiler warnings (part 1)

- Applied a bunch of code style recommendations (per-folder, in their own commits)
- Fixed a bunch of compiler nullability warnings

I recommend reviewing by-commit. Only the last commit contains manual changes :)

Part 1 for #2179

### DIFF
--- a/WolvenKit.RED4/Archive/IO/CR2WReader.cs
+++ b/WolvenKit.RED4/Archive/IO/CR2WReader.cs
@@ -100,14 +100,14 @@ public partial class CR2WReader : Red4Reader
 
             var propRedType = RedReflection.GetRedTypeFromCSType(prop.Type, prop.Flags);
 
-            if (redTypeName != CName.Empty && propRedType != redTypeName)
+            if (redTypeName.GetResolvedText() is string s && propRedType != redTypeName)
             {
-                if (TryReadValue(redTypeName!, size, out value))
+                if (TryReadValue(s, size, out value))
                 {
-                    var args = new InvalidRTTIEventArgs(fullName, propRedType, redTypeName!, value);
+                    var args = new InvalidRTTIEventArgs(fullName, propRedType, s, value);
                     if (!HandleParsingError(args))
                     {
-                        throw new InvalidRTTIException(fullName, propRedType, redTypeName!);
+                        throw new InvalidRTTIException(fullName, propRedType, s);
                     }
 
                     SetProperty(cls, prop.RedName, args.Value);
--- a/WolvenKit.RED4/Archive/IO/CR2WReader.cs
+++ b/WolvenKit.RED4/Archive/IO/CR2WReader.cs
@@ -100,14 +100,14 @@ public partial class CR2WReader : Red4Reader
 
             var propRedType = RedReflection.GetRedTypeFromCSType(prop.Type, prop.Flags);
 
-            if (propRedType != redTypeName)
+            if (redTypeName != CName.Empty && propRedType != redTypeName)
             {
                 if (TryReadValue(redTypeName!, size, out value))
                 {
-                    var args = new InvalidRTTIEventArgs(fullName, propRedType, redTypeName, value);
+                    var args = new InvalidRTTIEventArgs(fullName, propRedType, redTypeName!, value);
                     if (!HandleParsingError(args))
                     {
-                        throw new InvalidRTTIException(fullName, propRedType, redTypeName);
+                        throw new InvalidRTTIException(fullName, propRedType, redTypeName!);
                     }
 
                     SetProperty(cls, prop.RedName, args.Value);
--- a/WolvenKit.RED4/Archive/IO/RedPackageReader.cs
+++ b/WolvenKit.RED4/Archive/IO/RedPackageReader.cs
@@ -3,6 +3,7 @@ using System.Text;
 using WolvenKit.Core.Extensions;
 using WolvenKit.Core.Interfaces;
 using WolvenKit.RED4.Archive.Buffer;
+using WolvenKit.RED4.Archive.CR2W;
 using WolvenKit.RED4.IO;
 using WolvenKit.RED4.Types;
 using WolvenKit.RED4.Types.Exceptions;
@@ -53,9 +54,9 @@ public partial class RedPackageReader : Red4Reader
 
                 var propRedType = RedReflection.GetRedTypeFromCSType(prop.Type, prop.Flags);
 
-                if (propRedType != redTypeName)
+                if (redTypeName != CName.Empty && propRedType != redTypeName)
                 {
-                    if (TryReadValue(varName, redTypeName, compiledPropertyData, out value))
+                    if (TryReadValue(varName, redTypeName!, compiledPropertyData, out value))
                     {
                         var args = new InvalidRTTIEventArgs(fullName, propRedType, redTypeName!, value);
                         if (!HandleParsingError(args))
--- a/WolvenKit.RED4/Archive/IO/RedPackageReader.cs
+++ b/WolvenKit.RED4/Archive/IO/RedPackageReader.cs
@@ -54,14 +54,14 @@ public partial class RedPackageReader : Red4Reader
 
                 var propRedType = RedReflection.GetRedTypeFromCSType(prop.Type, prop.Flags);
 
-                if (redTypeName != CName.Empty && propRedType != redTypeName)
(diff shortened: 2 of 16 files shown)


rfuzzo (changes requested):
most of these are not warnings, but suggestions. I don't think we need these

manavortex:
the first bunch of commits, yes. I remember that we wanted to stick to code conventions more? Also, I already changed it, so un-changing it is more work at this point. 
I can do that if you want me to.

rfuzzo:
I wouldn't want to merge the suggestions, no. We could look at a PR that only addresses warnings in a separate one, but I don't see it as high priority rn.

rfuzzo:
thanks for leaving out the suggestions. There's still a bunch of logic changes that I'd need to review, and I'm not really sure we need them if nothing is broken. Seems a bit risky

rfuzzo:
Will take time to go over everything

WolvenKit.RED4/Archive/IO/CR2WReader.cs
```diff
@@ -100,14 +100,14 @@ public bool ReadVariable(RedBaseClass cls, ExtendedTypeInfo typeInfo)
 
             var propRedType = RedReflection.GetRedTypeFromCSType(prop.Type, prop.Flags);
 
-            if (propRedType != redTypeName)
+            if (redTypeName != CName.Empty && propRedType != redTypeName)
```
rfuzzo, line 103: This changes serialization logic (the string check), why?

WolvenKit.RED4/Archive/IO/RedPackageReader.cs
```diff
@@ -53,9 +54,9 @@ public override void ReadClass(RedBaseClass cls, uint size)
 
                 var propRedType = RedReflection.GetRedTypeFromCSType(prop.Type, prop.Flags);
 
-                if (propRedType != redTypeName)
+                if (redTypeName != CName.Empty && propRedType != redTypeName)
```
rfuzzo, line 57: Same here, why the additional check?

WolvenKit.RED4/Types/ClassesExt/Appendix/CMaterialInstance.cs
```diff
@@ -8,13 +8,15 @@ public partial class CMaterialInstance : IRedAppendix
...
+        get => GetPropertyValue<CArray<CKeyValuePair>>()!; // set in PostConstruct, so not nullable
         set => SetPropertyValue<CArray<CKeyValuePair>>(value);
+
     }
 
     partial void PostConstruct()
     {
-        Values = new CArray<CKeyValuePair>();
+        Values = [];
+        Metadata ??= new SerializationDeferredDataBuffer();
```
rfuzzo, line 19: Why this change?

WolvenKit.RED4/Types/ClassesExt/appearanceAppearanceDefinition.cs
```diff
@@ -12,7 +12,9 @@ public partial class appearanceAppearanceDefinition
     [REDProperty(IsIgnored = true)]
     public CArray<entIComponent> Components
     {
-        get => GetPropertyValue<CArray<entIComponent>>();
-        set => SetPropertyValue<CArray<entIComponent>>(value);
+        get => GetPropertyValue<CArray<entIComponent>>()!; // set in PostConstruct, so not nullable
+        set => SetPropertyValue(value);
     }
+
+    partial void PostConstruct() => Components ??= [];
```
rfuzzo, line 19: Why this change? It could have unforeseen consequences

WolvenKit.RED4/Types/ClassesExt/entEntityTemplate.cs
```diff
@@ -7,15 +7,22 @@ public partial class entEntityTemplate
...
     [REDProperty(IsIgnored = true)]
     public CArray<entIComponent> Components
     {
-        get => GetPropertyValue<CArray<entIComponent>>();
+        get => GetPropertyValue<CArray<entIComponent>>()!; // set in PostConstruct, so not nullable
         set => SetPropertyValue<CArray<entIComponent>>(value);
     }
+
+
+    partial void PostConstruct()
```
rfuzzo, line 23: Same here, changes serialization logic

rfuzzo (changes requested):

WolvenKit.RED4/Types/ClassesExt/Appendix/CMaterialInstance.cs
manavortex: Type checking. The property is not nullable, but it's not necessary initialized, and I actually ran into an exception about this.

WolvenKit.RED4/Types/ClassesExt/appearanceAppearanceDefinition.cs
manavortex: Same as above - it's not annotated as nullable, but the object can leave the constructor with the property not having been set.

WolvenKit.RED4/Types/ClassesExt/entEntityTemplate.cs
manavortex: but enforces type safety. Entity shouldn't be null, nor should components - not even if serializing. Pretty sure the serializer assigns empty objects there if they are?

WolvenKit.RED4/Archive/IO/CR2WReader.cs
manavortex: because we can't set a property from an empty string, it'll throw exceptions.

WolvenKit.RED4/Archive/IO/RedPackageReader.cs
manavortex: Same reason, empty string is not a valid key here

WolvenKit.RED4/Archive/IO/CR2WReader.cs
seberoth: `redTypeName != CName.Empty` isn't a nullable check.
`CName` is a value type and can't be null (also CName.Empty isn't null either (it's 0), so doesn't make sense here). 
Whats giving the warning is the exception constructor (or `TryReadValue`) which wants a string. It tries to convert the CName to string via implicit conversion which can yield null (unresolveable string).

Adding the `!` below (as you did) should be enough, since impossible that the string can't be resolved at this point.

Same for the other below

manavortex: Thanks, Seb <3

WolvenKit.RED4/Types/ClassesExt/Appendix/CMaterialInstance.cs
seberoth: General Info (not specialy for `CMaterialInstance`)

Changing the "default" value from `null` to `new()` indeed changes how the CR2W is written.
It compares the set value to the default one and if it is default, the property doesn't get written at all.

So the logic changes from:
If it's null, it doesn't get written
If it's a buffer (even a empty one) it gets written.

To:
If it's null, it would try to write that (don't know what would happen then currently)
If it's a empty buffer, it won't get written
If it's a filled buffer, it gets written.

So in theory such things can break in engine. 

If these warning annoys you, should probably disable them for `ClassesExt` via `.editorconfig` like in `Classes`

manavortex: OK, I get that it has side effects. My dilemma here is that I kind of want these properties to be initialized inside of WKit, because without this change I am constantly initializing them by hand (every time after creating a new material e.g.)

Should I define them in the initialization again, or what would be the engine-proof solution here?

seberoth: Easiest (without changing the logic) would probably be creating a factory (helper) for them (i.e in .Common / .ModKit)?
Which yields you the classes as you need them the most?

rfuzzo: For such cases like creating new materials you do indeed need to create new objects yeah, because null here is a valid value for the cr2w

manavortex: OK, I'll see that I try and create a factory here then

rfuzzo (approved):

PR #2214, opened 2025-01-19
Better dialog support

seberoth:
# Better dialog support

**Implemented:**
- Better dialog support

**Fixed:**
- Changed `IViewFor<AppViewModel>` to a static service. Solves issue where the service locator would create new instances of `MainView` though we only have one (the main Window)

**Additional notes:**
Propose to merge it as is and migrate the current logic step by step so it can be used for new dialogs/modals directly.
Added migration of the `SoundModdingViewModel` modal as an example.

Not sure yet, how we should handle MessageBoxes in the future though...

### DIFF
--- a/WolvenKit.App/Services/IDialogService.cs
+++ b/WolvenKit.App/Services/IDialogService.cs
@@ -1,0 +1,56 @@
+﻿using System.Threading.Tasks;
+using System;
+using WolvenKit.App.ViewModels.Dialogs;
+
+namespace WolvenKit.App.Services;
+
+/// <summary>
+/// Provides methods to display dialogs and modals in the application.
+/// </summary>
+public interface IDialogService
+{
+    /// <summary>
+    /// Displays a dialog for the specified view model.
+    /// </summary>
+    /// <param name="viewModel">The view model for the dialog.</param>
+    /// <returns>A nullable boolean indicating the result of the dialog.</returns>
+    public bool? ShowDialog(DialogExtViewModel viewModel);
+
+    /// <summary>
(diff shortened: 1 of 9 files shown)


manavortex (approved):

poirierlouis:
It will be great!
I haven't looked at everything (phone). It looks like only a callback is used to react when dialog is closed. Should we provide two callbacks instead, onNext and onCancel for example?onCancel could be optional and nullable.
I also wonder if it allows to send arbitrary data back to the caller?

Regarding MessageBox, I think these are good enough, no states should be stored so singleton/service pattern would be useless imo. Though it could be at least renamed to be more explicit. Iirc currently it is Interactions. It could be PopupHelper/MessageBoxHelper for example.

manavortex:
That's a great point for dialogues using the wizard. OnNext and OnBack, to be precise.

seberoth:
In my opinion, that should be handled in the callbacks itself.
The ViewModel gets passed into the callback.

Also would get hard to implement for dialogs with more states. `onNext`, `onOk`, `onCancel`, ...

seberoth:
Easier to just define a state var in the vm if needed

poirierlouis:
Hum. Only two could be enough, right path and wrong path.
I'm thinking about Curve Editor and imo Dialog should read/write with immutable instance, answer with a result when required and then caller update it states. Iirc CurveEditor directly get an instance of CVM, mess with it and then "Save". But afaik it only closes the dialog, as changes were already applied during user interactions in the dialog (chart).
This seems wrong regarding UX and SoC.
Hence this idea so when we refactor other dialogs, will be able to implement a stronger/resilient solution. wdyt?

PR #2215, opened 2025-01-19
Delete partsvalues and overrides

manavortex:
# .app file: delete partsValues and partsOverrides

... as discussed with psi, these aren't needed for CCXL addition apps, so we should be able to ditch them all at once

image


rfuzzo (approved):
why not

PR #2225, opened 2025-01-25
View: exposing save/delete default layout

manavortex:
# View: Default layout
 
- You can now save or delete the default layout from the view menu (previously, you could only save default layout by saving the project layout without an open project)
- The project layout operations are disabled while no project is opened

image
image


rfuzzo (approved):
very cool. Tested and works

PR #2226, opened 2025-01-25
rename/refactor: will now also work in inkcc files

manavortex:
# Rename/Refactor in .inkcharcustomization

sort of important for ccxl additions :)


rfuzzo (approved):
✅

PR #2227, opened 2025-01-25
fix exception when exporting materials

manavortex:
# fix exception when exporting materials

Materials with both local and external materials can be exported again now


rfuzzo:
should be fixed with #2229

PR #2229, opened 2025-01-26
Implemented IComparable for Fundamentals

seberoth:
# Implemented IComparable for Fundamentals

**Implemented:**
- Implemented IComparable for Fundamentals

### DIFF
--- a/WolvenKit.RED4/Types/Primitives/Fundamentals/CInt16.cs
+++ b/WolvenKit.RED4/Types/Primitives/Fundamentals/CInt16.cs
@@ -5,7 +5,7 @@ namespace WolvenKit.RED4.Types;
 
 [RED("Int16")]
 [DebuggerDisplay("{_value,nq}", Type = "CInt16")]
-public readonly struct CInt16 : IRedPrimitive<short>, IEquatable<CInt16>, IRedInteger
+public readonly struct CInt16 : IRedPrimitive<short>, IComparable, IComparable<CInt16>, IEquatable<CInt16>, IRedInteger
 {
     [DebuggerBrowsable(DebuggerBrowsableState.Never)]
     private readonly short _value;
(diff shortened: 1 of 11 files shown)


manavortex (approved):

WolvenKit.RED4/Types/Primitives/Fundamentals/CInt16.cs
```diff
@@ -45,4 +45,22 @@ public override bool Equals(object? obj)
...
+        {
+            return 1;
+        }
+        if (value is CInt16 i)
+        {
+            return CompareTo(i);
+        }
+        throw new ArgumentException("Value is not a CInt16", nameof(value));
+    }
+    public int CompareTo(CInt16 value) => _value - value._value;
```
rfuzzo, line 63: why does this work here and not for CInt32?

seberoth: Cannot use return (_value - value) as this causes a wrap around in cases where _value - value > MaxValue.

rfuzzo: right, btw would it make sense to just cast and deffer the implementation to the net types? or is there a problem with that? or is it not elegant?

rfuzzo (approved):

WolvenKit.RED4/Types/Primitives/Fundamentals/CInt16.cs
seberoth: Yeah, could do that to. (Except for CBool)
Asked myself which would be better.
Just decided for reimplementing them to save one function call. No other reason.
Can change if you want.

rfuzzo: nope, was just curious

PR #2232, opened 2025-01-26
fix: multiple selected properties

manavortex:
# fix: multiple selected properties

I don't understand it, but this works. If anyone wants to improve this PR, be my guest, I completely lost track regarding that fucking property sync

### DIFF
--- a/WolvenKit/Views/Documents/RDTDataView.xaml.cs
+++ b/WolvenKit/Views/Documents/RDTDataView.xaml.cs
@@ -20,20 +20,6 @@ namespace WolvenKit.Views.Documents
         public RDTDataView()
         {
             InitializeComponent();
-
-            //this.WhenAnyValue(x => x.DataContext).Subscribe(x =>
-            //{
-            //    if (x is RDTDataViewModel vm)
-            //    {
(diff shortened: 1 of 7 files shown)


WolvenKit/Views/Documents/RDTDataView.xaml.cs
```diff
@@ -42,87 +31,26 @@ public RDTDataView()
...
+        {
+            if (sender is not RedTreeView tree || e.PropertyName is not nameof(RedTreeView.SelectedItems) ||
+                ViewModel is null)
+            {
+                return;
+            }
 
+            ViewModel.SelectedChunks ??= [];
+            ViewModel.SelectedChunks.Clear();
+            ViewModel.SelectedChunks.AddRange(tree.SelectedItems.ToList());
```
manavortex, line 52: This is complete and utter bullshit. I have to shove them up the list in this event because the property doesn't update – but if I use SetProperty, then the two-way binding fires and nulls the tree's selection??? What the actual fuck! @rfuzzo help

rfuzzo:
fixed in #2249

PR #2237, opened 2025-01-27
fix: allow pasting of single selection into properties

manavortex:
# fix: allow pasting of single selection into properties


manavortex:
For some reason, CheckTypeCompatibility isn't yet returning true :|

rfuzzo:
this pulls in too many unrelated commits

PR #2241, opened 2025-01-30
Added basic support to view quest/scene graphes with custom/dynamic nodes

seberoth:
# Added basic support to view quest/scene graphes with custom/dynamic nodes

**Implemented:**
- Added basic support to view quest/scene graphes with custom/dynamic nodes

### DIFF
--- a/WolvenKit.App/ViewModels/GraphEditor/RedGraph.Quest.cs
+++ b/WolvenKit.App/ViewModels/GraphEditor/RedGraph.Quest.cs
@@ -211,6 +211,20 @@ public partial class RedGraph
         return nodeWrapper;
     }
 
+    private DynamicQuestViewModel WrapDynamicQuestNode(DynamicBaseClass node, bool isNew)
+    {
+        var nodeWrapper = new DynamicQuestViewModel(node);
+
+        if (isNew)
+        {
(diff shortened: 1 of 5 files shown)


WolvenKit.App/ViewModels/GraphEditor/RedGraph.Quest.cs
```diff
@@ -458,18 +472,23 @@ public static RedGraph GenerateQuestGraph(string title, graphGraphDefinition que
...
         foreach (var nodeHandle in questGraph.Nodes)
         {
-            ArgumentNullException.ThrowIfNull(nodeHandle.Chunk);
-
-            var node = nodeHandle.Chunk;
-            var nodeID = 0;
+            var node = nodeHandle.GetValue();
+            if (node is null)
+            {
+                throw new Exception("Empty nodes aren't supported!");
```
manavortex, line 478: Maybe you want to catch this exception somewhere further up, and only print the message? Because exceptions are scary to users

seberoth: We already do that?
I just added/changed the text so its more clear what is wrong

rfuzzo (approved):
Lgtm!

PR #2251, opened 2025-02-02
WScript improvements

### LINKED ISSUE #2174: OnSave hook: Allow enabling/disabling per file type
Currently, I can enable/disable the onSave hook in the script manager:
image

This will fully enable/disable the feature. 

- allow enabling/disabling it on a per-file basis
- move the setting into the C# layer into the SettingsManager (as an extra group)

seberoth:
# WScript improvements

**Implemented:**
- Added `// @hook_type` (see below)
- Added `// @hook_priority` (see below)
- Changed `// @hook_extension` (see below)
- Allow multiple onSave hooks

**Allowed/Added/Changed WScript metadata:**
 - `@hook_type`: `None,Save,Export,PreImport,ImportJson,ParseError` (case-insensitive, no blanks allowed)
 - `@hook_extension`: RedEngine file formats for `Save,Export,ImportJson` or import formats for `PreImport` (case-insensitive, no blanks allowed)
 - `@hook_priority`: if multiple hook script affect the same action. Higher priority executes first. (default prio is 100)

**Additional notes:**
Closes #2174

### DIFF
--- a/WolvenKit.App/Services/AppScriptService.Hook.cs
+++ b/WolvenKit.App/Services/AppScriptService.Hook.cs
@@ -32,6 +32,13 @@ public partial class AppScriptService
         }
     }
 
+    private IList<ScriptFile> GetHookScripts(HookType hookType = HookType.Global, string? ext = null) => 
+        GetScripts()
+            .Where(scriptFile => scriptFile.Type == ScriptType.Hook)
+            .Where(scriptFile => scriptFile.HookType.HasFlag(hookType))
+            .Where(scriptFile => scriptFile.IsHookExtensionSupported(ext))
+            .ToList();
+
     private bool OnSaveHook(string filePath, ref CR2WFile cr2wFile)
     {
         if (string.IsNullOrEmpty(filePath))
@@ -48,18 +55,8 @@ public partial class AppScriptService
         _settingsManager.ScriptStatus ??= new();
 
         ScriptFile? script = null;
-        foreach (var scriptFile in GetScripts())
+        foreach (var scriptFile in GetHookScripts(HookType.Save, extStr))
         {
-            if (scriptFile.Type != ScriptType.Hook)
-            {
-                continue;
-            }
-
-            if (scriptFile.HookExtension != "global"/* && scriptFile.HookExtension != extStr*/)
-            {
-                continue;
-            }
-
             if (_settingsManager.ScriptStatus.TryGetValue(scriptFile.Path, out var enabled) && !enabled)
             {
                 continue;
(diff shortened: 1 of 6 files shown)


WolvenKit.App/Services/AppScriptService.Hook.cs
```diff
@@ -47,57 +55,41 @@ private bool OnSaveHook(string filePath, ref CR2WFile cr2wFile)
...
         _settingsManager.ScriptStatus ??= new();
 
-        ScriptFile? script = null;
-        foreach (var scriptFile in GetScripts())
+        string? jsonContent = null;
+
+        foreach (var scriptFile in GetHookScripts(HookType.Save, extStr))
         {
-            if (scriptFile.Type != ScriptType.Hook)
+            if (_settingsManager.ScriptStatus.TryGetValue(scriptFile.Path, out var enabled) && !enabled)
```
manavortex, line 62:
```suggestion
            if (!_settingsManager.ScriptStatus.TryGetValue(scriptFile.Path, out var enabled) || enabled != true)
```

manavortex: Right now, it will not continue if TryGetValue fails

manavortex (approved):

WolvenKit.App/Services/AppScriptService.Hook.cs
seberoth: Thats true, but it was always like that.
Do you want it to be changed?

manavortex: Would make sense, yeah :)

seberoth: Then we should do it in a separate PR.
Would need some refactor, since it only gets added there if you toggle the checkbox in the scriptmanager.
Every hook would be ignored even if it appears checked in the manager.

seberoth: Wdym with "invalid headers"?
The dict stores the state of the scripts (Enabled/Disabled).
Currently it only gets added there, if the user switches it from the default state Enabled to Disabled.

manavortex: Oh - yeah, okay, that makes sense >.<  But if the headers can't be read, then the script shouldn't be enabled, should it?

manavortex: Anyhow, this was just a suggestion :) otherwise LGTM

seberoth: Ah, the wscript header.
If its invalid, `GetScripts()` / `GetHookScripts()` doesn't return.

PR #2265, opened 2025-02-09
Added settings to wscript, attempt no. 3

seberoth:
# Added settings to wscript, attempt no. 3

**Implemented:**
- WScript settings

**Notes:**
Currently supports the following types: `int`, `double`, `string`, `bool `, `BigInteger`+ array `[]` variants

**Example:**
```js
import * as Logger from 'Logger.wscript';

var config = {
	"myIntVar": 5,
	"myIntArrayVar": [6,7,8],
	"myDoubleVar": 3.141,
	"myDoubleArrayVar": [1.0,2.5,5.9],
	"myStringVar": "test",
	"myStringArrayVar": ["abc","def","123"],
	"myBoolVar": false,
	"myBoolArrayVar": [true,false,true,true],
	"myBigIntVar": BigInt(5),
	"myBigIntArrayVar": [BigInt(5), BigInt(4), BigInt(3)]
}

wkit.ShowSettings(config);

Logger.Info(config)
```

### DIFF
--- a/WolvenKit.App/Interaction/Interactions.cs
+++ b/WolvenKit.App/Interaction/Interactions.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using WolvenKit.App.Helpers;
 using WolvenKit.App.Models.ProjectManagement.Project;
+using WolvenKit.App.Scripting;
 using WolvenKit.App.Services;
 using WolvenKit.App.ViewModels.Dialogs;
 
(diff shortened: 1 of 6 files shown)


notaspirit:
Tested the artifact and with your example bigInt fails to be serialized when logging, otherwise works well. UI / UX is a bit rough but I assume you are aware and it's something that would make sense to optimize based on user feedback. 
The entire config being one object is really nice for the scripting side.

seberoth:
1. Yeah, haven't pushed the wscripts for BigInt logging yet. Do that soon.
2. Me not an UX guy. That's the best I can provide :P

seberoth:
Updating wscripts (or restarting wkit) should fix 1.

justarandomguyintheinternet:
Is this supposed to be persistent? (The name made me think so, but does not seem to be the case)

notaspirit:
I don't think it is supposed to or needs be persistent, if you need persistence you can save the values to a json in raw

justarandomguyintheinternet:
True, my brain just thinks settings -> persistent, otherwise i'd call it just user input or something

seberoth:
persistent storage for wscript should be easy.
Will add it to my todo list. Not for this PR

PR #2271, opened 2025-02-11
fix: hair ccxl button hides for .mesh files with no appearances

Silvakabe:
# Fixed hair CCXL conversion button logic

**Fixed:**
- Button no longer hides for .mesh files with no appearances, which is a case that is covered in the main logic.


WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
```diff
@@ -311,6 +311,156 @@ private void OnUnDynamifyMaterialsClick(object _, RoutedEventArgs e)
...
+            var mainHairMiType = dialog.ViewModel?.SelectedMiType;
+            var mainMiPath = (CName)(mainHairMiFile!);
+
+            if (mainHairMiType is "Cap01")
+            {
+                mainHairMiType = "Cap";
+            }
+
+
+            cvm.GetPropertyChild("appearances")?.CalculateProperties();
```
manavortex, line 354: Just as a feedback: instead of doing all this in the view and constantly (re)calculating all properties, you could also fix it directly in the CR2W and just refresh the result. Won't be possible to undo that, but it's an option :D

manavortex (approved):

WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
Emiukabe: highly appreciate your feedback! I'm not sure what you meant by fixing it directly in the CR2W but I'll try to look around for better implementations!

Ngl it did feel a little excessive in calc/recalc the affected properties, but i wanted to be in line with the other methods in the file😄

manavortex: Check TemplateFileHelper (I think I called it like that)?

manavortex (approved):

PR #2292, opened 2025-03-02
add material dependencies: will no longer abort after first material

manavortex:
# add material dependencies: will no longer abort after first material

**Fixed:**
- material depencendies replacement will no longer abort after 1st material

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
+++ b/WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
@@ -32,17 +32,19 @@ public partial class ChunkViewModel
     /// <param name="replace"></param>
     /// <param name="isWholeWord"></param>
     /// <param name="isRegex"></param>
+    /// <param name="ignoreCache"></param>
     /// <returns></returns>
-    private int SearchAndReplaceInternal(string search, string replace, bool isWholeWord, bool isRegex)
+    private int SearchAndReplaceInternal(string search, string replace, bool isWholeWord, bool isRegex,
+        bool ignoreCache = false)
     {
         NumReplacedEntries = 0;
         var isExpanded = IsExpanded;
 
         // SearchAndReplaceInProperties can influence the expansion state, even if nothing was changed
-        var hasReplacement = SearchAndReplaceInProperties(search, replace, isWholeWord, isRegex);
+        var hasReplacement = SearchAndReplaceInProperties(search, replace, isWholeWord, isRegex, ignoreCache);
         IsExpanded = isExpanded;
 
-        if (hasReplacement)
+        if (!hasReplacement)
         {
             return NumReplacedEntries;
         }
@@ -55,11 +57,12 @@ public partial class ChunkViewModel
 
 
     // Level 1 (will call itself recursively, so let's abort here if we can)
-    private bool SearchAndReplaceInProperties(string search, string replace, bool isWholeWord, bool isRegex)
+    private bool SearchAndReplaceInProperties(string search, string replace, bool isWholeWord, bool isRegex,
+        bool ignoreCache = false)
     {
         var properties = Properties.ToList();
 
-        if (s_resolvedHashes.Contains(GetHashCode()))
+        if (!ignoreCache && s_resolvedHashes.Contains(GetHashCode()))
         {
             return false;
         }
--- a/WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
+++ b/WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
@@ -1,7 +1,9 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 using WolvenKit.App.Helpers;
 using WolvenKit.Core.Extensions;
 using WolvenKit.RED4.Types;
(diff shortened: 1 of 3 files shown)


WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
```diff
@@ -32,17 +32,19 @@ public static void SearchAndReplace_ResetCaches()
...
 
         // SearchAndReplaceInProperties can influence the expansion state, even if nothing was changed
-        var hasReplacement = SearchAndReplaceInProperties(search, replace, isWholeWord, isRegex);
+        var hasReplacement = SearchAndReplaceInProperties(search, replace, isWholeWord, isRegex, ignoreCache);
         IsExpanded = isExpanded;
 
-        if (hasReplacement)
+        if (!hasReplacement)
         {
             return NumReplacedEntries;
```
rfuzzo, line 49: Where is this variable set?

WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
```diff
@@ -32,17 +32,19 @@ public static void SearchAndReplace_ResetCaches()
...
+    private int SearchAndReplaceInternal(string search, string replace, bool isWholeWord, bool isRegex,
+        bool ignoreCache = false)
     {
         NumReplacedEntries = 0;
         var isExpanded = IsExpanded;
 
         // SearchAndReplaceInProperties can influence the expansion state, even if nothing was changed
-        var hasReplacement = SearchAndReplaceInProperties(search, replace, isWholeWord, isRegex);
+        var hasReplacement = SearchAndReplaceInProperties(search, replace, isWholeWord, isRegex, ignoreCache);
         IsExpanded = isExpanded;
```
rfuzzo, line 45: Why set it back here and also before? It's not clear where IsExpanded is set

WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
```diff
@@ -55,11 +57,12 @@ private int SearchAndReplaceInternal(string search, string replace, bool isWhole
...
 
     // Level 1 (will call itself recursively, so let's abort here if we can)
-    private bool SearchAndReplaceInProperties(string search, string replace, bool isWholeWord, bool isRegex)
+    private bool SearchAndReplaceInProperties(string search, string replace, bool isWholeWord, bool isRegex,
+        bool ignoreCache = false)
     {
         var properties = Properties.ToList();
 
-        if (s_resolvedHashes.Contains(GetHashCode()))
+        if (!ignoreCache && s_resolvedHashes.Contains(GetHashCode()))
```
rfuzzo, line 65: What's the reason for this? Why ignore the hash cache?

manavortex: because if the operation is called async, then checking the cache will break things. It only works for the sequential case.

WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
manavortex: isExpanded will store the node's expansion state to restore it after potential replacements (this will always expand the node).

WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
rfuzzo: yeah

WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
rfuzzo: Yeah I know, but it's a bit of a bad design. the member variable just gets set in seemingly random places. We should remove it from the CVM tbh, and make it local

WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.SearchAndReplace.cs
rfuzzo: can just make it a ConcurrentBag

manavortex: I've never used that - can somebody else implement that part of it? @notaspirit maybe?

notaspirit: from what I can see order of the entries doesn't matter so you would just change the `List<int>` and `List<string>` in line 13 and 15 to use `ConcurrentBag<T>` instead

notaspirit: Using `ConcurrentBag<T>` won't change where the cache is, it would be the same as it is currently with the `List<T>` implementation (which is 1 for all instances since it is static). If you wanted to make it per thread one option would be to make it part of the instance and then making sure each thread gets their own instance of the class. 

But why would it need to have a cache per thread if it currently just has one cache for everything?

manavortex: Because I never anticipated that I would have to call this from threads. (Not doing it is not an option, though, don't want to wait two minutes for multiple search&replace options)

rfuzzo: This isn't even thraded. It's just a "fake" async task that is called on a background worker and not block the UI.

rfuzzo: But yeah, ConcurrentBag will do it

PR #2294, opened 2025-03-02
hopefully fixed 'paste handle'

### LINKED ISSUE #2289: ``Paste Handle`` function not exists for .inkwidget anymore
After Wolvenkit 8.15.1-nightly.2024-10-27 function` Paste Handle` in inkwdiget not exist in Context Menu, so work/edit on the .inkwidget cannot be completed because parent entries cannot be inserted

Reproducing an issue:
1. Go to (widget example): base\gameplay\gui\widgets\vehicle\porsche\911_turbo_ui.inkwidget
2. Open `libraryItems ` of hierarchy until you see `gameController `and `rootWidget`:
3. `rootWidget `-> `children `-> `children` -> `speedometer_mph` -> `parentWidget` -> (Context Menu) `ResetObject`
Image
4. Back to `rootWidget ` (Root) -> select `Copy Handle`
Image

5. Go to path 3. again -> select `ParentWidget` -> (Context Menu) Still see `Reset Object`, but there is no` Paste Handle` anymore
Image

**Expected behavior**
In addition to the `Reset Object `button, there should also be a` Paste Handle` button that will supply the copied array of components of the Root hierarchy or any other hierarchy

Image

Image

WITHOUT issue -- Wolvenkit 8.15.1-nightly.2024-10-27
WITH issue -- Wolvenkit 8.16.1-nightly.2025-02-19

manavortex:
# hopefully fixed 'paste handle'

closes #2289

### DIFF
--- a/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
@@ -2438,13 +2438,13 @@ public partial class ChunkViewModel : ObservableObject, ISelectableTreeViewItemM
         catch (Exception ex) { _loggerService.Error(ex); }
     }
 
-    public IRedType? CopyData()
+    public IRedType? CopyData(bool copyAsHandle = false)
     {
         try
         {
             return Data switch
             {
-                IRedBaseHandle baseHandle => baseHandle,
+                IRedBaseHandle baseHandle when copyAsHandle => baseHandle,
                 IRedCloneable irc => (IRedType)irc.DeepCopy(),
                 _ => Data
             };
--- a/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
@@ -2438,6 +2438,10 @@ public partial class ChunkViewModel : ObservableObject, ISelectableTreeViewItemM
         catch (Exception ex) { _loggerService.Error(ex); }
     }
 
+    /// <summary>
+    /// Handles data copy for pasting into another node.
+    /// </summary>
+    /// <param name="copyAsHandle">Copy reference instead of data?</param>
     public IRedType? CopyData(bool copyAsHandle = false)
     {
         try
--- a/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
@@ -2441,7 +2441,12 @@ public partial class ChunkViewModel : ObservableObject, ISelectableTreeViewItemM
     {
         try
         {
-            return Data is IRedCloneable irc ? (IRedType)irc.DeepCopy() : Data;
+            return Data switch
+            {
+                IRedBaseHandle baseHandle => baseHandle,
+                IRedCloneable irc => (IRedType)irc.DeepCopy(),
+                _ => Data
+            };
         }
         catch (Exception ex)
         {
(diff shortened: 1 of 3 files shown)


WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -2366,7 +2366,8 @@ private static bool IsHandle(IRedType? irb, bool weakHandleOnly)
         }
 
         var gt = propertyType.GetGenericTypeDefinition();
-        return (weakHandleOnly && gt == typeof(CWeakHandle<>)) || gt == typeof(CHandle<>);
+        return (weakHandleOnly && gt == typeof(CWeakHandle<>)) || gt == typeof(CWeakHandle<>) ||
```
rfuzzo, line 2369: This is missing a !weakHandleOnly in the right side of the OR

rfuzzo (changes requested):

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -3457,7 +3468,7 @@ private TypeCompability CheckTypeCompatibility(Type? destType, Type? srcType)
         }
 
         if (srcType.IsAssignableTo(typeof(IRedBaseHandle)) &&
-            srcType.GetGenericArguments()[0].IsAssignableTo(destType))
```
rfuzzo, line 3460: It should not be changed. Type.IsAssignableTo is true if "The current type is a generic type parameter, and targetType represents one of the constraints of the current type."

rfuzzo (changes requested):

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
manavortex: If it's not weakhandleonly, it doesn't matter :)

rfuzzo: but it's an OR. The whole conditional will be true if the type is a weakhandle even if weakHandleOnly is true.

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
manavortex: Is that not exactly the wrong way around? srcType is generic, and destType is not generic.

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -2453,7 +2455,12 @@ private void CopyChunk()
     {
         try
         {
-            return Data is IRedCloneable irc ? (IRedType)irc.DeepCopy() : Data;
+            return Data switch
+            {
+                IRedBaseHandle baseHandle => baseHandle,
```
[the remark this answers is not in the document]
manavortex: Right, yeah - because if I copy handle, then I don't want anything cloned and just the original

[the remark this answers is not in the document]
rfuzzo: Yeah but this is already handled by the default clause, if the type is not a IRedConeable then it just returns Data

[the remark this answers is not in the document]
manavortex: but the non-weak handle I tried it with was cloneable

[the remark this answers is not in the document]
rfuzzo: Hmm okay

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
manavortex: @rfuzzo?

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
rfuzzo, line 2460: Here, inside the function called “CopyChunk” we now don’t deepcopy the handle anymore but return a reference to it - even though it is an ICloneable, why?

rfuzzo:
As discussed, this should be the correct design: 
```
handle -> handle => deepcopy
handle -> whandle => ref
whandle -> whandle => ref

handle:something -> something => deepcopy
something -> handle:something => deepcopy
whandle:something -> something => deepcopy
something -> whandle:something = not allowed

(with type checks of course)
```

manavortex:
ok, once more unto the breach :)

rfuzzo:
Doesn't seem to work. 

Repro: 
1. open `base\gameplay\gui\widgets\vehicle\porsche\911_turbo_ui.inkwidget`
2. go to librayItems > package > root widget > children > children: duplicate the first element
3. click on parentWidget > reset
4. click on librayItems > package > root widget > copy handle
5. go back to parentWidget 

"Paste as Handle is missing"

rfuzzo (approved):
I reviewed myself and approved it fully

PR #2299, opened 2025-03-12
meshImportTools: LOD_0 => LOD_8

manavortex:
# meshImportTools: LOD_0 => LOD_8

If a mesh with LOD_0 is found, try renaming to LOD_8 before throwing exception


rfuzzo:
Why? Could you explain what this fixes and why it's needed?

manavortex:
image
https://discord.com/channels/717692382849663036/808086068023918603/1349392047219413002 this exact error :D

rfuzzo:
Yes but I still don't understand why we should rename a file called LOD0 to LOD8

rfuzzo:
Let's close this for now, the exporters and importers need to work without hacks

PR #2311, opened 2025-03-27
feature: project -> delete empty meshes

manavortex:
# feature: project -> delete empty meshes

Will also try to delete obsolete components while it's at it


rfuzzo (approved):
LGTM

PR #2313, opened 2025-03-29
better default tweak and xl paths

manavortex:
# better default tweak and xl paths

- xl files will now have `project_name.archive.extension` as default name upon creation
- tweaks will now be in `r6\tweaks\modderName\` and be called `project_name.extension`
- scripts will now be in `r6\scripts\modderName\` and be called `project_name.extension`

### DIFF
--- a/WolvenKit.App/Factories/DialogViewModelFactory.cs
+++ b/WolvenKit.App/Factories/DialogViewModelFactory.cs
@@ -10,18 +10,21 @@ public class DialogViewModelFactory : IDialogViewModelFactory
     private readonly ILoggerService _loggerService;
     private readonly IProjectManager _projectManager;
     private readonly INotificationService _notificationService;
+    private readonly ISettingsManager _settingsManager;
 
     public DialogViewModelFactory(
         IProjectManager projectManager,
         ILoggerService loggerService,
-        INotificationService notificationService
+        INotificationService notificationService,
+        ISettingsManager settingsManager
         )
     {
         _projectManager = projectManager;
         _loggerService = loggerService;
         _notificationService = notificationService;
+        _settingsManager = settingsManager;
 
     }
     public SoundModdingViewModel SoundModdingViewModel() => new(_notificationService, _loggerService, _projectManager);
-    public NewFileViewModel NewFileViewModel() => new(_projectManager);
+    public NewFileViewModel NewFileViewModel() => new(_projectManager, _settingsManager);
 }
--- a/WolvenKit.App/Services/ISettingsManager.cs
+++ b/WolvenKit.App/Services/ISettingsManager.cs
@@ -196,4 +196,5 @@ public interface ISettingsManager : ISettingsDto, INotifyPropertyChanged
     int NumFilesToReopen { get; set; }
 
     bool ShowVerboseLogOutput { get; set; }
+    bool GroupTweaksAndScriptsByModderName { get; set; }
 }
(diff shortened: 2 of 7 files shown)


MadradMagerman:
Will Wkit be able to automatically copy the .tweak extension from the `resource` folder to the game folder? Or does it remain as usual, only `.xl` and `.yaml`

manavortex:
> Will Wkit be able to automatically copy the .tweak extension from the `resource` folder to the game folder? Or does it remain as usual, only `.xl` and `.yaml`

Put it in the correct subfolder in resources, and it'll get copied

rfuzzo:
This changes the file defaults from `r6/tweaks/<project_name>/untitled.tweak` to `r6/tweaks/<author_name>/<project_name>.tweak`

But this is not a good change. At the very least it should be `r6/tweaks/<author_name>/<project_name>/untitled.tweak`

The addition of the author folder is fine, but I see no reason to name the file after the project.

rfuzzo:
One can have multiple tweak files for different purposes in one project and it's messy to all name them the same.

manavortex:
Of course, but before the suggested name would be `untitled.yaml`, which isn't any better. It's just what's prefilled in the edit box if you create a new file, you'd probably edit it anyway.

rfuzzo:
Untitled is the standard default for new files for most programs I know. But regardless of the default name, the new folder structure doesn't look like an improvement to me

manavortex:
That's how almost everyone effectively groups their tweaks, though :D

rfuzzo:
It would only make sense to have /author/project/file.tweak.

I certainly do not want to make /author/project.tweak, author/project2.tweak, author/project2_file2.tweak the default

manavortex:
People can still group their files into the folder later. Most do that

rfuzzo:
Indeed, that's exactly why we shouldn't give them bad defaults!

manavortex:
We already gave a bad default (r6/tweaks/untitled.yaml), IMO this makes it a lot better. At least for my use cases of one tweak or script per mod, I'll never have to rename anything. 
Would you prefer the scripts subdir'd by mod name?

rfuzzo:
That's not true. We give them r6/tweaks/<project_name>/untitled.tweak.

This is much better than r6/tweaks/<author>/<project_name>.tweak

It's better because your change would lump different mods into one big author folder. This is very unwieldy 

Like I said, I'm happy with adding an additional author folder, but we should not lump all tweak files into it

rfuzzo:
Also your use case for which this change is for, is not the standard. We support multiple tweak files per mod and while it may be easier for you, we shouldn't make life more difficult for people that do it like that.

manavortex:
It's standard among clothing and pose modders at least, which is why I implemented it in the first place. I guess I could add a setting for that?

rfuzzo:
I was looking through nexus right now and virtually all use our normal r6/tweaks/modname/file.tweak structure so I'm even less convinced we should change anything tbh

rfuzzo:
Tbh I'm not a fan of making this even a setting. The default makes sense. If users want to move files into their own folders they can simply do it in windows explorer.

manavortex:
I'm not sure what mods you looked at, but this is my tweaks folder right now:
image
image

rfuzzo:
As I said before, I don't see a problem with making it /author/project/file.tweak. Just not /author/project.tweak

PR #2314, opened 2025-03-30
File validation on entire project: Made it less spammy

manavortex:
# File validation on entire project: Made it less spammy

Logger will not print script execution time if you run file validation on the entire project


rfuzzo (commented):
Lgtm but it also ignores mlmask files now and didn't before. Is that intended? I didn't see it in the PR description?

manavortex:
Yeah, it's not implemented for mlmask files :D There's also sort of nothing to validate.

rfuzzo (approved):
Alright

PR #2316, opened 2025-04-01
mesh tools: unparented bones don't throw an exception anymore

manavortex:
# mesh tools: unparented bones don't throw an exception anymore


rfuzzo (approved):
👍 is it useful to the user if we add a warning that the mapping didn't contain the key?

manavortex:
nah, you can see that in Blender

PR #2318, opened 2025-04-02
.xl default file name => projectname.archive.xl

manavortex:
# .xl default file name => projectname.archive.xl

currently doesn't have one


rfuzzo (approved):
👍

PR #2322, opened 2025-04-07
prevent endless loop during startup

manavortex:
# prevent endless loop during startup

hidden / autohidden would cause an endless refresh loop where the dockState would jump between these two values. Way safer this way.


rfuzzo (approved):
Sure, though I don't get why this happens. A bug in SF?

PR #2328, opened 2025-04-10
Add contextual json convert to project browser

manavortex:
# Add contextual convert to project browser

## With shift key held down 
only active if corresponding file exists

image
image

## Open in MLSB
...will also be available from project browser now. 
If file does not exist: will create file
If file does exist: will ask if the file should be overwritten
image

### DIFF
--- a/WolvenKit.App/Models/FileSystemModel.cs
+++ b/WolvenKit.App/Models/FileSystemModel.cs
@@ -21,7 +21,19 @@ public class FileSystemModel : INotifyPropertyChanged
     private string _extension = "default";
 
 
+    private static readonly string s_archiveSubfolder =
+        $"source{Path.DirectorySeparatorChar}archive{Path.DirectorySeparatorChar}";
+
+    private static readonly string s_rawSubfolder =
+        $"source{Path.DirectorySeparatorChar}raw{Path.DirectorySeparatorChar}";
+
     [Browsable(false)] public FileSystemModel? Parent { get; }
+
+    [Browsable(false)] public bool IsInArchive => FullName.Contains(s_archiveSubfolder);
+    [Browsable(false)] public bool IsInRaw => FullName.Contains(s_rawSubfolder);
+
+    [Browsable(false)]
+    public bool IsGameFile => GameRelativePath.StartsWith("base") || GameRelativePath.StartsWith("ep1");
 
     public string Name
     {
--- a/WolvenKit.App/Models/FileSystemModel.cs
+++ b/WolvenKit.App/Models/FileSystemModel.cs
@@ -20,21 +20,8 @@ public class FileSystemModel : INotifyPropertyChanged
     private string _fileSizeStr = null!;
     private string _extension = "default";
 
-
-    private static readonly string s_archiveSubfolder =
-        $"source{Path.DirectorySeparatorChar}archive{Path.DirectorySeparatorChar}";
-
-    private static readonly string s_rawSubfolder =
-        $"source{Path.DirectorySeparatorChar}raw{Path.DirectorySeparatorChar}";
-
     [Browsable(false)] public FileSystemModel? Parent { get; }
-
-    [Browsable(false)] public bool IsInArchive => FullName.Contains(s_archiveSubfolder);
-    [Browsable(false)] public bool IsInRaw => FullName.Contains(s_rawSubfolder);
-
-    [Browsable(false)]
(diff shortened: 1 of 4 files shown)


rfuzzo:
What's the difference to what we already have?

rfuzzo (changes requested):
It doesn't work in-app. Nothing happens when I click on a json file in raw, hold shift and press "convert game file"

rfuzzo:
Also doesn't work the other way.

manavortex:
nothing _visible_ happens. The file should be re-created?

rfuzzo:
Nothing happens: no log entry, the file is not recreated.

rfuzzo:
Repro steps:
- add an mlsetup file to the project
- right click: convert to json
- then go to the json file in raw and open it in a text editor and delete all contents (now it's an empty file)
- go to the json file and shift-right-click "convert from game file"
- nothing happens: no log entry, the file is still empty

manavortex:
I do believe you, I'll double-check after work :)

manavortex:
Fixed

WolvenKit.App/Models/FileSystemModel.cs
```diff
@@ -21,8 +21,20 @@ public class FileSystemModel : INotifyPropertyChanged
     private string _extension = "default";
 
 
+    private static readonly string s_archiveSubfolder =
```
rfuzzo, line 24: This shouldn't be done like this. We already had Functions "IsInRawFolder" and "IsInArchiveFolder" that did this better.

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -554,7 +556,9 @@ private void CopyAbsPathToCurrentFolder() =>
     /// <summary>
     /// Reimports the game file to replace the current one
     /// </summary>
-    private bool CanAddDependencies() => ActiveProject != null && SelectedItem != null && IsInRawFolder(SelectedItem) && SelectedItem.Extension.ToLower().Contains("xml");
+    private bool CanAddDependencies() => ActiveProject != null && SelectedItem?.IsInRaw == true &&
```
rfuzzo, line 559: Rather use the existing IsInRawFolder methods

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -861,24 +858,64 @@ await _projectResourceTools.MoveAndRefactor(_projectManager.ActiveProject, relat
 
     #region red4
 
-    private bool IsInArchiveFolder(FileSystemModel? model) =>
```
rfuzzo, line 864: Don't delete this

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -861,24 +858,64 @@ await _projectResourceTools.MoveAndRefactor(_projectManager.ActiveProject, relat
...
-        ActiveProject is not null && model is not null && model.FullName.Contains(ActiveProject.ModDirectory);
+    private bool HasCorrespondingConvertFile(FileSystemModel? model)
+    {
+        if (model is null || ActiveProject is null)
+        {
+            return false;
+        }
 
-    private bool IsInRawFolder(FileSystemModel? model) =>
```
rfuzzo, line 867: Don't delete this

rfuzzo (changes requested):

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
manavortex: I moved these into properties of the model, seemed more object oriented that way :) any reason for that?

rfuzzo: Yeah the reason is that the "is in raw" check should check if the path is inside the project folder, and not just a substring check

manavortex: that stuff was super confusing, because the `relativeXXX` paths were different depending on whether the item is in raw or not, so the logic was flakey here and didn't work like I expected it :|

rfuzzo (approved):
yup works now in-app

PR #2332, opened 2025-04-12
cybersearch will no longer insist on spaces before > operator

manavortex:
# cybersearch will no longer insist on spaces before > operator

as discussed in wkit-dev last week :)


rfuzzo (changes requested):
Crashes with a threading exception on search now. 

Repro:
1. Asset browser > search for e.g. "judy>wet"
2. Observe crash

manavortex:
The exception happens only while WKit is starting up, I'm catching it now. The replace does not seem to do anything (Auska said it would), but the search works, I'm happy :)

rfuzzo:
Yeah I'm not a fan of the replacing tbh I think we don't need it

PR #2333, opened 2025-04-12
WKit will now export withRig at bone mismatch

manavortex:
# WKit will now export withRig at bone mismatch

Exception will print a warning for export jobs


rfuzzo (approved):
seems okay

PR #2338, opened 2025-04-14
fix "regenerate visual controllers"

manavortex:
# fix "regenerate visual controllers"


rfuzzo:
What does this fix exactly?

manavortex:
an unspecific exception if the wrong node is selected when you push the menu item

manavortex:
open an .app file, select a component node, and select "regenerate visual controllers" from the menu. Should do it

rfuzzo (approved):
Got the error 👍

PR #2339, opened 2025-04-16
preventing exception for morphtarget export without texture

manavortex:
# preventing exception for morphtarget export without texture

it's not as if we were using it for anything anyway


rfuzzo (approved):
Ok

PR #2341, opened 2025-04-17
fixed non-empty default paths (no hiding via editor mode!)

manavortex:
# editor mode: don't hide non-empty depot paths


rfuzzo (approved):
👍

PR #2349, opened 2025-04-20
Editor difficulty level: global property

manavortex:
# Editor difficulty level will no longer be set per editor tab

### DIFF
--- a/WolvenKit.App/ViewModels/Documents/RedDocumentViewModel.cs
+++ b/WolvenKit.App/ViewModels/Documents/RedDocumentViewModel.cs
@@ -64,8 +64,8 @@ public partial class RedDocumentViewModel : DocumentViewModel
         IArchiveManager archiveManager,
         IHookService hookService,
         INodeWrapperFactory nodeWrapperFactory,
-        EditorDifficultyLevel editorMode,
         Cr2WTools cr2WTools,
+        ISettingsManager settingsManager,
         bool isReadyOnly = false) : base(path)
     {
         _documentTabViewmodelFactory = documentTabViewmodelFactory;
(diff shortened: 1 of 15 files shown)


WolvenKit/Views/Documents/RedDocumentView.xaml.cs
```diff
@@ -36,25 +36,7 @@ private void OnEditorDifficultyChanged(object sender, EditorDifficultyLevel leve
                 return;
             }
 
-            if (vm.IsDirty)
```
rfuzzo, line 39: why was this deleted?

WolvenKit.App/ViewModels/Documents/RedDocumentViewModel.cs
```diff
@@ -89,18 +89,41 @@ public RedDocumentViewModel(CR2WFile file, string path, AppViewModel appViewMode
 
         _path = path;
 
+        if (s_globalEditorDifficultyLevel == EditorDifficultyLevel.None)
+        {
+            s_globalEditorDifficultyLevel = settingsManager.DefaultEditorDifficultyLevel;
+        }
 
+        EditorDifficultyLevel = s_globalEditorDifficultyLevel;
+        GlobalEditorDifficultyLevelChanged += OnGlobalEditorDifficultyChanged;
```
rfuzzo, line 98: This never gets unsubscribed

WolvenKit.App/ViewModels/Documents/RedDocumentViewModel.cs
```diff
@@ -89,18 +89,41 @@ public RedDocumentViewModel(CR2WFile file, string path, AppViewModel appViewMode
...
     }
 
+    private void OnGlobalEditorDifficultyChanged(object? sender, EventArgs e) =>
+        EditorDifficultyLevel = s_globalEditorDifficultyLevel;
+
     #region properties
 
+    private static event EventHandler? GlobalEditorDifficultyLevelChanged;
+
+    private static EditorDifficultyLevel s_globalEditorDifficultyLevel = EditorDifficultyLevel.None;
```
rfuzzo, line 115: rather than having static fields here, put them into the Settingsamanager

rfuzzo (changes requested):

WolvenKit.App/ViewModels/Documents/RedDocumentViewModel.cs
manavortex: you mean another hidden field that doesn't get serialized/read back? because the standard should be applied on first start

rfuzzo: yeah exactly. But I mean it could be serialized tbh Seems like something the user wants to have persist

WolvenKit/Views/Documents/RedDocumentView.xaml.cs
manavortex: the reload isn't needed anymore to apply the layout changes (that was due to the broken bindings introduced by your PR that changed the SelectedItems from <object> to <CVM>).

PR #2351, opened 2025-04-21
Trying to prevent "reset layout" bug

manavortex:
# Trying to prevent "reset layout" bug

Occasionally, resetting the project layout will reset in the view being covered in... a lot of windows. We know the bug, it's still there (although rarely). 
This is another attempt at preventing it. We'll see if it works, at least things haven't gotten worse.

### DIFF
--- a/WolvenKit/Views/Shell/DockingAdapter.xaml.cs
+++ b/WolvenKit/Views/Shell/DockingAdapter.xaml.cs
@@ -220,7 +220,8 @@ namespace WolvenKit.Views.Shell
             {
                 PART_DockingManager.BeginInit(); // Begin batch updates
 
-                using (var reader = XmlReader.Create(filePath)) {              
+                using (var reader = XmlReader.Create(filePath))
+                {              
 
                     var defaultXmlSerializer = DockingManager.CreateDefaultXmlSerializer(typeof(List<DockingParams>));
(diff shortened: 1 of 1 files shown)


WolvenKit/Views/Shell/DockingAdapter.xaml.cs
```diff
@@ -274,8 +278,12 @@ private bool LoadLayout(string filePath, string layoutSource = "")
             {
                 _logger.Error(e);
             }
+            finally
+            {
+                reader?.Close();
```
rfuzzo, line 283: XmlReader is IDisposable, should probably Dispose it, or wrap in a using clause... Not sure why we're not doing it here

manavortex: not sure, I just moved the close into the finally case :D

manavortex:
I don't know how - ran into the bug again 10 minutes before filing this pull request... but it's sporadic.

WolvenKit/Views/Shell/DockingAdapter.xaml.cs
manavortex: after work :) Feel free to take over!

rfuzzo (approved):
looks good and works in-app, thanks for using the IDisposable now

PR #2352, opened 2025-04-21
Cvm selection and descriptor

manavortex:
# Cvm selection and descriptor

- Improved selection behaviour for duplicating single items
- Added descriptor for `gameEntitySpawnerComponent` in CVM

### DIFF
--- a/WolvenKit.App/ViewModels/Documents/RDTDataViewModel.cs
+++ b/WolvenKit.App/ViewModels/Documents/RDTDataViewModel.cs
@@ -416,6 +416,7 @@ public partial class RDTDataViewModel : RedDocumentTabViewModel
     public void AddToSelection(ChunkViewModel? chunk)
     {
         var selectedChunks = SelectedChunks?.ToList() ?? [];
+        
         ClearSelection();
         if (chunk is null)
         {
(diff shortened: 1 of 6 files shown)


WolvenKit.App/ViewModels/Documents/RDTDataViewModel.cs
```diff
@@ -473,21 +474,23 @@ public void SetSelection(List<ChunkViewModel> chunks)
...
-
+        
         // remove duplicates
         var uniqueChunks = new HashSet<ChunkViewModel>(chunks);
-        SelectedChunk = uniqueChunks.LastOrDefault();
-
         foreach (var cvm in uniqueChunks)
         {
-            cvm.IsSelected = true;
+            // cvm.IsSelected = true;
```
rfuzzo, line 482: why comment this out?

manavortex: because it works this way. Should have deleted it

rfuzzo: yeah but where is the IsSelected property then set?

manavortex: It happens implicitly via binding, I think? All this stuff is too nested >.< I guess commenting this back in won't hurt, I'll test after work :)

rfuzzo: nah I just wanted to know the reason for this change... Yeah I guess it is set through the OnSelected Event. lemme check

manavortex: And? :D

rfuzzo (approved):
seems to work

PR #2355, opened 2025-04-24
generate inkatlas. it just works.

manavortex:
# generate inkatlas. it just works.

File -> Add... now has an inkatlas dialog, with custom popup and everything


rfuzzo (approved):
seems to work, I didn't check all kinds of possibilities though

PR #2364, opened 2025-04-30
Added CName Hashmap Key to Hash Tool

flibdev:
# Added CName Hashmap Key to Hash Tool

**Additional notes:**
Was useful in decoding the shader cache files, might be useful for someone else in the future.


rfuzzo (approved):
👍

PR #2368, opened 2025-05-02
prevent throwing exception for duplicate components

manavortex:
# ent preview: exception fix

prevent throwing exception for duplicate components


rfuzzo (approved):
seems to work

PR #2370, opened 2025-05-03
AskForTextInput: allow passing of text value

manavortex:
# AskForTextInput: allow passing of text value

Also added "move to index..." to CVM's material context menu


manavortex:
I added this to `close-active-document-improvements` to prevent merge conflicts

PR #2371, opened 2025-05-03
"your document has changes" popup

manavortex:
# please review by commit
  
  
  
----------
  
  
  
## "your document has changes" popup

- clearer buttons / better popup
image

- put all the logic in one place
- closing tabs feels LOTS faster now

- DockingAdapter: Ctrl+W closes active document

## rename dialogue
- Now takes two parameters: title and previous (default) value

## cvm

- added "move to index..." to CVM's material context menu


WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -3416,12 +3416,27 @@ private bool CreateArray()
...
+        {
+            CalculateProperties();
+        }
+
+        if (Properties is not ObservableCollectionExtended<ChunkViewModel> children)
+        {
+            return null;
+        }
+
+        if (index < children.Count)
```
rfuzzo, line 3434: I think you meant to return null here.

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -3416,12 +3416,27 @@ private bool CreateArray()
...
     public ChunkViewModel? GetChildNode(int index)
     {
-        if (!IsArray || Properties is not ObservableCollectionExtended<ChunkViewModel> children ||
-            index >= children.Count)
+        if (!IsArray)
         {
             return null;
         }
 
+        if (Properties.Count == 0)
```
rfuzzo, line 3424: Move this after the return null checks

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
manavortex: ?

WolvenKit/Views/Shell/DockingAdapter.xaml.cs
```diff
@@ -345,11 +343,24 @@ private async Task<bool> TryCloseDocument(DocumentViewModel vm)
         /// </summary>
         /// <param name="sender"></param>
         /// <param name="e"></param>
-        private async void PART_DockingManagerOnCloseButtonClick(object sender, CloseButtonEventArgs e)
+        private void PART_DockingManagerOnCloseButtonClick(object sender, CloseButtonEventArgs e)
         {
             if (e.TargetItem is ContentControl { Content: DocumentViewModel vm })
             {
-                e.Cancel = !await TryCloseDocument(vm);
```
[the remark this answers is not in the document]
manavortex: I get linter warnings about async void methods - the compiler wants those to have Task as return type, which doesn't work for the control interface. This here makes the problem go away.

WolvenKit/Views/Shell/DockingAdapter.xaml.cs
```diff
@@ -358,11 +369,11 @@ private async void PART_DockingManagerOnCloseButtonClick(object sender, CloseBut
         /// </summary>
         /// <param name="sender"></param>
         /// <param name="e"></param>
-        private async void PART_DockingManager_WindowClosing(object sender, WindowClosingEventArgs e)
+        private void PART_DockingManager_WindowClosing(object sender, WindowClosingEventArgs e)
         {
             if (e.TargetItem is ContentControl { Content: DocumentViewModel vm })
             {
-                e.Cancel = !await TryCloseDocument(vm);
```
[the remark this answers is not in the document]
manavortex: I get linter warnings about async void methods - the compiler wants those to have Task as return type, which doesn't work for the control interface. This here makes the problem go away.

WolvenKit/Views/Shell/DockingAdapter.xaml.cs
```diff
@@ -479,7 +490,8 @@ private async void PART_DockingManager_OnCloseAllTabs(object sender, CloseTabEve
                     }
                 });
 
-            await Task.WhenAll(closeTasks);
+            Task.WhenAll(closeTasks).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
manavortex: I get linter warnings about async void methods - the compiler wants those to have Task as return type, which doesn't work for the control interface. This here makes the problem go away.

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
rfuzzo: The original check is if greater than or equal and returns null, your check doesn't do this anymore

manavortex: Right! Thanks, fixed

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -3416,8 +3416,22 @@ private bool CreateArray()
...
+        {
+            return null;
+        }
+
+        if (Properties is not ObservableCollectionExtended<ChunkViewModel> children)
+        {
+            return null;
+        }
+
+        if (Properties.Count == 0)
```
rfuzzo, line 3429: Alright, let's now just move this after the null checks

rfuzzo (changes requested):

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
manavortex: Can't do that - if properties isn't null but has the length 0, CalculateProperties will populate it. (Unless there really are no properties)

rfuzzo: No I meant move the calculate properties after the children index check

manavortex: Can you please suggest a change? I think we're talking past each other

rfuzzo: I meant this

WolvenKit/Views/Shell/DockingAdapter.xaml.cs
```diff
@@ -476,25 +494,43 @@ private void PART_DockingManager_ActiveWindowChanged_1(DependencyObject d, Depen
...
+        public async Task<bool> CloseAllAsync()
+        {
+            if (_viewModel == null)
+            {
+                return true;
+            }
+
+            var allClosed = true;
+
+            for (var i = _viewModel.DockedViews.Count - 1; i >= 0; i--)
```
manavortex, line 519: Would it not make sense to do an await Task.WhenAll here, rather than serially awaiting?

rfuzzo (approved):

PR #2372, opened 2025-05-03
refactor&rename will now replace in cnames (CPUNameU64)

### LINKED ISSUE #2369: Rename Path with Update Files issue
Renaming folder/file path in Wolvenkit while selecting 'update files' skips over @context material in ccxl hair.

 - Version 8.16.1

Image

manavortex:
# refactor&rename will now replace in cnames (CPUNameU64)

It will also no longer work on a per-folder base (false positives), but rather only replace files that actually exist in the project.

**Additional notes:**
closes #2369

### DIFF
--- a/WolvenKit.App/Helpers/ProjectResourceTools.cs
+++ b/WolvenKit.App/Helpers/ProjectResourceTools.cs
@@ -429,8 +429,7 @@ public class ProjectResourceTools
         // the user is moving an empty directory
         if (files.Count == 0 && sourceIsDirectory)
         {
-            MoveDirectoryTreeWithoutFiles(sourceFileOrDirAbsPath, destAbsPath);
-            DeleteEmptyDirectoriesRecursive(sourceFileOrDirAbsPath);
+            Directory.Move(sourceFileOrDirAbsPath, destAbsPath);
             return;
         }
 
@@ -518,29 +517,7 @@ public class ProjectResourceTools
         await ReplacePathInProjectAsync(activeProject, successfulReplacements);
         
     }
-
-    private static void MoveDirectoryTreeWithoutFiles(string sourceAbsPath, string destAbsPath)
-    {
-        if (!Directory.Exists(sourceAbsPath))
-        {
-            return;
-        }
-
-        if (!Directory.Exists(destAbsPath))
-        {
-            Directory.CreateDirectory(destAbsPath);
-        }
-
-        foreach (var directory in Directory.GetDirectories(sourceAbsPath, "*", SearchOption.AllDirectories))
-        {
-            var relativePath = Path.GetRelativePath(sourceAbsPath, directory);
-            var targetDirPath = Path.Combine(destAbsPath, relativePath);
-
-            Directory.CreateDirectory(targetDirPath);
-        }
-
-        Directory.Delete(sourceAbsPath, true);
-    }
+   
 
     /// <summary>
     /// Deletes all empty directories in a tree (directories not containing files), finally deleting the root if it's empty
--- a/WolvenKit.App/Helpers/ProjectResourceTools.cs
+++ b/WolvenKit.App/Helpers/ProjectResourceTools.cs
@@ -17,9 +17,7 @@ using WolvenKit.Core.Interfaces;
 using WolvenKit.Helpers;
 using WolvenKit.Interfaces.Extensions;
 using WolvenKit.RED4.Archive.CR2W;
-using WolvenKit.RED4.Archive.IO;
 using WolvenKit.RED4.Types;
-using EFileReadErrorCodes = WolvenKit.Common.EFileReadErrorCodes;
 
 namespace WolvenKit.App.Helpers;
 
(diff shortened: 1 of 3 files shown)


WolvenKit.RED4/Archive/IO/CR2WReader.File.cs
```diff
@@ -73,13 +73,13 @@ public EFileReadErrorCodes ReadFileInfo(out CR2WFileInfo? info, ILoggerService?
     public EFileReadErrorCodes ReadFile(out CR2WFile? file, bool parseBuffer = true)
     {
         var result = ReadFileInfo(out var info);
-        if (result == EFileReadErrorCodes.NoCr2w)
+        if (result == EFileReadErrorCodes.NoCr2w || info is null || info.StringDict.Count == 0)
```
seberoth, line 76: Just as info: `info` can't be `null` below, so the `!` should be enough. But the null check to suppress the compiler warning below is also fine.

But adding `info.StringDict.Count == 0` here is technicly wrong. If the previous checks are false, it is a CR2W file and the empty StringDict has another reason. The return value of `NoCr2w` is wrong then.

WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -358,141 +361,208 @@ public async Task MoveAndRefactor(Cp77Project activeProject, string sourcePath,
...
+            var relativePath = Path.GetRelativePath(sourceAbsPath, directory);
+            var targetDirPath = Path.Combine(destAbsPath, relativePath);
+
+            Directory.CreateDirectory(targetDirPath);
+        }
+
+        Directory.Delete(sourceAbsPath, true);
+    }
+
+    private static void DeleteEmptyDirectoriesRecursive(string directoryPath)
```
rfuzzo, line 514: Should this function 
1. delete all directories recursively if the whole tree is only empty directories or
2. delete only the empty directories of a tree that may also include non-empty directories?

WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -358,141 +361,208 @@ public async Task MoveAndRefactor(Cp77Project activeProject, string sourcePath,
...
+            DeleteEmptyDirectoriesRecursive(subDirectory);
+        }
+
+        if (Directory.GetDirectories(directoryPath, "*", SearchOption.AllDirectories).Length == 0)
+        {
+            Directory.Delete(directoryPath, true); // Delete the root directory
+        }
+    }
+
+    private Task MoveFileAsync(string sourcePath, string targetPath, Cp77Project activeProject, bool isDirectory)
```
rfuzzo, line 533: this isn't an async method and should not be called async

WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -358,141 +361,208 @@ public async Task MoveAndRefactor(Cp77Project activeProject, string sourcePath,
...
+    {
+        try
```
rfuzzo, line 535: couldn't we just use FileInfo.MoveTo ?

manavortex: That caused other issues in the past, which is why this function exists :/

WolvenKit.App/Helpers/ProjectResourceTools.cs
manavortex: This should delete everything that is an empty directory, but nothing that isn't

WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -350,514 +342,531 @@ private void AddFileToProjectFolder(string projectRoot, ResourcePath resourcePat
...
-
-        ReplacePathInProject(activeProject, sourceAbsPath, newRelPath);
+        await ReplacePathInProjectAsync(activeProject, successfulReplacements);
     }
 
-    /// <summary>
-    /// File extensions in resources that we want to update  
-    /// </summary>
-    private static readonly List<string> s_resourceFileExtensions = [".lua", ".xl", ".json", ".yaml", ".yml", ".reds"];
+    private static void MoveDirectoryAndChildren(string sourceAbsPath, string destAbsPath)
```
notaspirit, line 477: why do we need a special method for moving just empty directories? 
Also from the method name it isn't obvious that it will only copy the directory structure not the content

WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -350,514 +342,531 @@ private void AddFileToProjectFolder(string projectRoot, ResourcePath resourcePat
...
+        }
 
-    /// <summary>
-    /// Extensions of files that we don't need to update 
-    /// </summary>
-    private static readonly List<string> s_fileExtensionsWithoutPaths = [".xbm", ".mlmask"];
+        if (!Directory.Exists(destAbsPath))
+        {
+            Directory.CreateDirectory(destAbsPath);
+        }
```
notaspirit, line 487: checking if the directory exists before calling Directory.CreateDirectory() isn't required as that method already handles that case see [.NET Documentation](https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?view=net-9.0&viewFallbackFrom=windowsdesktop-8.0), also you just need to call it for the top level directories that don't contain any more directories as it will create the necessary subdirectories automatically

WolvenKit.App/Helpers/ProjectResourceTools.cs
manavortex: Because I changed the logic to only apply files, and then delete empty directories after. Then I found myself trying to move an (as of yet) empty directory tree, which didn't work.

notaspirit: replacing `MoveDirectoryTreeWithoutFiles(sourceFileOrDirAbsPath, destAbsPath)` and `DeleteEmptyDirectoriesRecursive(sourceFileOrDirAbsPath)` with `Directory.Move(sourceFileOrDirAbsPath, destAbsPath)` works as expected without errors when renaming / moving a directory with 0 files

rfuzzo (approved):
Checked in-app and seems to work.

PR #2374, opened 2025-05-04
gameJournalEntry: CVM value display

manavortex:
# gameJournalEntry: CVM value display

image

### DIFF
--- a/WolvenKit.App/Helpers/StringHelper.cs
+++ b/WolvenKit.App/Helpers/StringHelper.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Text;
 using System.Text.RegularExpressions;
 using System.Windows.Media;
 using DynamicData;
@@ -12,22 +13,6 @@ namespace WolvenKit.App.Helpers;
 
 public abstract partial class StringHelper
 {
-    public static string Stringify(CArray<scnOutputSocket> scnOutputAry)
-    {
-        if (scnOutputAry.Count == 0)
-        {
-            return "";
-        }
-
-        var nodeIds = scnOutputAry.Select(
-            scnOutput => Stringify(
-                scnOutput.Destinations.Select(dest => dest.NodeId.Id.ToString())
-                    .Where(s => s != "").ToArray()
-            )
-        ).ToArray();
-        return Stringify(nodeIds);
-    }
-
     public static string Stringify(Vector4 vec, bool defaultIsOne = false)
     {
         var val = defaultIsOne ? 1 : 0;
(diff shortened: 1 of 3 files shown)


rfuzzo:
The stringify changes are good, but the text file validation changes aren't: the are packed when they are in resources, and they may have an effect if the modders wants to. I think this commit should be reverted.

manavortex:
I just added that because of someone's use case, but let me add a better check real quick

WolvenKit.App/ViewModels/Shell/AppViewModel.WScript.cs
```diff
@@ -76,6 +76,19 @@ private async Task RunFileValidationOnProject()
             return;
         }
 
+        var txtFilesInResources = ActiveProject.ResourceFiles.Where(f => f.EndsWith(".txt"))
+            .Select(f => ActiveProject.GetRelativeResourcePath(f).GetResolvedText())
```
rfuzzo, line 80: Why is it need to get the relative resource path if we only check the extension?

WolvenKit.App/ViewModels/Shell/AppViewModel.WScript.cs
```diff
@@ -94,6 +107,12 @@ private async Task RunFileValidationOnProject()
 
         foreach (var file in filesToValidate)
         {
+            if (file.Extension == ".txt")
```
rfuzzo, line 110: Does this check all filed in the archive folder? If yes, then the validation check should be "if extension is not a REDextension".

rfuzzo (changes requested):

WolvenKit.App/ViewModels/Shell/AppViewModel.WScript.cs
manavortex: Good point. What's a quick way to check that?

rfuzzo: We have an enum for it: ERedExtension iirc

WolvenKit.App/ViewModels/Shell/AppViewModel.WScript.cs
manavortex: For the output below

WolvenKit.App/ViewModels/Shell/AppViewModel.WScript.cs
```diff
@@ -76,6 +76,19 @@ private async Task RunFileValidationOnProject()
             return;
         }
 
+        var txtFilesInResources = ActiveProject.ResourceFiles.Where(f => f.EndsWith(".txt"))
+            .Select(f => ActiveProject.GetRelativeResourcePath(f).GetResolvedText())
+            .Where(f => f is not null)
+            .Where(f => Path.GetExtension(f?.Replace(".txt", "")) is not null)
```
rfuzzo, line 82: It should check for string is not empty here not null.

rfuzzo (changes requested):

WolvenKit.App/ViewModels/Shell/AppViewModel.WScript.cs
manavortex: fixed

rfuzzo (approved):
Looks good

PR #2376, opened 2025-05-04
fixed packing soundbanks.json

JakubMarecek:
Fixed packing file `base\sound\event\soundbanks.json`


rfuzzo:
What was the problem here exactly?

JakubMarecek:
Since it's json, it packed it with compression, but this file is pure json format, without cr2w binary, so I added condition for this single file to pack it without compressing.

rfuzzo (approved):
👍

PR #2381, opened 2025-05-07
Initialize fields in cvm

manavortex:
# Initialize fields in cvm

now uses a factory to initialize a bunch of fields in cvm nodes

image

### DIFF
--- a/WolvenKit.RED4/Types/Classes/questAddRemoveContact_NodeType.cs
+++ b/WolvenKit.RED4/Types/Classes/questAddRemoveContact_NodeType.cs
@@ -14,7 +14,7 @@ namespace WolvenKit.RED4.Types
 
 		public questAddRemoveContact_NodeType()
 		{
-			// Params = new() { new questChangeContactList_NodeTypeParams() };
+			Params = new() { new questChangeContactList_NodeTypeParams() };
 
 			PostConstruct();
 		}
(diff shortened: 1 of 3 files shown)


WolvenKit.RED4/Types/Classes/questAddRemoveContact_NodeType.cs
```diff
@@ -14,7 +14,7 @@ public CArray<questChangeContactList_NodeTypeParams> Params
 
 		public questAddRemoveContact_NodeType()
 		{
-			Params = new() { new questChangeContactList_NodeTypeParams() };
```
rfuzzo, line 17: That's the same problem as originally: can't change stuff here as it will break the serializer. This commit came from a different PR?

rfuzzo (changes requested):

WolvenKit.RED4/Types/Classes/questAddRemoveContact_NodeType.cs
manavortex: no, I just ran into the issue when working on this (probably because it broke a mesh). I can fix the check so that it returns a cr2w, but I stress, _its absence led to an un-caught exception_.

rfuzzo: What check? But please don't change stuff in the /Types folders as they are taken from the engine as-is.

manavortex: Ohh - sorry, wrong PR. @MrBilL61 we'll have to revert this one, sorry

MrBilL61: > What check? But please don't change stuff in the /Types folders as they are taken from the engine as-is.

I wasn't aware that they mustn't be change. I'll remember that.

PR #2383, opened 2025-05-08
preventing uncaught index out of bound exception

manavortex:
# preventing uncaught index out of bound exception

Without the length check, line 98 will run into an out-of-bounds exception (which results in no cr2w file being returned)


manavortex:
Yeah, it was on a different PR :) I'll do that!

seberoth (approved):
LGTM

PR #2386, opened 2025-05-10
#2380 check for OneDrive location

### LINKED ISSUE #2380: Check for OneDrive path
Since it always cause issue, maybe add a function like (see below) and check
1. The path where Wkit is executed from
2. The path of new/opened projects

And show the user a Warning MessageBox about it (like the long paths warning)

Something like this SHOULD work (not sure for non-Windows OS):
```cs
static bool IsOneDrivePath(string path)
{
    // There are more special cases, but this should work mostly
    if (path.StartsWith(@"\\?\") || path.StartsWith(@"\\.\"))
    {
        path = path[4..];
    }

    path = Path.GetFullPath(path);
    
    // %OneDrive% -> If only one of both versions is installed
    // %OneDriveConsumer% -> If both are installed, the path to the personal OneDrive
    // %OneDriveCommercial% -> If both are installed, the path to the commercial OneDrive
    return IsInEnv("OneDrive") || IsInEnv("OneDriveConsumer") || IsInEnv("OneDriveCommercial");

    bool IsInEnv(string envName)
    {
        return Environment.GetEnvironmentVariable(envName) is { } str && path.StartsWith(str);
    }
}
```

manavortex:
# Check for OneDrive in paths

- on startup (assembly location)
- on project change

Will show an annoying popup, which you can not currently disable.

**Additional notes:**
closes #2380

### DIFF
--- a/WolvenKit.App/Helpers/FilePathHelper.cs
+++ b/WolvenKit.App/Helpers/FilePathHelper.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using System.IO;
+
+namespace WolvenKit.App.Helpers;
+
+public abstract class FilePathHelper
+{
+    public static bool IsOneDrivePath(string path)
+    {
+        // There are more special cases, but this should work mostly
+        if (path.StartsWith(@"\\?\") || path.StartsWith(@"\\.\"))
+        {
+            path = path[4..];
+        }
+
+        path = Path.GetFullPath(path);
+
+        // %OneDrive% -> If only one of both versions is installed
+        // %OneDriveConsumer% -> If both are installed, the path to the personal OneDrive
+        // %OneDriveCommercial% -> If both are installed, the path to the commercial OneDrive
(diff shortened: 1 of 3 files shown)


seberoth:
LGTM

Just a question: Why did you call it from `RefreshProjectData`? Isn't in `AppViewModel_OnInitialProjectLoaded` enough?

manavortex:
I figured it'd be better to check for every project, not just for the first per session

seberoth:
Yeah, that's good.
But I tested it and `AppViewModel_OnInitialProjectLoaded` seems like get called on every project reload (not sure if it's intended or not though...).
Switched multiple times and every time I switched to the OneDrive one, I got 3 warnings (one from `AppViewModel_OnInitialProjectLoaded` and two from `RefreshProjectData`)

manavortex:
Oh... thanks for testing. Why didn't that happen for me? 👀 I'll take out the second one, then.

rfuzzo (approved):
only one warning is shown now 👍

PR #2388, opened 2025-05-10
#2373 hopefully fixed hash helper

### LINKED ISSUE #2373: HashHelper is broken
Is it really needed? 
Only `CalculateDepotPathHash(ResourcePath resourcePath)` is used in code which can be replaced with `(ulong)resourcePath` (since there isn't anything to calculated, `ResourcePath` is already a hash).

If so:
1. Why is there 3 times the same internal helper function? Would be more usefull to place it in one project which is shared with all.
2. `public static ulong CalculateDepotPathHash(string? filePath)` doesn't sanitize paths
3. `public static ulong CalculateDepotPathHash(ResourcePath resourcePath)` yields broken result for unresolveable paths

I.e.
```c#
var testPath = (ResourcePath)1234; // unresolvable, but still valid
// Call to CalculateDepotPathHash(ResourcePath resourcePath)
CalculateDepotPathHash(null) // since its unresolvable, resourcePath.GetResolvedText() == null
// Call to CalculateDepotPathHash(string? filePath)
filePath is null => return 0
```
So a valid path `(ResourcePath)1234` becomes `0` after using the method.

The logic should be:
```c#
public static ulong CalculateDepotPathHash(ResourcePath resourcePath) => (ulong)resourcePath;
public static ulong CalculateDepotPathHash(string? filePath) => filePath is null ? 0 : ResourcePath.CalculateHash(filePath);
```

manavortex:
# hopefully fixed hash helper

closes #2373


WolvenKit.Common/Tools/HashHelper.cs
```diff
@@ -16,7 +17,13 @@ internal static class HashHelper
     /// </summary>
     /// <param name="resourcePath"></param>
     /// <returns></returns>
-    public static ulong CalculateDepotPathHash(ResourcePath resourcePath) => CalculateDepotPathHash(resourcePath.GetResolvedText());
-
+    public static ulong CalculateDepotPathHash(ResourcePath resourcePath)
+    {
+        if (resourcePath.GetResolvedText() is string str && str != "")
```
seberoth, line 22: Whats that for?
It only adds overhead to the function call. Since `GetResolvedText()` will lookup the hash of `ResourcePath` in our lookup table and if it's found you do the FNV1A64 on it which will result in the same hash as you started the lookup with.

So it should be one of these:
```cs#
public static ulong CalculateDepotPathHash(ResourcePath resourcePath) => resourcePath.GetRedHash();
or
public static ulong CalculateDepotPathHash(ResourcePath resourcePath) => resourcePath;
```

Or did I miss something?

seberoth (changes requested):
Also `public static ulong CalculateDepotPathHash(string? filePath)` really should use the sanitized path. Like mentioned in the issue

WolvenKit.Common/Tools/HashHelper.cs
manavortex: That's probably me not trusting implicit casts anymore and being paranoid and/or confused. You're totally right.

rfuzzo (approved):

PR #2390, opened 2025-05-17
CName dropdown

manavortex:
# CName dropdown

Some fields are CNames, but there's a non-technical limitation of what could/should go there.

image

### DIFF
--- a/WolvenKit/Views/Templates/RedCNameEditor.xaml.cs
+++ b/WolvenKit/Views/Templates/RedCNameEditor.xaml.cs
@@ -161,6 +161,10 @@ namespace WolvenKit.Views.Editors
             }
             set
             {
+                if (string.IsNullOrEmpty(value))
+                {
+                    value = "0";
+                }
                 if (_settingsManager.ShowCNameAsHex)
                 {
                     SetValue(RedStringProperty, (CName)ulong.Parse(value, NumberStyles.HexNumber));
--- a/WolvenKit/Views/Templates/RedCNameEditor.xaml.cs
+++ b/WolvenKit/Views/Templates/RedCNameEditor.xaml.cs
@@ -72,7 +72,7 @@ namespace WolvenKit.Views.Editors
         {
             var hasWarning = false;
             var hasError = false;
-            
+
             if (Text == "None" || _settingsManager?.UseValidatingEditor != true)
             {
                 TextBoxToolTip = "CName";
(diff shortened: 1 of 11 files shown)


rfuzzo:
I don't understand what this is. Can you describe more what this does and what's it used for?

rfuzzo:
Also please set the PR to draft if it has known issues / help wanted.

manavortex:
It sets the CName's RedString to whatever the user selects in the dropdown, but also allows regular editing

rfuzzo:
What does it affect? Is it a custom dropdown for certain fields? Where does it get its values from?

manavortex:
From a list of strings that are currently hand-assembled based on the cvm. I need to find a better way for it, but that's a different problem

WolvenKit/Views/Templates/RedTypeView.xaml
```diff
@@ -420,6 +420,12 @@
                     </DataTemplate>
                 </converters:RedEditorTemplateSelector.RedVectorColorEditor>
 
+                <converters:RedEditorTemplateSelector.FilterableDropdownEditor>
+                    <DataTemplate>
+                        <editors:FilterableDropdownCNameMenu Options="{Binding Data}" />
```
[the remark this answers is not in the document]
manavortex: The FilterableDropdownMenu (without CName) used to take the dictionary as a param. Now I was putting in the CVM directly. That might be the reason why it was behaving oddly O_O

WolvenKit/Views/Templates/RedCNameEditor.xaml.cs
```diff
@@ -161,6 +161,10 @@ public string Hash
             }
             set
             {
+                if (string.IsNullOrEmpty(value))
```
rfuzzo, line 164: This doesn't work, deleting the text in a CName editor doesn't set the Hash to 0

rfuzzo (changes requested):
The new CName dropdown editor works fine, but the unrelated changes to the REDCNameEditor (the hash 0) changes donb't work - please put those unrelated changes into a separate PR

WolvenKit/Views/Templates/RedCNameEditor.xaml.cs
rfuzzo: I have reverted it

rfuzzo (approved):

PR #2397, opened 2025-05-19
fix empty cname hash according to psi

manavortex:
# fix empty cname hash 

image


WolvenKit.RED4/Types/Primitives/Simples/CName.cs
```diff
@@ -33,7 +33,7 @@ public bool TryGetResolvedText([NotNullWhen(true)] out string? result)
     [Obsolete("Use GetRedHash instead")]
     public uint GetOldRedHash() => (uint)(_hash & 0xFFFFFFFF);
 
-    public static ulong CalculateHash(string resourcePath) => resourcePath == "None" ? 0 : FNV1A64HashAlgorithm.HashString(resourcePath);
```
rfuzzo, line 36: The argument name should be 'name' imo, not resourcePath

rfuzzo (changes requested):

PR #2404, opened 2025-06-02
inkatlas dialog: more improvements

manavortex:
# inkatlas dialog: more improvements

image
- Some UI improvements (inkatlas folder path doesn't look like two different fields anymore)
- everything will be forced to lowercase in the backend
- if you import a single png, the inkatlas part will always be named `icon`

### DIFF
--- a/WolvenKit.App/ViewModels/Documents/RedDocumentViewToolbarModel.cs
+++ b/WolvenKit.App/ViewModels/Documents/RedDocumentViewToolbarModel.cs
@@ -446,8 +446,7 @@ public partial class RedDocumentViewToolbarModel : ObservableObject
             && FilePath?.RelativePath(new DirectoryInfo(modDir)) is string activeFilePath
             && Path.GetDirectoryName(activeFilePath) is string dirName && !string.IsNullOrEmpty(dirName))
         {
-            // we're in a .mi file
-            if (!useTextureSubfolder)
+            if (!useTextureSubfolder || dirName.Contains("textures"))
             {
                 destFolder = dirName;
             }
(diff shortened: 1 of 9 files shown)


rfuzzo:
image
Is it intended that the selected folder is just one squashed foldername without slashes?

WolvenKit.App/ViewModels/Documents/RedDocumentViewToolbarModel.cs
```diff
@@ -460,7 +460,14 @@ public string GetTextureDirForDependencies(bool useTextureSubfolder)
             {
                 // if the folder contains more than one .mesh file, add a subdirectory inside "textures"
                 var fileName = Path.GetFileName(CurrentTab.FilePath.Split('.').FirstOrDefault() ?? "").ToFileName();
-                destFolder = Path.Combine(dirName, "textures", fileName);
+                if (!dirName.Contains("textures"))
```
rfuzzo, line 463: This got smuggled in from the other branch now

WolvenKit.Core/Extensions/StringExtensions.cs
```diff
@@ -136,16 +136,35 @@ public static bool IsNullOrEmptyOrEndsWith(this string? target, string value) =>
...
+        /// </summary>
+        public static string ToFilePath(this string target) => string.Join(Path.DirectorySeparatorChar,
+            target.Split(Path.DirectorySeparatorChar).Select(s => s.ToFileName()));
+
         /// <summary>
         /// Generates redengine friendly file name 
         /// </summary>
         public static string ToFileName(this string target) =>
-            new string(target.Where(c => !Path.GetInvalidFileNameChars().Contains(c)).ToArray()).Trim()
+            new string(target.Where(c => GetInvalidFileNameChars().Contains(c)).ToArray()).Trim()
```
rfuzzo, line 161: why check here for GetInvalidFileNameChars? `new string(target.Where(c => GetInvalidFileNameChars().Contains(c)).ToArray()).Trim()` I don't get it

rfuzzo (changes requested):

WolvenKit.App/ViewModels/Documents/RedDocumentViewToolbarModel.cs
manavortex: not smuggled - I decided to take it over, because it's now no longer arbitrary or hidden :)

WolvenKit.Core/Extensions/StringExtensions.cs
manavortex: This drops all characters that are not supposed to be in a file name from the string, e.g. backslashes or ascii farts. I've refactored the call to `Path` to get rid of the ReSharper warning.

PR #2408, opened 2025-06-07
add dependencies: removed double texture subdir from generated path

manavortex:
# add dependencies: removed double texture subdir from generated path


rfuzzo:
Why would it come to a double texture dir in the first place? This seems confusing.

manavortex:
If you already have a textures folder e.g.

rfuzzo:
Yeah but dirname is the path of the current File (relative), so it would be a special case just for files inside a folder called textures? This seems arbitrary

rfuzzo:
Doesn't seem like a good change tbh, it's arbitrary hidden logic.

PR #2409, opened 2025-06-09
Fixed importing meshes with ShapeKeys/MorphTargets

seberoth:
# Fixed importing meshes with ShapeKeys/MorphTargets

**Fixed:**
- Importing a glb with ShapeKeys now checks if it is named "GarmentSupport" before importing it as GarmentSupport


manavortex (approved):
LGTM, thank you Seb <3

Issue #2410, opened 2025-06-09
Mesh import option "ImportGarmentSupport" is misleading/wrong


seberoth:
Currently all it does is to update the mesh parameters (`meshMeshParamGarmentSupport` and `garmentMeshParamGarment`) if its set.

Whatever the option is set or not, if the glb contains a ShapeKey that will get imported as GarmentSupport into the blob.

The following problems occur if its NOT checked:
- if the old mesh contains GS and the glb hasn't a ShapeKey for it, the parameters won't get removed
- if the old mesh contains GS and the glb has a ShapeKey for it, the blob gets updated but the parameters won't get updated

My proposal would be:
If it's checked and oldMesh contains GS:
- Import ShapeKeys (with the name "GarmentSupport") in the blob
- Generate/Update parameters

If it's checked and oldMesh doesn't contains GS:
- Import ShapeKeys (with the name "GarmentSupport") in the blob
- Generate parameters

If it's NOT checked and oldMesh contains GS:
- Ignore ShapeKeys (with the name "GarmentSupport") in the blob
- Remove the parameters

If it's NOT checked and oldMesh doesn't contains GS:
- Ignore ShapeKeys (with the name "GarmentSupport") in the blob
- Don't update parameters

PR #2414, opened 2025-06-15
Refactor GS import code

seberoth:
# Refactor GS import code

**Fixed:**
- Implemented logic as in #2410


seberoth:
Would be cool if @Simarilius-uk and/or @islanddancer could test the artifact

PR #2415, opened 2025-06-17
drag and drop: abort without select

manavortex:
# drag and drop: abort without select

**Fixed:**
You can no longer drag and drop nodes that are not selected:
- Project tree will allow dragging, but won't handle any non-selected files
- RedTreeView will not allow dragging if any node is not selected

**Additional notes:**
Project Browser now has a shortcut for "Create Folder" (F7)

### DIFF
--- a/WolvenKit/Views/Tools/RedTreeView.xaml.cs
+++ b/WolvenKit/Views/Tools/RedTreeView.xaml.cs
@@ -482,8 +482,10 @@ namespace WolvenKit.Views.Tools
         
         private bool IsAllowDrop(TreeViewItemDragOverEventArgs e)
         {
-            if (e.DraggingNodes?[0].Content is not ChunkViewModel { IsSelected: true } source ||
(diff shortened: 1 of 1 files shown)


notaspirit:
the shortcut doesn't seem to work (pressing F7 does nothing) but it does show up in next to the button in the right click menu and creating a folder from there works fine.
Other than that looks good to me and drag and drop works as expected

notaspirit:
Shortcut works now after having clicked in the file explorer, however if you never click anything after opening a project or have clicked a different panel e.g. logs panel the shortcut silently does nothing which could unintuitive or confusing.

manavortex:
unfortunately it needs _some_ kind of selection to work. I think that's okay, since the shortcut should "feel" as if it exposed the context menu item?

notaspirit:
if others behave the same then it's fine

notaspirit (approved):

PR #2417, opened 2025-06-17
hot reload: user feedback, handling of already-installed archive

manavortex:
# hot reload: user feedback, handling of already-installed archive

Hot Reload install button will now try deleting the .archive from `archive/pc/mod` and print a warning if it fails.

### DIFF
--- a/WolvenKit.Modkit/RED4/ModTools.Helper.cs
+++ b/WolvenKit.Modkit/RED4/ModTools.Helper.cs
@@ -118,29 +118,42 @@ public partial class ModTools
         {
             // Get the relative path of the file with respect to the packedDirectory
             var relativePath = file.FullName[(packedDirectory.FullName.Length + 1)..];
-
+            var destPath = Path.Combine(gameRootDir.FullName, relativePath);
+            
             if (file.FullName.EndsWith(".archive") && installToHot)
             {
+                // try to delete already-installed archive from mod folder. This might fail. 
+                if (File.Exists(destPath))
+                {
+                    try
+                    {
+                        File.Delete(destPath);
+                    }
+                    catch
+                    {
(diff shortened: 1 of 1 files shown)


notaspirit:
plus if hot reloading were to fail for whatever reason after the archive file ends up in the correct directory (`Cyberpunk 2077\archive\pc\hot`) that is the responsibility of redhottools not wolvenkit

manavortex:
The problem here is the following workflow:
- hot reload (everything works)
- close the game for whatever reason
- keep modding
- hot reload install
- start game
- it doesn't work because the archive was in archive/pc/mod and the game loaded it via default

... and the user is _not_ the only person who ran into this. It's part of the reason why I'm not using HotReload.

notaspirit:
So the issue is that when you hot reload while the game is not running the changes don't get applied? in that case this pr doesn't change anything about this behavior. 
To fix it you would need to determine whether the game is running or not when you hot install and adjust accordingly at which point you might as well merge the install and hot install buttons to have one that 

1. checks if the target files are used by the game 
2. if they are hot install if they are not install regularly

notaspirit:
After psi has fixed hot reloading some months ago I wasn't able to recreate the issue of archives not reloading, and even if it were to fail the logic in the pr makes little sense as it always fails to delete the archive assuming the game is running. So you may as well skip trying to delete it and always log the message, but again I have not been able to recreate the issue of hot reloading not reloading after it has been fixed from rht's side. If hot reloading failing is tied to certain resource types you could have the message be tied to that instead.

psiberx:
> * it doesn't work because the archive was in archive/pc/mod and the game loaded it via default

There's absolutely no difference between archive already present on game launch and archive hot loaded after launch. If there was some issue, it couldn't be that for sure.

PR #2418, opened 2025-06-17
feat: Rename scnAnimName unks

misterchedda:
# Rename scnAnimName unks

This change adds labels for the unk1 and unk2 properties within the `scnAnimName` class (The serialization/node wrapper logic is unchanged), we already knew the type and the usage logic in the node wrapper was added by @JakubMarecek last year. This change just adds the labels to the fields:

- Renamed `scnAnimName.unk1` to `AnimNames` (CArray<CName>)
- Renamed `scnAnimName.unk2` to `ReferenceIndices` (CArray<CUInt16>)

Also fixes a bug in the GraphEditor which will now show all animation types (direct, reference, container, etc.)

### DIFF
--- a/WolvenKit.App/ViewModels/GraphEditor/Nodes/Scene/scnSectionNodeWrapper.cs
+++ b/WolvenKit.App/ViewModels/GraphEditor/Nodes/Scene/scnSectionNodeWrapper.cs
@@ -36,25 +36,42 @@ public class scnSectionNodeWrapper : BaseSceneViewModel<scnSectionNode>
                         {
                             animNameObj = handle.GetValue() as scnAnimName;
                         }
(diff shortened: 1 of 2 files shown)


WolvenKit.App/ViewModels/GraphEditor/Nodes/Scene/scnSectionNodeWrapper.cs
```diff
@@ -36,25 +36,42 @@ public scnSectionNodeWrapper(scnSectionNode scnSectionNode, scnSceneResource scn
                         {
                             animNameObj = handle.GetValue() as scnAnimName;
                         }
-                        if (animNameObj != null && animNameObj.Unk1 != null && animNameObj.Unk1.Count > 0)
+                        
+                        if (animNameObj != null)
```
manavortex, line 40: This looks like an amazing opportunity for an early return...?

PR #2425, opened 2025-06-19
fix "hides inherited member" warnings

manavortex:
# fix "hides inherited member" warnings in classesExt


rfuzzo:
Should be fine, but let's compare json convert outputs if this breaks serialisation somewhere

manavortex:
I checked this for the first type, and it serializes fine. Want me to test the others as well, or do we trust that?

rfuzzo:
Nope, that's good enough :) it shouldn't matter anyway, I'm just being overly cautious

manavortex:
@rfuzzo since you already reviewed this, how about merging it? :D

PR #2435, opened 2025-06-27
fix: "replace in project" exception

manavortex:
# fix: "replace in project" exception

**Fixed:**
- An exception would be thrown if "replace in project" would find any `*.*.json` files under `raw`. This will now be handled gracefully.
- Renaming mesh files will also update paths in resource files (e.g. `*.archive.xl` patch files)

### DIFF
--- a/WolvenKit.App/Helpers/ProjectResourceTools.cs
+++ b/WolvenKit.App/Helpers/ProjectResourceTools.cs
@@ -60,7 +60,7 @@ public class ProjectResourceTools
         ".yaml",
         ".yml",
         ".reds",
-        "tweak"
+        ".tweak"
     ];
 
     /// <summary>
--- a/WolvenKit.Core/Extensions/StringPathExtensions.cs
+++ b/WolvenKit.Core/Extensions/StringPathExtensions.cs
@@ -76,7 +76,7 @@ namespace WolvenKit.Interfaces.Extensions
         /// Generates redengine friendly file name 
         /// </summary>
         public static string ToFileName(this string target) =>
-            new string(target.Where(c => Path.GetInvalidFileNameChars().Contains(c)).ToArray()).Trim()
+            new string(target.Where(c => !Path.GetInvalidFileNameChars().Contains(c)).ToArray()).Trim()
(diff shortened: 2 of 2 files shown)


WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -38,12 +39,29 @@ public class ProjectResourceTools
...
-    private static readonly List<string> s_resourceFileExtensions = [".lua", ".xl", ".json", ".yaml", ".yml", ".reds"];
+    private static readonly List<string> s_resourceFileExtensions =
+    [
+        ".lua",
+        ".xl",
+        ".json",
+        ".yaml",
+        ".yml",
+        ".reds",
+        "tweak"
```
misterchedda, line 63: should be .tweak I think?

WolvenKit.Core/Extensions/StringPathExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// Generates redengine friendly file path. 
+        /// </summary>
+        public static string ToFilePath(this string target) => string.Join(Path.DirectorySeparatorChar,
+            target.Split(Path.DirectorySeparatorChar).Select(s => s.ToFileName()));
+
+        /// <summary>
+        /// Generates redengine friendly file name 
+        /// </summary>
+        public static string ToFileName(this string target) =>
+            new string(target.Where(c => Path.GetInvalidFileNameChars().Contains(c)).ToArray()).Trim()
```
misterchedda, line 79: missing a ! 

should be `!Path.GetInvalidFileNameChars().Contains(c)`

misterchedda (changes requested):

WolvenKit.Core/Extensions/StringPathExtensions.cs
manavortex: Thanks for catching that one! Pretty sure I fixed it on another branch already, but you know how it goes >.<

misterchedda (approved):

PR #2438, opened 2025-06-28
supporting changes for file validation (xl and yaml)

manavortex:
# supporting changes for file validation (xl and yaml)

File validation will now verify file paths in xl files. Since it can also parse yaml files, perhaps I'll implement that as well!

### DIFF
--- a/WolvenKit.Modkit/Scripting/ScriptFunctions.cs
+++ b/WolvenKit.Modkit/Scripting/ScriptFunctions.cs
@@ -192,10 +192,20 @@ public partial class ScriptFunctions
         // remove yaml tags like !include, !append etc.
         yamlText = YamlTagRegex().Replace(yamlText, "");
 
-        // deserialize it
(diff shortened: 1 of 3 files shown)


WolvenKit.Modkit/Scripting/ScriptFunctions.cs
```diff
@@ -188,8 +189,11 @@ public virtual IEnumerable GetArchiveFiles()
     /// <returns>The converted JSON string</returns>
     public virtual string YamlToJson(string yamlText)
     {
-        var deserializer = new Deserializer();
-        var yamlObject = deserializer.Deserialize(yamlText);
+        // remove yaml tags like !include, !append etc.
+        yamlText = YamlTagRegex().Replace(yamlText, "");
```
seberoth, line 193: Why do you strip the tags here?

manavortex: because the yaml doesn't compile while they're present :|

PR #2444, opened 2025-06-28
feat: add scene editor

misterchedda:
# Scene Editor

image

## Features

*   **Integrated `SceneGraphView` layout with dynamic property panel**
    The editor is built around the new **`SceneGraphView`**, a central workspace with a visual graph editor on the right and a multi-tabbed property panel on the left. This unified view is powered by **`SceneGraphViewModel`**
    *   The property panel's tabs are dynamically generated by **`CollectionViewHelper`**, which uses filters to organize the raw scene data into more useful groups like "Actors & Props," "Logic & Flow," "Dialogue", and more. 
    *   "Node Properties" tab for editing the selected node's property is powered by a new, cached `RedTypeToChunkViewModelCollectionConverter`. This dynamically builds the property tree for the currently selected node.

*   **Realtime UI sync**
    The graph editor and the Node Properties panel are decoupled but remain in sync
    *   Communication is mediated by services like **`NodeSelectionService`** (for tracking the selected node) and **`NodePropertyUpdateService`** (for broadcasting data changes).
    *   This is powered by the **`NodeViewModel`**, which now subscribes to `PropertyChanged` events on its underlying data model. Both property updates from the panel -> graph and socket/connection updates from the graph -> panel update in-realtime. 

https://github.com/user-attachments/assets/0f599d09-0927-43bb-a0cf-e390fb21c977

_Node-driven workflow: select a node to edit its properties with realtime property sync_

*   **Improved node visualization**
    *   Nodes are now **color-coded** based on their type (`scnQuestNode` -> `questNode` types also have unique colors, and not just scene node types)
    *   The UI also displays titles and notable points more prominently, and other details directly on the node
    *   Node connections now have some basic progressive path highlighting with variable thickness. Paths leading to "dead ends" (like connections leading to a single workspot node or a couple of journal nodes) are rendered with a thinner line, while paths that continue the main flow are thicker to hint on the scene's progression

*   **Smart node details**
    The `NodeProperties` helper has been refactored to use reflection for **auto discovery of useful properties**. This supplements existing curated details and ensures that nearly all nodes will now display useful information automatically.

image

*   **QoL improvements**
    The editor now supports more actions via the right-click context menu and keyboard shortcuts:
    *   **Duplicate Node** (`Ctrl+D`): Creates a copy of the selected node
    *   **Delete Node** (`Delete` key): A "soft delete" that replaces the node with a `DeletionMarkerNode` to preserve all incoming and outgoing connections
    *   **Destroy Node** (`Shift+Delete`): A "hard delete" that permanently removes the node and its connections
    *   **Detach Node**: Severs all connections to and from the selected node
    *   **New Node**: (`Ctr+N`) shortcut to trigger the new node dialog
    *   **Arrow-key Navigation**: Traverse the graph spatially between nodes
    *   **Right-click a connection** to delete it directly
    *   Right-click Section node to add output sockets
    *  Adds a grid background to the graph editor

image

*   **Data aware editing**
    *   **Contextual dropdowns**: The `CVMDropdownHelper` has been extended to provide human-readable dropdowns for selecting actors, performers, props, dialogue, and animation resources, replacing raw integer IDs
    *   **Automatic ID/Index assignment**: A new `SceneEditingHelper` handles automatic ID and index assignment when creating new actors, performers, props, and dialogue lines.

image

Also includes a fix (commit `73d159d`) for a longstanding bug in scene nodes:
- Scene nodes in the Graph View were missing the standard "CutDestination" input socket (at Name: 1, Ordinal: 0), which is common in quest nodes. Scene nodes will now correctly display this socket.
- I ran a wscript to see how many connections were affected. This change will "unhide" ~9,000 previously ignored connections to CutDestination sockets on scene nodes across existing .scene files
- Thanks to Deceptious for reporting this issue

I've tested most of the core functionality decently I think, and this could use a lot more testing! @MrBilL61 has kindly offered to help, and anyone else is also free to pitch in 😄  

Some more 'big picture' things that I plan to do in the near future in hopefully smaller, separate PRs:
- Extend to .questphase files (edit: see #2459 )
- Search
- Timeline view for Section nodes
- Smaller tasks: ~~Improve the 'New Node' right-click dialog with better categories~~, expose an easy way to change node type, etc


MrBilL61:
I did quick testing, here are few things I've noticed:
1.  `fileEntryIndex` fields in questJournalPath are not editable
2. Right-clicking on connection doesn't seem to work
3. All nodes in questphases are now displaying notable points
4. Would be nice to display `isPlayer` field for any node if the entity or actor reffered is player

I'll do more testing tomorrow.

MrBilL61:
Both fixes worked

WolvenKit.App/ViewModels/Documents/RedDocumentViewModel.cs
```diff
@@ -211,24 +214,40 @@ public bool SaveSync(object? _, string? filePath = null)
...
+
         if (!File.Exists(FilePath) || (!force && IsDirty))
         {
             return false;
         }
 
+        // If we got here, it means the files were different, or an error occurred during comparison.
+        // Proceed with the original, destructive reload.
         using var fs = File.Open(FilePath, FileMode.Open);
         if (!_parserService.TryReadRed4File(fs, out var cr2WFile))
```
manavortex, line 252: You could use my cr2wtools - it takes a path and gives you either the cr2w or an exception

manavortex (approved):

WolvenKit.App/ViewModels/Documents/RedDocumentViewModel.cs
misterchedda: Nice thanks! also fixed

misterchedda:
Pending more bug reports and testing, the only big thing left in this PR is socket creation + cleanup logic for some scene nodes:

**Confirm name-ordinal input and output socket utility by testing in-game for**

- [ ] Choice: we have this mostly mapped out already, but apparently base game connects to Input sockets N: 2, O: 0 (ie 2,0 ) a lot (378 instances) and N:3, O:0 (3,0) also once
- [ ] Hub: has multiple out sockets from 0,0 to 0,6. Might just be for categorising and not have any specific meaning attached 
- [ ] Xor and And: also have out socks from 0,0 to 0,6... might be similar to Hub. Also confirm if Cut (assumed at 1,0) is supposed to work (there are ~30 instances in base game and if there's supposed to be a dedicated Cut out/OnCut socket)
- [ ] Investigate Flow Control N & O: havent run any data scripts on this yet so no idea about its sockets yet
- [ ] Account for all the weird in socket N/O combinations in base game files for eg Deletion Marker has a bunch of sockets at 666,0. 1026,0 is also used in many nodes incl And. Currently the node will render all sockets from 0 to 666/1026 if a node has a connection to them resulting in awkardly long node heights (might have to change the logic around input sock creation)

seberoth:
Not sure if it helps (for e.g. the 666 stuff) but: https://github.com/WolvenKit/WolvenKit/compare/main...refactor/scene_graph_connectors

misterchedda:
wow - that helps **a lot** 😄  @seberoth 

Totally forgot about rewindable section too and you have mapped flow control completely as well so that helps a ton. So 2,0 and 3,0 on choice is for reactivation and time limit finished, I see. I also noticed you use "Cancel" instead of "Cut", do you happen to have a preference? In my PR I was trying to use the CutDestination/Cut terms since we also use it in quest nodes 

I might just try cherry picking that commit in itself (but i fear ive changed some more things in between)

PR #2447, opened 2025-07-02
Clear Chunk Materials in Selection

manavortex:
# Clear Chunk Materials

if multiple appearances are selected, this command will clear the chunk materials in each of them. Requires the selection sync to be fixed, though.


misterchedda (approved):
tested this earlier as part of #2460, seems to work fine

PR #2455, opened 2025-07-05
feature/tweakxl support

manavortex:
# feature/tweakxl support

Originally @jackhumbert 's merge request - back merge to main, please check if this makes things better or worse (I think better?)


poirierlouis:
I don't know much of TweakXL, do you know what is the use of `_inline` and `!append`?

Do we actually need to parse TweakXL files? Like is there a use for them in WKit already?

PR #2466, opened 2025-07-09
Fixing error that prevents scene file creation

manavortex:
# Fixing error that prevents scene file creation


WolvenKit.RED4/Types/ClassesExt/scnSceneResource.cs
```diff
@@ -2,6 +2,11 @@ namespace WolvenKit.RED4.Types;
 
 public partial class scnSceneResource
 {
+    partial void PostConstruct()
```
manavortex, line 5: @rfuzzo is this a problem for the CLI?

PR #2471, opened 2025-07-12
feature: project color configurable

manavortex:
# feature: project color configurable

Project color is now configurable.

### DIFF
--- a/WolvenKit.App/Helpers/ColorHelper.cs
+++ b/WolvenKit.App/Helpers/ColorHelper.cs
@@ -1,0 +1,76 @@
+﻿using System;
+using System.Text.RegularExpressions;
+using System.Windows.Media;
+using WolvenKit.Core.Exceptions;
+
+namespace WolvenKit.App.Helpers;
+
+public partial class ColorHelper
+{
+    private static readonly Random s_rnd = new();
+
+    /// <returns>A <see cref="Color"/> with a random color value.</returns>
+    public static Color GetRandomColor()
+    {
+        return Color.FromRgb((byte)s_rnd.Next(0, 256), (byte)s_rnd.Next(0, 256),
+            (byte)s_rnd.Next(0, 256));
+    }
+
+    /// <param name="clampBrightness">Color must have a brightness value of at least x</param>
+    /// <returns>A <see cref="Color"/> with a random color value and a brightness >= the parameter value.</returns>
+    public static Color GetRandomColor(double clampBrightness)
+    {
+        var color = GetRandomColor();
+
+        while (System.Drawing.Color.FromArgb(color.A, color.R, color.G, color.B).GetBrightness() < clampBrightness)
+        {
+            color = GetRandomColor();
+        }
+
+        return color;
+    }
+
+    // https://stackoverflow.com/a/1636354/5825537 - this will capture ARGB and RGB hex colors
+    [GeneratedRegex("^#(?:[0-9a-fA-F]{3,4}){1,2}$")]
+    private static partial Regex s_hexColorRegexARGB();
+
+    /// <summary>
+    /// Will create a color from a hex code string.
+    /// </summary>
+    /// <param name="hex">A hex string. Must start with # and contain either 3, 4, 6 or 8 characters.</param>
+    /// <returns>A <see cref="Color"/> object</returns>
+    /// <exception cref="ArgumentException">if the string can't be parsed.
+    /// For an exception-safe fallback, please use <see cref="CreateFromHexOrRandom"/></exception>
+    public static Color CreateFromHex(string hex)
+    {
+        ArgumentNullException.ThrowIfNull(hex);
+
+        if (!s_hexColorRegexARGB().IsMatch(hex))
+        {
+            throw new ArgumentException(
+                $"String {hex} is not a valid hex color code. Must start with # and be 3, 4, 6 or 8 characters long.");
+        }
+
+        if (ColorConverter.ConvertFromString(hex) is Color color)
+        {
+            return color;
+        }
+
+        throw new WolvenKitException(0x5002, $"String {hex} could not be parsed to a color.");
+    }
+
+    /// <summary>
+    /// Will create a color if the parameter is a valid hex code, otherwise return null.
+    /// </summary>
+    public static Color? CreateFromHexOrNull(string? hex)
+    {
+        if (IsHexColorCode(hex))
+        {
+            return CreateFromHex(hex!);
+        }
+
+        return null;
+    }
+
+    public static bool IsHexColorCode(string? s) => s is not null && s_hexColorRegexARGB().IsMatch(s);
+}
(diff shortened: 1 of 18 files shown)


manavortex:
@misterchedda, could I bug you again? I haven't been able to get the "recent projects" screen to update after changing the project colour. Everything else works.

WolvenKit/Views/Others/ColorPickerButton.xaml.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    public Color Color
+    {
+        get => (Color)GetValue(ColorProperty);
+        set => SetValue(ColorProperty, value);
+    }
+
+    private void ColorPickerControl_OnChange(DependencyObject d, DependencyPropertyChangedEventArgs e)
+    {
+        var oldValue = Color;
```
poirierlouis, line 48: Is it not useless/unused as-is?

WolvenKit/Views/Others/ColorPickerButton.xaml
```diff
@@ -0,0 +1,23 @@
...
+    syncfusionskin:SfSkinManager.Theme="{syncfusionskin:SkinManagerExtension ThemeName=MaterialDark}">
+    <Grid>
+        <syncfusion:ColorPicker
+            Name="ColorPickerControl"
+            Width="40"
+            VerticalAlignment="Bottom"
+            Background="{StaticResource BackgroundColor_Control}"
+            BorderBrush="{DynamicResource BorderBrush}"
+            BorderThickness="1"
+            syncfusionskin:SfSkinManager.Theme="{syncfusionskin:SkinManagerExtension ThemeName=MaterialDark}"
```
poirierlouis, line 17: Is this `SfSkinManager.Theme` required?

WolvenKit/Views/Dialogs/ProjectWizardView.xaml
```diff
@@ -66,6 +66,16 @@
             TextChanged="ProjectNameTextBox_OnTextChanged"
             Validation.ErrorTemplate="{StaticResource ValidationErrorTemplate}" />
 
+        <controls:ColorPickerButton
+            x:Name="ProjectColorButton"
+            Grid.Row="1"
+            Grid.Column="1"
+            Width="40"
+            Height="22"
```
poirierlouis, line 74: Maybe this should be using `DynamicResource` to scale UI?

WolvenKit.App/Services/ProjectManager.cs
```diff
@@ -25,14 +25,16 @@ public partial class ProjectManager : ObservableObject, IProjectManager
     private readonly IHashService _hashService;
     private readonly IArchiveManager _archiveManager;
     private readonly ISettingsManager _settingsManager;
+    private readonly IRecentlyUsedItemsService _itemsSvc;
```
poirierlouis, line 28: Please, use a more explicit name like everywhere else.
e.g. `recentlyUsedItemsService`, it is verbose but it tracks with the naming convention already in-place.

For example, try to read `itemsSvc` in code where you don't have context of the type declaration, in a git diff. It would be impossible to understand what it is about precisely.

WolvenKit.App/Models/ProjectManagement/Project/Cp77Project.cs
```diff
@@ -22,15 +23,20 @@ public sealed partial class Cp77Project : IEquatable<Cp77Project>, ICloneable
...
+        Color? color = null)
     {
         Location = location;
         Name = name;
         ModName = modName;
         OpenProjectFiles = openProjectFiles;
+        if (color is Color c)
+        {
+            ProjectColor = c;
+        }
```
poirierlouis, line 36: Seems overcomplicated imo, is it correct to just do something like?
```cs
if (color != null)
{
  ProjectColor = c;
}
```

WolvenKit.App/Helpers/ColorHelper.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System;
+using System.Text.RegularExpressions;
+using System.Windows.Media;
+using WolvenKit.Core.Exceptions;
+
+namespace WolvenKit.App.Helpers;
+
+public partial class ColorHelper
```
poirierlouis, line 8: I wonder, don't we already have some color functions spread out as helpers? Would be nice to uniformize / put everything in this ColorHelper I guess.

poirierlouis:
Alignment is off regarding other rows, it also needs margin with color picker to be coherent (like between rows):
image

I could reproduce the issue. My first thought is that new color is not notified in the WelcomePageView.

poirierlouis:
What did you try?

One comes to mind, which is to update lists of pinned/recent projects. As I assume view-side listen for changes of the list itself, not for child's property changes.

Also, after pulling/building this branch, I'm having an error with Cp77Project test, not compiling. Any idea?

manavortex:
Still not refreshing, but the cache now updates at least

WolvenKit.App/Models/ProjectManagement/Project/Cp77Project.cs
poirierlouis: Oh, C# can't analyze past the conditional expression...
Though is not enough to use?
```cs
if (color is not null)
{
    ProjectColor = color!;
}
```

manavortex: I could, yeah, but I don't like ! (got burned in Typescript)

manavortex: You can use a ! independent from null checks - just declare "this here exists, I promise!", so it's a red flag for me (I don't trust meatbrains - hell, I barely even trust compilers :D)

manavortex:
Because that led to the property not being written back to the itemModel :thisisfine:

manavortex:
On `feature/project-color-configurable`, it didn't

poirierlouis:
Back to the issue, WelcomePageView is not using the same ViewModel.

This is where the list for pinned/recent projects is defined:
https://github.com/WolvenKit/WolvenKit/blob/b7c308496045ca2d583f4f0c704758e6b97039ac/WolvenKit.App/ViewModels/HomePage/Pages/WelcomePageViewModel.cs#L261

`FancyProjectObject` is the type used to represent a project in the view:
https://github.com/WolvenKit/WolvenKit/blob/b7c308496045ca2d583f4f0c704758e6b97039ac/WolvenKit.App/ViewModels/HomePage/Pages/WelcomePageViewModel.cs#L87

On the opposite, `ProjectSettingsDialog` is using a `Cp77Project` in its ViewModel.

It's why the change is not reflected. You'll need to refactor this part, to I guess, directly reuse `Cp77Project` instance with `FancyProjectObject` to correctly propagate changes.

poirierlouis:
If you can point me the binding/connection between the two, I may have missed it...

PR #2474, opened 2025-07-13
fixed missing 'copy to clipboard' icon in logview

manavortex:
# fixed missing 'copy to clipboard' icon in logview

image

- Added comments with icon gallery urls in .xaml.xl file 
- Updated spellcheck dictionary with a few things that are not typos


manavortex:
These are IDE settings, I always commit them as I encounter them. it's just for Rider's spell check :)

PR #2476, opened 2025-07-14
Log file viewer: Open most recent file

manavortex:
# Log file viewer: Open most recent file

... on shift-click on "open log folder" button.

- Removed auto-open from file validation
- made "open log folder" button yellow

### DIFF
--- a/WolvenKit.App/ViewModels/Shell/AppViewModel.WScript.cs
+++ b/WolvenKit.App/ViewModels/Shell/AppViewModel.WScript.cs
@@ -123,24 +123,10 @@ public partial class AppViewModel : ObservableObject /*, IAppViewModel*/
             await _scriptService.ExecuteAsync(code);
         }
 
-        // This should never happen, but better safe than sorry
-        if (FileHelper.GetMostRecentlyChangedFile(Path.Combine(ISettingsManager.GetAppData(), "Logs"), "*.txt") is not FileInfo fI)
-        {
-            _loggerService.Info("Done.");
-            return;
-        }
-
-        _loggerService.Info($"Done. The most recent log file is {fI.FullName}.");
-
-        try
-        {
-            Process.Start(new ProcessStartInfo(fI.FullName) { UseShellExecute = true });
-        }
-        catch (Exception ex)
-        {
-            _loggerService.Error($"Failed to open log file: {ex.Message}");
-        }
-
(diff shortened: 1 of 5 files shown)


WolvenKit/Views/Tools/LogView.xaml.cs
```diff
@@ -72,24 +74,18 @@ public LogView()
...
-                    default:
-                        return true;
-                }
+                    Logtype.Error => ViewModel?.FilterByLevel[0] == true,
+                    Logtype.Warning => ViewModel?.FilterByLevel[1] == true,
+                    Logtype.Success => ViewModel?.FilterByLevel[2] == true,
+                    Logtype.Normal or Logtype.Important => ViewModel?.FilterByLevel[3] == true,
+                    Logtype.Debug => ViewModel?.FilterByLevel[4] == true,
+                    _ => true
+                };
```
poirierlouis, line 88: I think `== true` is redundant, `FilterByLevel[x]` already returns a boolean:
https://github.com/WolvenKit/WolvenKit/blob/1481f03c3f6f9eeeb389ff7b140e184e54eb4301/WolvenKit.App/ViewModels/Tools/LogViewModel.cs#L43

WolvenKit.App/ViewModels/Shell/AppViewModel.WScript.cs
```diff
@@ -123,24 +123,10 @@ private async Task RunFileValidationOnProject()
...
-        }
-        catch (Exception ex)
-        {
-            _loggerService.Error($"Failed to open log file: {ex.Message}");
-        }
-
+        _loggerService.Info("Done.");
+        _loggerService.Info(
+            "To open the most recent log file, shift-click on the 'open folder' button to the right.");
+
```
poirierlouis, line 129: Could the UI be updated too, to show a tooltip when hovering the button? Something like:
`Hold SHIFT to open the most recent log file`

poirierlouis (changes requested):
Two points, otherwise it looks good :)

WolvenKit/Views/Tools/LogView.xaml.cs
manavortex: View model could theoretically be null at this point, so the result of the call could be null

poirierlouis (approved):
LGTM

PR #2287, opened 2025-02-24
Replace the logo with a custom-made one

platyple:
# Replace the logo with a custom-made one

**Implemented:**
- replaces the splash screen and the white banner on the welcome page with a new logo i made from scratch. the splash screen is now transparent. pretty sleek if you ask me

**Fixed:**
- The program is no longer called wocvenkit

**Additional notes:**
i put the original svg in `WolvenKit/Resources/Media/Images/Application`, if you want to change it or want to reuse the letters for something else. it's GPL-licensed, after all...
oh, and i've never used github for more than downloading projects before this, so no idea how any of this works lmao

PR #2477, opened 2025-07-14
added yellow background back to splash screen

manavortex:
# added yellow background back to splash screen

Before:
SplashScreen1

After:
SplashScreen1


poirierlouis (approved):
Back and forth with #2287 ^^
LGTM

PR #2480, opened 2025-07-16
ChunkViewModel: exception proofing

manavortex:
# ChunkViewModel: exception proofing

After chedda's synchronization changes, the CVM doesn't like pasting much (it complains about index errors). 

This PR resolves the issue by introducing a slight delay before the search is re-applied, and by not suspending the collection refresh for certain operations.


misterchedda (approved):
thanks!

PR #2481, opened 2025-07-17
fix: unknown enum value

manavortex:
# fix: unknown enum value

Fix for this. Not entirely sure where it comes from, because during the objects' construction, all seems well.

This could also be solved by removing the numbers from the enum definition (then 0 will have a valid value), but I don't know if that has any side effects. 

image

### DIFF
--- a/WolvenKit.Common/RED4/CommonFunctions.cs
+++ b/WolvenKit.Common/RED4/CommonFunctions.cs
@@ -46,7 +46,7 @@ public static class CommonFunctions
     public static XbmImportArgs TextureSetupFromTextureGroup(GpuWrapApieTextureGroup textureGroup)
     {
         var outSetup = new XbmImportArgs(textureGroup);
-        
+
 
         switch (textureGroup)
         {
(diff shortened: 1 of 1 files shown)


WolvenKit.App/ViewModels/Tools/ImportableItemViewModel.cs
```diff
@@ -153,12 +153,18 @@ public static bool TryLoadXbmSettingsFromGame(string fileName, IArchiveManager a
...
             bitmapTexture.RenderTextureResource.RenderResourceBlobPC.Chunk is not rendRenderTextureBlobPC blob)
         {
             return false;
         }
 
+        // Invalid enum value. Not sure where this is coming from, in the constructors everything is assigned correctly.
+        if (setup.Group.ToString() == "0")
+        {
+            setup.Group = Enums.GpuWrapApieTextureGroup.TEXG_Generic_Color;
+        }
```
seberoth, line 166: That looks wrong. Do you have an example file?
Or a way to reproduce this?

Not sure why it is in 2 PRs...

seberoth: Just tested it with ~6000 random files (including the mentioned ones)
Not a single throw that issue for me (with your fix removed from dev)

PR #2494, opened 2025-07-27
Feature in app updater

notaspirit:
# Feature in app updater

**Implemented:**
- in app updater to replace previous version check + manual install
- additional setting to still allow checking for updates on startup with the user having to agree to update

### DIFF
--- a/WolvenKit.App/Services/ISettingsDto.cs
+++ b/WolvenKit.App/Services/ISettingsDto.cs
@@ -56,7 +56,7 @@ public interface ISettingsDto
 
     #region General
 
-    public bool SkipUpdateCheck { get; set; }
+    public bool AutoUpdateOnStartup { get; set; }
     public bool AlwaysAskBeforeUpdating { get; set; }
     public EUpdateChannel UpdateChannel { get; set; }
     public bool ShowGuidedTour { get; set; }
--- a/WolvenKit.App/Services/SettingsDto.cs
+++ b/WolvenKit.App/Services/SettingsDto.cs
@@ -64,7 +64,7 @@ public class SettingsDto : ISettingsDto
         TreeViewIgnoredExtensions = settings.TreeViewIgnoredExtensions;
 
         // General
-        SkipUpdateCheck = settings.SkipUpdateCheck;
+        AutoUpdateOnStartup = settings.AutoUpdateOnStartup;
         AlwaysAskBeforeUpdating = settings.AlwaysAskBeforeUpdating;       
         UpdateChannel = settings.UpdateChannel;
         ShowGuidedTour = settings.ShowGuidedTour;
@@ -136,7 +136,7 @@ public class SettingsDto : ISettingsDto
 
     #region General
 
-    public bool SkipUpdateCheck { get; set; }
+    public bool AutoUpdateOnStartup { get; set; } = true;
     public bool AlwaysAskBeforeUpdating { get; set; }
     public EUpdateChannel UpdateChannel { get; set; }
     public bool ShowGuidedTour { get; set; }
@@ -207,7 +207,7 @@ public class SettingsDto : ISettingsDto
         settingsManager.TreeViewIgnoredExtensions = TreeViewIgnoredExtensions;
 
         // General
-        settingsManager.SkipUpdateCheck = SkipUpdateCheck;
+        settingsManager.AutoUpdateOnStartup = AutoUpdateOnStartup;
         settingsManager.AlwaysAskBeforeUpdating = AlwaysAskBeforeUpdating;      
         settingsManager.UpdateChannel = UpdateChannel;
         settingsManager.ShowGuidedTour = ShowGuidedTour;
@@ -272,7 +272,7 @@ public class SettingsDto : ISettingsDto
 
             SettingsVersion = 4;
         }
-
+        
         SettingsVersion = s_currentSettingsVersion;
 
         IsDirty = true;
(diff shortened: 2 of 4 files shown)


WolvenKit.App/Services/ISettingsManager.cs
```diff
@@ -80,6 +80,17 @@ public static string GetWorkDir()
         return dir;
     }
 
+    public static string GetTemp_DownloadsPath()
```
dragonzkiller, line 83: You should use `System.IO.Path.GetTempPath()` for doing temp things.
1. That's why it's there
2. So if something happens during the update, they can always clear any junk left behind in the Windows Disk Cleanup

WolvenKit.App/Services/IUpdateService.cs
```diff
@@ -0,0 +1,10 @@
+using System.Threading.Tasks;
+
+namespace WolvenKit.App.Services;
+
+public interface IUpdateService
+{
+    public Task<bool> IsUpdateAvailable();
```
dragonzkiller, line 7: We had similar functions in `AppViewModel` on wkit startup, is this what you're removing from there and reimplementing as a service here?

WolvenKit.App/Services/UpdateService.cs
```diff
@@ -0,0 +1,151 @@
...
+            _loggerService.Error(ex);
+            return;
+        }
+        
+        var unzipPath = Path.Join(ISettingsManager.GetTemp_DownloadsPath(), "unzip");
+        Directory.CreateDirectory(unzipPath);
+        ZipFile.ExtractToDirectory(downloadZipPath, unzipPath);
+        
+        var exePath = Environment.ProcessPath ?? Path.Combine(AppContext.BaseDirectory, "WolvenKit.exe");
+        var scriptPath = Path.Combine(ISettingsManager.GetTemp_DownloadsPath(), "update.ps1");
```
dragonzkiller, line 68: Why do we need both a Powershell and VBS to perform the update? I understand offloading the download and update to a separate process so you can access the locked files, but why both?

WolvenKit.App/Services/UpdateService.cs
```diff
@@ -0,0 +1,151 @@
...
+Start-Process -FilePath $exePath
+
+Remove-Item -Path $rootTempPath -Recurse -Force -ErrorAction SilentlyContinue
+
+");
+        
+        File.WriteAllText(vbsScriptPath, $@"Set objShell = CreateObject(""WScript.Shell"")
+objShell.Run ""powershell.exe -ExecutionPolicy Bypass -File """"{scriptPath}"""""", 0, False
+");
+        Process.Start(new ProcessStartInfo
```
dragonzkiller, line 90: Have you tested this on systems where the Execution Policy for Powershell scripts was at a higher security level?

WolvenKit.App/Services/UpdateService.cs
```diff
@@ -0,0 +1,151 @@
...
+        catch (HttpRequestException ex)
+        {
+            _loggerService.Error($"Failed to download update: {portableAsset.BrowserDownloadUrl}");
+            _loggerService.Error(ex);
+            return;
+        }
+        
+        var unzipPath = Path.Join(ISettingsManager.GetTemp_DownloadsPath(), "unzip");
+        Directory.CreateDirectory(unzipPath);
+        ZipFile.ExtractToDirectory(downloadZipPath, unzipPath);
```
dragonzkiller, line 65: We should be checking the SHA256 digest before we do any unzipping to verify the download.

dragonzkiller (changes requested):
Admittedly I'm on mobile right now, so my review is limited to a quick look at the moment. I think this currently relies on accepting the scripts will always execute properly. You use C# to download and manage the error handling of the download, but it assumes everything has downloaded and installs properly.

WolvenKit.App/Services/ISettingsManager.cs
notaspirit: changed that in the latest commit

WolvenKit.App/Services/UpdateService.cs
notaspirit: it now checks that in the latest commit (had to sadly switch away from octokit as it doesn't expose the sha256 which was added somewhat recently)

WolvenKit.App/Services/UpdateService.cs
```diff
@@ -0,0 +1,168 @@
...
+        }
+        
+        var unzipPath = Path.Join(tempPath, "unzip");
+        Directory.CreateDirectory(unzipPath);
+        ZipFile.ExtractToDirectory(downloadZipPath, unzipPath);
+        
+        var exePath = Environment.ProcessPath ?? Path.Combine(AppContext.BaseDirectory, "WolvenKit.exe");
+        var scriptPath = Path.Combine(tempPath, "update.ps1");
+        var vbsScriptPath = Path.Combine(tempPath, "update.vbs");
+        File.WriteAllText(scriptPath, $@"
```
wopss, line 79: You should be really careful with this, AV software might flag the whole WKit if you run files like this.

Or, they could even block the execution of these scripts until they are analyzed in their clouds.

WolvenKit.App/Services/UpdateService.cs
dragonzkiller: Hmmm I see that now. While I know it seems advantageous for us to use it, this would seem like a pretty big security flaw if that's the case. Seems to be it though... Alright

notaspirit:
It's ready for the next review cycle, changed the scripts out for a separate executable that is part of the release assets and added a simple dialog.

to test this new version change the repo owner in `WolvenKit\WolvenKit.App\Services\UpdateService.cs` Line `180` to `notaspirit` I have used the modified build actions to build release assets like they would be after this is merged (only change is the repo owner so it continues to pull from the test assets)

manavortex (approved):

dragonzkiller:
Merged!

PR #2504, opened 2025-07-31
feat: display uint props as known enum and bitfield types

flibdev:
# feat: Display `uint` props as known `enum` types

**Implemented:**
Some `uint` properties represent values of existing `enum` and `bitfield` types, but aren't stored as such. Rather than change the way they're stored or loaded, I added the `DisplayAsEnum` attribute and two new editor templates to render them as if they were those types.

Currently the following fields are affected:
- `CMaterialParameterInfo.Type`
- `MaterialTechnique.StreamsToBind`
- `MaterialTechnique.FeatureFlagsMask.Flags`
- `rendChunk.VertexFactory`

### DIFF
--- a/WolvenKit.RED4/Types/Classes/rendChunk.cs
+++ b/WolvenKit.RED4/Types/Classes/rendChunk.cs
@@ -45,7 +45,8 @@ namespace WolvenKit.RED4.Types
 		}
 
 		[Ordinal(5)] 
-		[RED("vertexFactory")] 
+		[RED("vertexFactory")]
+		[DisplayAsEnum<EMaterialVertexFactory>]
 		public CUInt8 VertexFactory
 		{
 			get => GetPropertyValue<CUInt8>();
--- a/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
@@ -85,6 +85,7 @@ public partial class ChunkViewModel : ObservableObject, ISelectableTreeViewItemM
     private readonly RDTDataViewModel? _tab;
 
     private Flags? _flags;
+    private Type? _displayAsEnumType;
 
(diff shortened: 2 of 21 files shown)


seberoth:
Just checked `WolvenKit.RED4` changes. Looks good to me. Can't test right now, though.

flibdev:
- Removed leftover restriction from `UIntAsEnumEditor` for values to be unsigned or non-zero (leftover from bitfield logic).
- Removed unused enum helper methods from DisplayAsEnumAttribute.
- Renamed editors to `DisplayAs___Editor`

WolvenKit.RED4/Types/Classes/rendChunk.cs
```diff
@@ -45,7 +45,8 @@ public CArray<CName> MaterialId
 		}
 
 		[Ordinal(5)] 
-		[RED("vertexFactory")] 
+		[RED("vertexFactory")]
+		[DisplayAsEnum<EMaterialVertexFactory>]
```
manavortex, line 49: Should this maybe go into ClassesExt?

manavortex (approved):

WolvenKit.RED4/Types/Classes/rendChunk.cs
flibdev: It could do, but I was worried about obfuscation and confusion if someone in the future wanted to know why the fields were rendered as they were. I just needed some way to pass info from the RedClasses down the pipeline to the renderer without hardcoding it, an attribute with a strongly-typed param seemed the cleanest way to do that

seberoth: Its fine as is.
Doesn't make sense to put the attribute in ClassesExt

seberoth: Only thing which is questionable is, if its really needed to do add a generic attribute.

PR #2505, opened 2025-07-31
Feature: create phase node from selected nodes

MrBilL61:
added option to create phase node from from multiple selected nodes


manavortex (approved):
Wonder if "Convert to phase node" wouldn't be better as label for the context menu item, but LGTM :)

PR #2507, opened 2025-08-01
World Builder Prop List generator

manavortex:
# World Builder Prop List generator

image


justarandomguyintheinternet:
Looks correct to me, paths check out and all that. Would just maybe change some of the user-facing terminology (I think most people think of `.ent`'s when hearing the word "Prop", altho in this context it can also be a `.mesh` (And in the future potentially other resource types)) 👍

PR #2510, opened 2025-08-03
feat(CLI): add build command

poirierlouis:
# feat(CLI): add build command

**Implemented:**
- `build <path1> <pathN>` command in `WolvenKit.CLI`. It builds a WKit project like the button "Build Project" of the application.

**Additional notes:**
- implementation is derived from [RED4Controller](https://github.com/WolvenKit/WolvenKit/blob/e81cbad020ddf085bd0b50a844aabd1a4ffcc2e8/WolvenKit.App/Controllers/RED4Controller.cs#L467).
- it doesn't support REDmod content: can be implemented later when requested by a user.
- it doesn't create a `.zip` file: can be implemented later with an option like `--zip`.

### DIFF
--- a/WolvenKit.CLI/Commands/BuildCommand.cs
+++ b/WolvenKit.CLI/Commands/BuildCommand.cs
@@ -1,0 +1,37 @@
+using System.CommandLine;
+using System.CommandLine.NamingConventionBinder;
+using System.IO;
+using CP77Tools.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using WolvenKit.Core.Interfaces;
+
+namespace CP77Tools.Commands;
+
+internal class BuildCommand : CommandBase
+{
+    private const string s_description = "Build project(s) like WolvenKit application.";
+    private const string s_name = "build";
+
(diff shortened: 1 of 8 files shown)


WolvenKit.CLI/Commands/BuildCommand.cs
```diff
@@ -0,0 +1,37 @@
...
+using CP77Tools.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using WolvenKit.Core.Interfaces;
+
+namespace CP77Tools.Commands;
+
+internal class BuildCommand : CommandBase
+{
+    private const string s_description = "Build project(s) like WolvenKit application.";
```
dragonzkiller, line 13: Should say something like `Builds all WolvenKit .cpmodproj project(s) found in <paths>`. We should double check if `[path]` or `<path>` is better. But you get the idea.

dragonzkiller (changes requested):
Everything looks good to me. Just some grammatical stuff and wording suggestions to make it sound better for a CLI vs a GUI

dragonzkiller (approved):

PR #2622, opened 2025-09-04
fix(ImportViewModel): prevent concurrent file access errors when importing meshes/morphtargets

### LINKED ISSUE #2343: WolvenKit 8.16.1 randomly fails to import some of the files in a batch.
Hi.

I have some projects with fairly large number of mesh files, which i import all at once, and the latest version of WolvenKit would randomly fail to import some of them. I say randomly because it's not the same amount of files that fail (i did a few runs, and sometimes it would be 72/86, other times it's 76/86 and so on), and not necessarily the same files, though it seems to affect the larger files more frequently.

The error it throws says that the process can't access the .mesh file, and that it is used by another process, but there isn't any other process. Restarting the program doesn't help. I downgraded back to version 8.15.0, and it successfully imports all of the files in the same project.

Here's an example of an error message:

```
[4/17/2025 23:38:32] [Error    ] Message: The process cannot access the file 'D:\Games\add\cp2077\wolvenKitFiles\projects\cvbTPP\source\archive\base\characters\common\player_base_bodies\player_female_average\arms_hq\nails\a0_000_pwa_base__nails_l.mesh' because it is being used by another process.
Source: System.Private.CoreLib
StackTrace:    at Microsoft.Win32.SafeHandles.SafeFileHandle.CreateFile(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.Strategies.OSFileStreamStrategy..ctor(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.Strategies.FileStreamHelpers.ChooseStrategyCore(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.File.OpenRead(String path)
   at WolvenKit.App.Models.FileSystemArchive.CopyFileToStream(Stream stream, UInt64 hash, Boolean decompressBuffers)
   at WolvenKit.RED4.Archive.FileEntry.Extract(Stream output, Boolean decompressBuffers)
   at WolvenKit.RED4.Archive.FileEntry.Extract(Stream output)
   at WolvenKit.Modkit.RED4.ModTools.InternalTryFindFile(UInt64 hash, FindFileRecord& result, Boolean excludeCustomArchives)
   at WolvenKit.Modkit.RED4.ModTools.TryFindFile(ResourcePath path, FindFileRecord& result, Boolean excludeCustomArchives)
   at WolvenKit.Modkit.RED4.ModTools.ImportMorphTargets(FileInfo inGltfFile, Stream inTargetStream, GltfImportArgs args)
   at WolvenKit.Modkit.RED4.ModTools.ImportGltf(RedRelativePath rawRelative, DirectoryInfo outDir, GltfImportArgs args)
```

poirierlouis:
# fix(ImportViewModel): prevent concurrent file access errors when importing meshes/morphtargets

**Fixed:**
- importing a batch of .gltf / .glb files with meshes and morphtargets would fail. Process uses parallelism which lead to concurrent access of files due to dependency between mesh/morphtarget.

**Additional notes:**
Implementation can be easily updated in the future to account for new conflicts / order of processing based on file's extension and data type.

Closes #2343

### DIFF
--- a/WolvenKit.App/ViewModels/Importers/ImportViewModel.cs
+++ b/WolvenKit.App/ViewModels/Importers/ImportViewModel.cs
@@ -126,38 +126,47 @@ public partial class ImportViewModel : AbstractImportExportViewModel
         var progress = 0;
         _progressService.Report(0.1);
 
-        var total = 0;
-        var successful = 0;
-
-        //prepare a list of failed items
-        var failedItems = new List<string>();
-
-        var toBeImported = Items
-            .Where(importExportItem => importExportItem.IsChecked ||
-                                       (all && (VisibleItemPaths.Count == 0 || VisibleItemPaths.Contains(importExportItem.BaseFile))))
-            .Where(x => !x.Extension.Equals(ERawFileFormat.wav.ToString()))
+        var importItems = Items
+            .Where(x => x.IsChecked || (all && (VisibleItemPaths.Count == 0 || VisibleItemPaths.Contains(x.BaseFile))))
+            .Where(x => !x.Extension.Equals(nameof(ERawFileFormat.wav)))
             .Cast<ImportableItemViewModel>()
             .ToList();
 
-        total = toBeImported.Count;
-        await Parallel.ForEachAsync(toBeImported, async (item, cancellationToken) =>
-        {
-            await Application.Current.Dispatcher.InvokeAsync(() => _appViewModel.SaveFile(item.BaseFile));
-            if (await ImportSingleAsync(item, projectArchive))
-            {
-                Interlocked.Increment(ref successful);
-            }
-            else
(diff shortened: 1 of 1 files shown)


poirierlouis:
Great! This patch will be available in the nightly channel, likely tomorrow.

PR #2642, opened 2025-09-10
fix(MainView): preserve window's size after minimizing

### LINKED ISSUE #2375: Wolvenkit resolution/fullscreen issue
Any time I minimize the window, it returns to 1080p window on my 1440p display, which is quite annoying, having to adjust it constantly.

Only happens on 8.16 (and the current nightly), previous versions work fine.

Video illustrating issue:
https://youtu.be/4PjqlfMMWuk

poirierlouis:
# fix(MainView): preserve window's size after minimizing

**Fixed:**
- Attempt to be tested by OP of the issue, as I can't reproduce this on my side.

**Additional notes:**
Closes #2375


seberoth:
Did some digging:
1. It always happened after the first minimizing (gone back to 8.12.0, couldn't go much further due to missing Net Framework 7)
2. It started happening after every minimizing after merging #2112 . So probably due to SF update

seberoth:
Updating SF doesn't solve the issue

manavortex:
Moving them across monitors doesn't bust the UI scale anymore - and maximizing/minimizing restores to non-maximized size instead of the resolution of the smallest display. I'd say this is definitely a step forward!

poirierlouis:
Hum, it's up to you then, should I fix the merge conflicts (to merge as-is anyway) or we drop this PR for now until someone else can fully fix it?
I only have one monitor so I can't work on this further.

notaspirit:
I'd say drop it for now considering it doesn't really improve the behavior and having half fixed things can make it harder to maintain. 
Appreciate your effort though, putting all that time in despite not being able to replicate the issue.

notaspirit:
I don't really think it's better, mainly different, and from the end users perspective in a way less transparent. Since now instead of it always incorrectly remaximizing to 1080p it sometimes would do 1080p sometimes less.

seberoth:
I think, its worse for me though.
Now the window randomly goes from fullscreen into windowed and back to fullscreen while performing various actions.
Even taking a screenshot of wkit does that...

poirierlouis:
No problem, was a bit silly of me to try that anyway, without a second monitor...
I'll close it for now, so we can keep the issue active, and anyone can later take a look at what I've done if it can help them.

PR #2653, opened 2025-09-14
Fix for packing project with invalid characters in "Mod Name"

dragonzkiller:
Ref: https://discord.com/channels/717692382849663036/808086068023918603/1416827068263694396

**Fixed:**
- In project configuration "Mod Name" can contain invalid characters when generating the `.archive`. Fix sanitizes before exporting.

**Additional notes:**
- Warns user when issues are found.

### DIFF
--- a/WolvenKit.Modkit/RED4/Pack.cs
+++ b/WolvenKit.Modkit/RED4/Pack.cs
@@ -1,7 +1,11 @@
 using System.IO;
+using System.Linq;
+using Serilog;
 using WolvenKit.Helpers;
+using WolvenKit.Interfaces.Extensions;
 using WolvenKit.RED4.Archive;
 using WolvenKit.RED4.Archive.IO;
+using WolvenKit.RED4.Types;
 
 namespace WolvenKit.Modkit.RED4
 {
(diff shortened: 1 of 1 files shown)


WolvenKit.Modkit/RED4/Pack.cs
```diff
@@ -34,7 +34,17 @@ public bool Pack(DirectoryInfo infolder, DirectoryInfo outpath, string? modname
             var outFile = Path.Combine(outpath.FullName, $"{infolder.Name}.archive");
             if (modname != null)
             {
-                outFile = Path.Combine(outpath.FullName, $"{modname}.archive");
+                var sanitizedModName = new string(modname.Where(c => !Path.GetInvalidFileNameChars().Contains(c)).ToArray()).Trim();
```
manavortex, line 37: Look into `StringPathExtensions.cs`, this looks like it could go there (or as if you could use methods from there)

manavortex: > I purposely avoid using that because it doesn't exactly work for some of the trails I did. however, moving it to "INVALID NAME" wasn't working either, so I'll fix that too.

You could fix the functions? We're trying to cut down on code duplication regarding string path validation stuff :)

> We discussed this a bit on the discord, but we should try to limit how much we're restricting the user to name their work.
As we have seen, users can't be trusted :P But the fields should act the same IMO

PR #2663, opened 2025-09-19
add Visibility binding for Mlsetup item

misterchedda:
**Implemented:**
Small QoL: show 'Open in MlSetupBuilder' option only for mlsetup files. It currently shows for all files on right click (but greyed out). PR just adds a Visibilty binding for the menu item so it shows up only for supported files


poirierlouis (approved):
LGTM

PR #2667, opened 2025-09-23
feat: add scene editor button bar

misterchedda:
# Add a button bar in Scene Editor

**Implemented:**
PR adds a button bar to the Scene Editor that will allow users to create commonly used scene properties easily.

Currently adding new entries for scene properties like an actor or dialogue requires:
- Knowing how the ids for each type of property is calculated. For eg actors have performer ids that are 256 increments, but a prop's performer id is in 256+2 increments. This gets really annoying as a scene grows beyond the first couple of entries and having to note and remember the formulas without mistakes
- Tasks that are mentally atomic in nature like "create an actor" actually mean adding entries in multiple places (for actors, this means appending in `actors` and in `debugSymbols` for example) so again error prone

Note that - there are no workflow changes for editing and defining the properties after creating them. The buttons only help create new entries and automatically calculate ids + prefill defaults.

PR adds buttons that create the following:

In "Actors & Props" tab:
- Actor
- Prop
image

In "Dialogue" tab:
- Dialogue
- Option
image

In "Asset Library" tab:
- Workspot
- Animation
- Effect
image

Note: Animation creation loads .anims files to extract and populate animation names automatically and effect creation loads .effect files to populate compiled effect event information.

The buttons go along with a new input dialog (`SceneInputDialogView`). The buttons execute commands via new [RelayCommand] handlers in SceneGraphViewModel. Also added some resource helpers in scnSceneResource to insert props and derive performer ids and debug symbols.

Mental note to myself for a future PR: we will also need a "Recalculate Actor ID" button. This is because of `playerActors` which is annoyingly appended to `actors` so typically when you're working on a scene and decide to add a new actor but you already have `playerActors` populated, you need to now recalculate new ids for the player actor and its usage across the entire scene (ie not just in all nodes but also in `screenplayStore`). This button would theoretically do that for you but its a lot more work and destructive in nature (and will likely be slow)

### DIFF
--- a/WolvenKit.App/Helpers/CvmDropdownHelper.cs
+++ b/WolvenKit.App/Helpers/CvmDropdownHelper.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using WolvenKit.App.ViewModels.Shell;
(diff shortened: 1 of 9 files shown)


MrBilL61:
@misterchedda I run quick tests, here's what I noticed:

In "Actors & Props" tab:

- Actor -> creates duplicated entry in `performersDebugSymbols` array
- Prop -> works fine

In "Dialogue" tab:

- Dialogue - works fine
- Option - works fine

(It would be nice to have `Generate New CRUID` button for locStringID for both)

In "Asset Library" tab:

- Workspot -> `dataId` for the next new workspot should always be greater than the last one and `workspot` array should be sorted in ascending order
- Animation -> works fine
- Effect -> `effectInstanceId` can start from index 0 (according to vanilla scenes)

PR #2677, opened 2025-09-26
Auto loading for HashService/CRUIDService

seberoth:
# Auto loading for HashService/CRUIDService

**Implemented:**
- Added optional auto loading to HashService and CRUIDService. Ensures compatibility with external applications from before 8.17 release (nuget)

### DIFF
--- a/WolvenKit/GenericHost.cs
+++ b/WolvenKit/GenericHost.cs
@@ -62,10 +62,10 @@ namespace WolvenKit
                 {
                     // services
                     services.AddSingleton(typeof(ISettingsManager), SettingsManager.Load());
-                    services.AddSingleton<IHashService, HashServiceExt>();                      // can this be transient?
-                    services.AddSingleton<CRUIDService>();                                      // can this be transient?
-                    services.AddSingleton<MySink>();                                            // can this be transient?
-                    services.AddSingleton<ILoggerService, SerilogWrapper>();                    // can this be transient?
+                    services.AddSingleton<IHashService, HashServiceExt>();                                      // can this be transient?
+                    services.AddSingleton<CRUIDService>(x => new CRUIDService(false));    // can this be transient?
+                    services.AddSingleton<MySink>();                                                            // can this be transient?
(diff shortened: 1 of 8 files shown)


WolvenKit/GenericHost.cs
```diff
@@ -62,10 +62,10 @@ public static IHostBuilder CreateHostBuilder() => Host
...
                     // services
                     services.AddSingleton(typeof(ISettingsManager), SettingsManager.Load());
-                    services.AddSingleton<IHashService, HashServiceExt>();                      // can this be transient?
-                    services.AddSingleton<CRUIDService>();                                      // can this be transient?
-                    services.AddSingleton<MySink>();                                            // can this be transient?
-                    services.AddSingleton<ILoggerService, SerilogWrapper>();                    // can this be transient?
+                    services.AddSingleton<IHashService, HashServiceExt>();                                      // can this be transient?
+                    services.AddSingleton<CRUIDService>(x => new CRUIDService(false));    // can this be transient?
+                    services.AddSingleton<MySink>();                                                            // can this be transient?
+                    services.AddSingleton<ILoggerService, SerilogWrapper>();                                    // can this be transient?
```
poirierlouis, line 68: If I understand well, and to be sure:
- now default ctor of `HashService` / `CRUIDService` will auto-load synchronously. Which brings back the default code-flow before my changes?
- `L65` and `L66` disable auto-load to preserve improvements regarding FCP I introduced?

It seems I complicated myself when changing the behavior, thanks!

LGTM :)

poirierlouis (approved):

WolvenKit/GenericHost.cs
seberoth: Yeah, thats right.
HashServiceExt (which is only used in GUI) defaults to no auto load (keeping your improvements)
Everything else will by default auto load again (ext. projects using nugets, cli, unit tests, ...) which is fine.
Could do CRUIDServiceExt to but doesn't make sense, so it gets constructed with auto load turned off.

PR #2679, opened 2025-09-27
importargs: disable garment support param by default for wepaons etc

manavortex:
# importargs: disable garment support param by default for wepaons etc

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/ImportableItemViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/ImportableItemViewModel.cs
@@ -191,5 +191,5 @@ public class ImportableItemViewModel : ImportExportItemViewModel
     ];
 
     private static bool ShouldHaveGarmentSupport(string relativePath) =>
-        s_partialsToDisableGarmentSupport.Any(relativePath.Contains);
+        !s_partialsToDisableGarmentSupport.Any(relativePath.Contains);
 }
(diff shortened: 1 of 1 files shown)


WolvenKit.App/ViewModels/Tools/ImportableItemViewModel.cs
```diff
@@ -59,9 +61,11 @@ or ERawFileFormat.cube
                 or ERawFileFormat.gltf
                 => new GltfImportArgs()
                 {
-                    ImportFormat = formatFromFilename ?? GltfImportAsFormat.Mesh,
                     // Mesh won't have an internal extension
-                    ImportGarmentSupport = formatFromFilename is null or GltfImportAsFormat.MeshWithRig 
+                    ImportFormat = formatFromFilename ?? GltfImportAsFormat.Mesh,
+                    // Garment support: disable for vehicles, weapons, effects etc
+                    ImportGarmentSupport = ShouldHaveGarmentSupport(fileName),
+                    IgnoreGarmentSupportUVParam = ShouldHaveGarmentSupport(fileName),
```
poirierlouis, line 68: Currently, `ShouldHaveGarmentSupport` returns true whenever a word of the list `s_partialsToDisableGarmentSupport` is present in the `fileName`.
Either comments are wrong or implementation is wrong, no?

poirierlouis (changes requested):

WolvenKit.App/ViewModels/Tools/ImportableItemViewModel.cs
manavortex: Absolutely, that's missing a !

poirierlouis (approved):
LGTM

PR #2682, opened 2025-09-27
throw if bones are imported into a mesh without

### LINKED ISSUE #2680: Weapon armature gets lost on export
## Exporting a weapon mesh leads to the armature getting lost. 

This does not happen for all meshes - I tried with a random clothing item (torso), and the armature was preserved.

## Example file(s)
[source.zip](https://github.com/user-attachments/files/22572458/source.zip)

## Steps to reproduce the behavior:
1. Use a project with the attached `source` folder
2. Import `high_frequency_blade_with_armature.glb` into blender and see that it has an armature
3. Overwrite `high_frequency_blade.glb` with the control file to rule out user error
4. Import `high_frequency_blade.glb` into `high_frequency_blade.mesh`
5. Export `high_frequency_blade.mesh`
6. Import `high_frequency_blade.glb` into Blender and see that it no longer has an armature

## Version 
 - `main` as of Sep 27

manavortex:
# check for mesh bones

throw exception if bones are imported into a mesh that doesn't have any

fixes #2680

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshTools.cs
@@ -46,7 +46,7 @@ namespace WolvenKit.Modkit.RED4.Tools
             catch (ArgumentNullException)
             {
             }
-            
+
 
             if (model == null)
             {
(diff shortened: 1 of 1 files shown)


WolvenKit.Modkit/RED4/Tools/MeshTools.cs
```diff
@@ -655,6 +655,11 @@ public static (int numLodLevels, int numSubmeshesPerLod) GetLodInfo(CMesh mesh)
 
         public static void UpdateMeshJoints(ref List<RawMeshContainer> meshes, RawArmature? existingJoints, RawArmature? incomingJoints, string fileName = "")
         {
+            if (existingJoints is not { BoneCount: > 0 } && incomingJoints is { BoneCount: > 0 })
```
poirierlouis, line 658: Does C# allow `is { BoneCount: == 0}` like syntax instead? It would be a tad easier to read imo.
It means the number of incoming joints is not required to equal the number of existing joints? Only this rule, has incoming joints and no existing joints, is considered an error?

poirierlouis: Ah alright thanks, not used to this C# syntax enough.

poirierlouis (approved):
LGTM

PR #2691, opened 2025-10-03
New setting for subfolder name (modname/moddername)

manavortex:
# New setting for subfolder name (modname/moddername)

Added a setting to use the modder name as default folder for generated files (default: false)
Exposing paths via CP77project: CET directory, Tweak directory, Script directory


seberoth:
Just a thought: Wouldn't it make sense if the options to enable it was also a project option, not a global one?
So if someone shares the project or reinstalls wkit, it would yield the same output as before?

manavortex:
That's a good point, but I'm not certain in regards of separation of concerns etc. Does it belong there?
In my experience, people rarely share full projects anyway, it's just the source folder

seberoth:
Still don't know which popup...
But other then that, yeah I would suggest to replace existing calls now. Else it will be forgotten...
Other then that it LGTM

manavortex:
image

The popup that shows when you don't have a modder name set - called from `AppViewModel` in the getter. I figured this would be easiest. 

Think this is good to merge now? I adjusted the remaining calls.

seberoth:
Ah, but that's only shown if someone is adding and ArchiveXL file, not in the other cases its used.
Thought about a general warning, since it not only used there.
Also the dialog mention the global settings now while the Author field of the project settings is used.

Also the merge/rebase didn't work out well :P
grafik

manavortex:
@seberoth, think it's OK now?

PR #2694, opened 2025-10-04
Dynamically show used args based on selections

seberoth:
# Dynamically show used args based on selections

**Implemented:**
- Dynamically show used args based on selections

https://github.com/user-attachments/assets/6960c57f-c607-4182-88fc-dd102b313fd7

If any "dependencies" are missing, pls tell me. (Or add them yourself :P)

### DIFF
--- a/WolvenKit.Common/Model/Arguments/ImportExportArguments.cs
+++ b/WolvenKit.Common/Model/Arguments/ImportExportArguments.cs
@@ -1,8 +1,34 @@
+using System;
 using System.ComponentModel;
+using System.Runtime.CompilerServices;
 using CommunityToolkit.Mvvm.ComponentModel;
 
 namespace WolvenKit.Common.Model.Arguments
 {
+    /// <summary>
+    /// Tags a property as accessible in a WolvenKit Script
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    public class WkitScriptAccess : Attribute
+    {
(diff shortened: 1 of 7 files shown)


manavortex (approved):

manavortex:
LGTM, let's test it in production

manavortex:
I'm seeing Schroedinger's missed import arg. If we wait, it'll be fine - if we include it in the release, we'll notice that something critical is missing and a feature is unusable now. >.<

WolvenKit.Common/Model/Arguments/ImportExportArguments.cs
```diff
@@ -1,8 +1,34 @@
...
+    {
+        public string PropertyName { get; }
+        public object?[] Values { get; }
+
+        public UsedWith(string propertyName, params object?[] values)
+        {
+            PropertyName = propertyName;
+            Values = values;
+        }
+    }
```
poirierlouis, line 31: Perhaps this, and associated logic below, could be refactored in some Helper class? So it can be reused elsewhere when needed?
I wonder if ReactiveUI already allow this kind of implementation. I'll check it, but it would be nice to have it ReactiveUI-ready regarding WolvenKit@Next :)

poirierlouis (commented):
Seems nice, I left a comment. Perhaps if `[UsedWith]` is made reusable, add documentation in `CONTRIBUTING.md` too, to explain its usage?

PR #2695, opened 2025-10-04
fix: info bar text scaling and padding in scene/questPhase editors

misterchedda:
Fixes info bar text scaling and padding issues in scene and quest editor (@poirierlouis I somehow missed seeing this in #2688 and noticed that the info bar was inappropriately scaled after this change)

This PR completes the fix by replacing hardcoded font sizes, margins with dynamic resources and also doing the same for the questphase edtitor which has a similar info bar

Before:
image 2 (4)

After:
image 1 (6)

### DIFF
--- a/WolvenKit/Views/Documents/QuestPhaseGraphView.xaml
+++ b/WolvenKit/Views/Documents/QuestPhaseGraphView.xaml
@@ -23,26 +23,26 @@
                 <!-- Merge Scene Editor specific converters (reused for quest phase) -->
                 <ResourceDictionary Source="/WolvenKit;component/Resources/Converters/SceneEditorConverters.xaml" />
             </ResourceDictionary.MergedDictionaries>
-            
-        <Style x:Key="SpacedWolvenKitTabControl" 
+
+            <Style x:Key="SpacedWolvenKitTabControl" 
                BasedOn="{StaticResource WolvenKitTabControl}"
                TargetType="{x:Type syncfusion:TabControlExt}">
-            <Setter Property="TabItemSelectedBackground" Value="{StaticResource WolvenKitRedShadow}" />
-            <Setter Property="TabItemHoverBackground" Value="{StaticResource WolvenKitRed}" />
-            <Setter Property="TabItemSelectedBorderBrush" Value="{StaticResource WolvenKitRedShadow}" />
-            <Setter Property="TabItemHoverBorderBrush" Value="{StaticResource WolvenKitRed}" />
-        </Style>
-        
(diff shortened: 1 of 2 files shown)


WolvenKit/Views/Documents/QuestPhaseGraphView.xaml
```diff
@@ -52,84 +52,84 @@
...
             <RowDefinition Height="Auto" />
             <RowDefinition Height="*" />
         </Grid.RowDefinitions>
-        
         <!-- Toolbar -->
-        <Border Grid.Row="0" Background="{StaticResource ContentBackground}" BorderBrush="{StaticResource BorderAlt}" BorderThickness="0,0,0,1" Padding="8,4">
+        <Border Grid.Row="0" Background="{StaticResource ContentBackground}" BorderBrush="{StaticResource BorderAlt}" BorderThickness="0,0,0,1" Padding="{DynamicResource WolvenKitMarginTiny}">
             <StackPanel Orientation="Horizontal">
-                <iconPacks:PackIconMaterial Kind="InformationOutline" VerticalAlignment="Center" Margin="0,0,8,0" Width="12" Height="12"/>
+                <iconPacks:PackIconMaterial Kind="InformationOutline" VerticalAlignment="Center" Margin="{DynamicResource WolvenKitMarginSmallRight}" Width="12" Height="12"/>
```
poirierlouis, line 86: Can you replace with `IconBox` instead (see contributing.md)? `Size=` can be defined using dedicated resources you'll find at the beginning of `App.Sizes.xaml`.
Same for other `PackIconMaterial` below ofc :)
I'll be able to run it tomorrow to test at 100% and 150%

poirierlouis (changes requested):
Overall looks good, see comment and I'll be able to give the go ahead tomorrow :)

WolvenKit/Views/Documents/QuestPhaseGraphView.xaml
misterchedda: Used `IconBox` where I can. Seems like it doesn't support dynamic data bindings like `Kind="{Binding Icon}"`  due to its internal DataContext override so the questphase/scene tab icons will continue using `PackIconMaterial` for now

poirierlouis: Nice :) yes I don't remember exactly, was a pain at the time. No problem in this case.

poirierlouis:
Current rendering after fixes for UI scaling

**At 100%:**
image

**At 150%:**
image

PR #2704, opened 2025-10-06
fix: use correct id for first item in dialogue creation

misterchedda:
**Fixed:**

First id for dialogue lines in a scene always start from 1, not 2 (2 is for options) (This is used by the "Create Dialogue" button)

I didn't notice this earlier because I didn't test in a fresh scene with no lines


poirierlouis (approved):
LGTM

PR #2719, opened 2025-10-16
ignore !!+ in cybersearch

manavortex:
# ignore !!+ in cybersearch

just accidentally crashed the asset browser by throwing in a mod name


notaspirit:
Do you still have the search term that caused it to freeze for you in the first place? Maybe it's not just a lot of `!` but a combination of some other factors as well?

manavortex:
I was in the mod browser, and I wanted to search in an archive that started with a million of bangs, and I forgot to put the archive: prefix.

PR #2723, opened 2025-10-19
Inkatlas generator: improvements

manavortex:
Inkatlas generator: improvements
- There's now a dropdown menu with a few presets for icon file formats
- You can now pick an existing file to overwrite

image


rfuzzo (approved):
seems to work

PR #2726, opened 2025-10-20
Feature/generate props

manavortex:
Includes [2691](https://github.com/WolvenKit/WolvenKit/pull/2691), so please merge that first :)

# Feature/generate props

image

Generates an AMM/WB prop. User can add up to 4 meshes. txt/lua files and/or corresponding entries will be generated automatically.


seberoth:
Could you sync it with main?
With #2691 merged now?

PR #2732, opened 2025-11-01
Tweak: sort before writing

manavortex:
# Tweak: sort before writing


seberoth:
Can you remove #2455 from this?

PR #2746, opened 2025-11-22
AXL generator: Use project files

### LINKED ISSUE #2744: FormalShirt creats a t-shirt in the AXL Creator tool in Wkit.
(replying to @manavortex on Discord)

There is an issue when trying to create a `FormalShirt` 

The issue is that it creats a T-shirt instead, (which is REALLY annoying for me)

As Manvortex requested, here are the meshes to a formal shirt with male and female versions 

```
base\characters\garment\player_equipment\torso\t1_102_shirt__vest\t1_102_pma_shirt__vest.mesh

base\characters\garment\player_equipment\torso\t1_102_shirt__vest\t1_102_pwa_shirt__vest.mesh
```

manavortex:
# AXL generator: Use project files

- Let user pick from existing meshes in project
- Allows reading variants from mesh appearances
- Fixes #2744

### DIFF
--- a/WolvenKit.App/ViewModels/Dialogs/AddArchiveXlFilesDialogViewModel.cs
+++ b/WolvenKit.App/ViewModels/Dialogs/AddArchiveXlFilesDialogViewModel.cs
@@ -13,21 +13,6 @@ using WolvenKit.Interfaces.Extensions;
 
 namespace WolvenKit.App.ViewModels.Dialogs;
 
-public class FilePathItemDataTemplateSelector : DataTemplateSelector
-{
-    public override DataTemplate?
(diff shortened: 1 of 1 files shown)


WolvenKit.App/ViewModels/Dialogs/AddArchiveXlFilesDialogViewModel.cs
```diff
@@ -10,22 +13,39 @@
 
 namespace WolvenKit.App.ViewModels.Dialogs;
 
+public class FilePathItemDataTemplateSelector : DataTemplateSelector
```
seberoth: Can that be removed?
Always returns `null` and is never used?

PR #2748, opened 2025-11-23
disable auto-creation of tweak folder

manavortex:
# disable auto-creation of tweak folder

No longer auto-creating tweak directory if it doesn't exist (added logic to calling methods instead)


manavortex:
Some calling methods are trying to write into that folder, which won't work if it doesn't exist. It's where I added the Directory.CreateDirectory calls.

PR #2751, opened 2025-11-24
fixed mlmask roundtrip 

notaspirit:
# fixed mlmask roundtrip 

**Fixed:**
- mlmask roundtrip (import export with unchanged files) failing for mlmasks with dimensions that are only divisible by two while not being a power of two which some vanilla mlmasks are

**Additional notes:**
example asset is the embedded mlmask in `base\worlds\03_night_city\sectors\_global\terrain\terrain_aaayaai_cell_16335_16431.mesh`


seberoth:
I think there were issues with that in the past. (In-Engine / In-Game i think, but not 100% sure)
If you say it works ingame, I will approve it. Can't test it currently.

notaspirit:
I did test it with the vanilla file that uses that 68x68 mlmask and it worked fine before and after the roundtrip. 

If you need to I can test it with another modified mlmask that was originally a power of 2 but now is only divisible by 2 as well.

seberoth (approved):
Should be fine then :)

PR #2757, opened 2025-11-29
use project's ModName instead of Name

manavortex:
# better distinction of active project's Name/ModName

`ModName` will be used inside file paths etc
`ProjectName` is the name of the project for the UI


poirierlouis:
Is it possible to explain "high-level" changes, reasons, etc? Please :)
It would make for a nice ticket to refer/link to... in case this topic is brought back, it be settled already.

manavortex:
We had a discussion about this somewhere - maybe on the server. I'll do a write-up when I fix this :)

PR #2761, opened 2025-12-07
ent preview: resolve dynamic names in .app

manavortex:
# ent preview: resolve dynamic names in .app

Currently, root entity preview fails if appearance names in .app file use conditionals like `&camera=tpp`.


seberoth (approved):
LGTM

PR #2763, opened 2025-12-07
find broken references: filtering more false positives

manavortex:
# find broken references: filtering more false positives

will now filter out "0", "none", and CDPR's red ent flags


poirierlouis (changes requested):

poirierlouis (approved):
LGTM, thanks :)

PR #2773, opened 2025-12-20
fix component names for amm props

manavortex:
# fix component names for amm props

underscore in component name breaks resizing


poirierlouis (commented):
Does it mean underscore suffix is not used in AMM or it's only WKit related?

PR #2776, opened 2026-01-02
added TintColor attribute to vectors-as-colors

manavortex:
# added TintColor attribute to vectors-as-colors


poirierlouis (approved):
LGTM

PR #2785, opened 2026-01-13
feat(GraphEditor): add timeline editor for scnSectionNode

misterchedda:
# Add timeline editor for scnSectionNode in Scene Editor

image

One of the remaining few things that are painful to work with in scenes. Especially when working on longer Sections, it's quite tedious and error-prone to rearrange events and manually recalculate StartTime + Duration for each.

**Implemented:**

The PR adds a visual timeline editor panel that appears when selecting a scnSectionNode in a scene file:

- Drag events horizontally to change StartTime (and rearrange order in events list)
- Resize events to change Duration
- Snap to grid (configurable interval) and zoom/pan
- Auto-recalculate section duration based on events

Events are grouped into tracks by category with labels showing useful metadata (just like in the graph view) w actor names, dialogue text, etc

Changes:
- New `SectionTimelineViewModel`, `TimelineService`
- New `SectionTimelineView` panel integrated into `SceneGraphView`

### DIFF
--- a/WolvenKit.App/Models/Timeline/TimelineEvent.cs
+++ b/WolvenKit.App/Models/Timeline/TimelineEvent.cs
@@ -261,31 +261,38 @@ public partial class TimelineEvent : ObservableObject
     
     private string GetDialogText(scnDialogLineEvent dialogLine)
     {
-        if (dialogLine.ScreenplayLineId != null && _sceneResource != null)
+        if (dialogLine.ScreenplayLineId == null || _sceneResource == null)
         {
-            CUInt32 screenplayId = dialogLine.ScreenplayLineId.Id;
-            var screenplayLine = _sceneResource.ScreenplayStore?.Lines?.FirstOrDefault(line => line.ItemId?.Id == screenplayId);
-            
-            if (screenplayLine?.LocstringId != null && 
-                _sceneResource.LocStore?.VdEntries != null &&
-                _sceneResource.LocStore?.VpEntries != null)
-            {
-                CRUID locstringRuid = screenplayLine.LocstringId.Ruid;
-                var preferredLocaleId = WolvenKit.RED4.Types.Enums.scnlocLocaleId.en_us;
-                var vdEntry = _sceneResource.LocStore.VdEntries.FirstOrDefault(vd => 
-                    vd.LocstringId?.Ruid == locstringRuid && vd.LocaleId == preferredLocaleId);
-                
-                if (vdEntry?.VariantId != null)
-                {
-                    var vpEntry = _sceneResource.LocStore.VpEntries.FirstOrDefault(vp => vp.VariantId?.Ruid == vdEntry.VariantId.Ruid);
-                    if (vpEntry?.Content != null)
-                    {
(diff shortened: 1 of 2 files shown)


notaspirit:
I'll review the code side of things, someone more familiar with the quest systems should review the workflow / ux.

misterchedda:
> I'll review the code side of things, someone more familiar with the quest systems should review the workflow / ux.

thanks. @MrBilL61 can provide feedback on ux

notaspirit:
I have a couple of points of feedback on different abstraction levels so I will write the more conceptual stuff here with the specifics being attached to their corresponding code.

You have ViewModels without corresponding views, those should just be regular objects instead and not handle any directly UI related things like pixel layout. There should be more abstract values for zoom level, row height etc that the backend provides for the UI layer to interpret. 
Meaning it should structurally be something akin to a TimelineService with the View and ViewModel only handling the platform dependent logic.

I also noticed quite a few comments that don't add any valuable information which clutters the code.

WolvenKit.App/ViewModels/Timeline/TimelineEventViewModel.cs
```diff
@@ -0,0 +1,761 @@
...
+                var performer = ResolvePerformerName(ik.IkData?.Basic?.PerformerId?.Id ?? uint.MaxValue);
+                return string.IsNullOrEmpty(performer) ? $"{prefix} IK" : $"{prefix} IK ({performer})";
+            }
+            
+            if (_event is scneventsSocket socket)
+            {
+                return $"{prefix} Socket";
+            }
+
+            return $"{prefix} {TypeName}";
```
notaspirit: this is better suited to be a switch statement

WolvenKit.App/ViewModels/Timeline/TimelineEventViewModel.cs
```diff
@@ -0,0 +1,761 @@
...
+            if (_event is scnIKEvent ik)
+            {
+                return GetIkDetails(ik);
+            }
+            
+            if (_event is scneventsSocket socket)
+            {
+                return GetSocketDetails(socket);
+            }
+
```
notaspirit: same ad for the titleline, use a switch statement

notaspirit (changes requested):

MrBilL61:
From the quick test I gave it:
1. It sometimes doesn't properly unload the Timeline window if you click out of the Section:
WolvenKit_Mm1Upv6ybc
2. It doesn't properly register mouse button release - still executing event trimming. It needs to be clicked second time to register change
WolvenKit_Vp53PK7vCi
WolvenKit_vXxkRtmfTZ
3. Moving events around works fine.

misterchedda:
@notaspirit thanks for the feedback, I've refactored a bunch. There's now a TimelineService and ViewModels without views converted to regular objects, along with your other feedback resolved

@MrBilL61 thanks for testing, I was able to repro and should be fixed now

notaspirit:
github ux is so great it pings everyone in the quote reply, how many billion dollars is microslop again?

misterchedda:
🫡 thanks all

PR #2788, opened 2026-01-18
Refactor: CVM mesh tools

manavortex:
# Refactor: CVM mesh tools

- Moved a bunch of mesh-related logic out of the CVM into a Tools class 
- Removed unused commands and methods from CVM in favour of directly calling the tools
- To do: Clean this up some more, I guess

### DIFF
--- a/WolvenKit.App/Helpers/CvmMaterialTools.cs
+++ b/WolvenKit.App/Helpers/CvmMaterialTools.cs
@@ -730,7 +730,7 @@ public class CvmMaterialTools
 
                 break;
             case CMesh mesh:
-                // Why is this async?
+                // Async as a workaround for editor freezes if the lists get very long
                 foreach (var externalMaterial in mesh.ExternalMaterials)
                 {
                     await Task.Run(() => ret.Add(externalMaterial.DepotPath));


manavortex:
Thank you :) We can also do 4 eyes or you DM me with questions if you'd like

WolvenKit.App/ViewModels/Tools/CvmMaterialTools.cs
```diff
@@ -0,0 +1,789 @@
...
+        cvm.GetPropertyChild("localMaterialBuffer")?.GetPropertyChild("materials")?.RecalculateProperties();
+        cvm.GetPropertyChild("localMaterialBuffer")?.RecalculateProperties();
+        cvm.GetPropertyChild("externalMaterials")?.RecalculateProperties();
+    }
+
+    public static bool HasPreloadMaterials(ChunkViewModel cvm) =>
+        cvm.GetRootModel() is { ResolvedData: CMesh mesh } && HasPreloadMaterials(mesh);
+
+    private static bool HasPreloadMaterials(CMesh mesh) =>
+        mesh.PreloadExternalMaterials.Count > 1 || mesh.PreloadLocalMaterialInstances.Count > 0;
```
notaspirit: why is this 1?

WolvenKit.App/ViewModels/Tools/CvmMaterialTools.cs
```diff
@@ -0,0 +1,789 @@
...
+        {
+            keepLocal = mesh.LocalMaterialBuffer.Materials
+                .Where((material, i) => localMatIdxList.Contains(i)).ToArray();
+        }
+
+        if (mesh.ExternalMaterials is not null && mesh.ExternalMaterials.Count > 0)
+        {
+            keepExternal = mesh.ExternalMaterials
+                .Where((material, i) => externalMatIdxList.Contains(i)).ToArray();
+        }
```
notaspirit: would prefer null conditional operator in combination with a null coalescing operator to oneline each of these

notaspirit (changes requested):
Generally looks good, flagged a couple of smaller to medium things in CvmMaterialTools, and considering that CvmMaterialTools isn't a ViewModel it shouldn't be in the ViewModel directory, services or helpers would be a better fit.

Issue #1361, opened 2023-09-11
GLB import fails with Value was either too large or too small for an Int16


gewfsmc:
Hi,

When importing GLB I get this error:
```
Message: Value was either too large or too small for an Int16.
Source: System.Private.CoreLib
StackTrace:    at System.Convert.ThrowInt16OverflowException()
   at System.Convert.ToInt16(Int32 value)
   at WolvenKit.Modkit.RED4.ModTools.RawMeshToRE4Mesh(RawMeshContainer mesh, Vector4 qScale, Vector4 qTrans) in D:\a\WolvenKit\WolvenKit\WolvenKit.Modkit\RED4\Tools\MeshImportTools.cs:line 693
   at WolvenKit.Modkit.RED4.ModTools.<>c__DisplayClass67_0.<ImportMesh>b__1(RawMeshContainer _) in D:\a\WolvenKit\WolvenKit\WolvenKit.Modkit\RED4\Tools\MeshImportTools.cs:line 445
   at System.Linq.Enumerable.SelectListIterator`2.ToList()
   at WolvenKit.Modkit.RED4.ModTools.ImportMesh(FileInfo inGltfFile, Stream inMeshStream, GltfImportArgs args) in D:\a\WolvenKit\WolvenKit\WolvenKit.Modkit\RED4\Tools\MeshImportTools.cs:line 481
   at WolvenKit.Modkit.RED4.ModTools.ImportGltf(RedRelativePath rawRelative, DirectoryInfo outDir, GltfImportArgs args) in D:\a\WolvenKit\WolvenKit\WolvenKit.Modkit\RED4\Import.cs:line 487
```

mesh has weights, is triangulated and has UV. Mesh can be found: [Mesh](https://www.file.io/niPR/download/W3IYfEeYb1kS)

effs:
This is a float/double precision problem in quantization, values at or very close to min or max extents can go out of range by a very small amount because floating point math.

The quick and probably correct fix is to clamp the conversions (the precision loss is < 0.1 mm, need to write it down exactly), but it's possible at least some cases we could alternatively compute with double precision instead, or explicitly rely on an epsilon.

(OP's file is now fixed but the problem itself remains.)

PR #2789, opened 2026-01-18
math.min-ing to prevent out of bounds error

manavortex:
# mesh import: math.min-ing to prevent out of bounds error

Unsure what the risks of this are - it works for a static mesh

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
@@ -772,9 +772,9 @@ namespace WolvenKit.Modkit.RED4
                 var x = qScale.X != 0 ? (mesh.positions[i].X - qTrans.X) / qScale.X : 0;
                 var y = qScale.Y != 0 ? (mesh.positions[i].Y - qTrans.Y) / qScale.Y : 0;
                 var z = qScale.Z != 0 ? (mesh.positions[i].Z - qTrans.Z) / qScale.Z : 0;
-                re4Mesh.ExpVerts[i, 0] = Convert.ToInt16(Math.Max(-1, Math.Min(x, 1) * short.MaxValue));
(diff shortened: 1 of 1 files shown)


notaspirit:
32767 is short.MaxValue

In that case it'd be better to at least investigate where it is coming from even if it doesn't resolve in a concrete answer as otherwise it's likely patching symptoms with unknown side effects.

notaspirit:
Looking at older issues on it seems like #1361 could be related.

manavortex:
Shit, no - not anymore, sorry. But I'm sure you can find someone in mod dev chat who still has the problem?

WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
```diff
@@ -772,9 +772,9 @@ private static Re4MeshContainer RawMeshToRE4Mesh(RawMeshContainer mesh, Vec4 qSc
                 var x = qScale.X != 0 ? (mesh.positions[i].X - qTrans.X) / qScale.X : 0;
                 var y = qScale.Y != 0 ? (mesh.positions[i].Y - qTrans.Y) / qScale.Y : 0;
                 var z = qScale.Z != 0 ? (mesh.positions[i].Z - qTrans.Z) / qScale.Z : 0;
-                re4Mesh.ExpVerts[i, 0] = Convert.ToInt16(x * 32767);
-                re4Mesh.ExpVerts[i, 1] = Convert.ToInt16(y * 32767);
-                re4Mesh.ExpVerts[i, 2] = Convert.ToInt16(z * 32767);
+                re4Mesh.ExpVerts[i, 0] = Convert.ToInt16(Math.Max(-1, Math.Min(x, 1) * short.MaxValue));
```
notaspirit: use `Math.Clamp()`, that's cleaner

notaspirit (changes requested):
other than that looks good

notaspirit (approved):

PR #2794, opened 2026-01-20
prevent file names from containing ()

manavortex:
# prevent file names from containing ()


notaspirit:
What does it fix / where were they an issue before? Filepath wise they are valid characters.

notaspirit:
What's the reason for not wanting them in the archives? Does it break something? If so what?

notaspirit:
Makes sense. Since we have a defined range of valid characters it would be better to make it a white list then, both for functionality and readability.

notaspirit (approved):

PR #2798, opened 2026-01-21
extract embedded files to their relative filepath

### LINKED ISSUE #1746: export embedded to filepath
the new right click extract embedded is great, but the embedded files know where they think they were, can we have an option that just exports it to that filename (maybe only if it doenst exist already?) as at the minute you can see the path in the menu, but cant copy it, and have to try remember/recreate it with the save dialog.

Extract to filepath as an option in the menu.

notaspirit:
# extract embedded files to their relative filepath

**Implemented:**
- save embedded file to the archive section at it's relative file path when extracting, use dialog (previous behavior) when project is not available

**Additional notes:**
closes #1746

### DIFF
--- a/WolvenKit.App/ViewModels/Dialogs/ExtractEmbeddedFileDialogViewModel.cs
+++ b/WolvenKit.App/ViewModels/Dialogs/ExtractEmbeddedFileDialogViewModel.cs
@@ -12,7 +12,7 @@ public partial class ExtractEmbeddedFileDialogViewModel : DialogViewModel
     private string _embeddedFilePath;
     [ObservableProperty]
     private string _newFilePath;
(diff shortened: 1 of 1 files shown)


notaspirit:
The first approach as in integrating the `File Already Exists` and `Filepath Unchanged` popups into the main dialog? 

For the popup about the unchanged file a notification might actually be better true, considering nothing bad, irreversible is going to happen immediately and the user already made the choice to keep it the same despite a warning. 
The popup about the file already existing should stay a popup though since that actually requires a the user to make a choice.

notaspirit:
Adjusted the unchanged path to now just be a warning notification (don't think logging is necessary for this)

PR #2799, opened 2026-01-22
CVM: entEntityParameterBuffer

manavortex:
# CVM: entEntityParameterBuffer
Adds descriptors and values for entEntityParameterBuffer in .app file, and handles default expansion states

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.Value.cs
+++ b/WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.Value.cs
@@ -472,8 +472,15 @@ public partial class ChunkViewModel
 
             #region mesh
             case meshMeshAppearance { ChunkMaterials: not null } appearance:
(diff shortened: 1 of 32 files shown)


WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.Value.cs
```diff
@@ -433,6 +434,43 @@ when compOverride.MeshAppearance.GetResolvedText() is string s:
 
                 IsValueExtrapolated = Value != string.Empty;
                 break;
+            case entEntityParametersBuffer buf
+                when buf.ParameterBuffers.Count > 0 && buf.ParameterBuffers[0] is
```
notaspirit: ary and o variable names are unclear

manavortex: does it really matter for a switch case? they just need to be unique

PR #2809, opened 2026-01-31
fix: RecalculateSectionDuration() should never decrease duration

misterchedda:
**Fixed:**
https://discord.com/channels/717692382849663036/803201431657250857/1467008826074202307

### DIFF
--- a/WolvenKit.App/Services/TimelineService.cs
+++ b/WolvenKit.App/Services/TimelineService.cs
@@ -280,10 +280,7 @@ public partial class TimelineService : ObservableObject, IDisposable
         var newDuration = Math.Max(maxEndTime, currentDuration);
 
         SectionDuration = newDuration;
-        if (newDuration > currentDuration)
-        {
-            _sectionNode.SectionDuration = new scnSceneTime { Stu = newDuration };
-        }
+        _sectionNode.SectionDuration = new scnSceneTime { Stu = newDuration };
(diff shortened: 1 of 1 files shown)


WolvenKit.App/Services/TimelineService.cs
```diff
@@ -276,8 +276,14 @@ private void RecalculateSectionDuration()
...
-        SectionDuration = maxEndTime;
-        _sectionNode.SectionDuration = new scnSceneTime { Stu = maxEndTime };
+        var currentDuration = _sectionNode.SectionDuration?.Stu ?? 0;
+        var newDuration = Math.Max(maxEndTime, currentDuration);
+
+        SectionDuration = newDuration;
+        if (newDuration > currentDuration)
+        {
+            _sectionNode.SectionDuration = new scnSceneTime { Stu = newDuration };
+        }
```
notaspirit: this can lead to SectionDuration and the duration within _sectionNode to be out of sync, not sure if it can cause an issue though. Just wanted to point that out.

MrBilL61:
> Like previously I can only review the code side of it. Someone else needs to do the functionality test. @MrBilL61 do you have time for that?

I can test it tomorrow

MrBilL61:
Did some testing in multiple section nodes in couple of scenes - fix works as expected.

PR #2816, opened 2026-02-06
scnChoiceNodeOption: better value

manavortex:
# scnChoiceNodeOption: better value

@misterchedda @MrBilL61 can one of you make it show the real thing like in the editor, or would it have to parse files for that?


notaspirit:
https://discord.com/channels/717692382849663036/717720094196760760/1477749475333242934
image

PR #2819, opened 2026-02-08
ux: selection dialog

manavortex:
# ux: selection dialog

Do not select anything by default
If ctrl key is held while opening, re-use last filter


notaspirit:
sure then just drop / revert the ctrl behavior commit and I can approve that.

PR #2822, opened 2026-02-09
cvm: dropdown for entPhysicalMeshComponent

manavortex:
# fixed: dropdowns for properties of entPhysicalMeshComponent

To verify, add one to an .app or .ent file, or use the "generate props" dialogue

### DIFF
--- a/WolvenKit.App/Helpers/CvmDropdownHelper.cs
+++ b/WolvenKit.App/Helpers/CvmDropdownHelper.cs
@@ -864,6 +864,14 @@ public abstract class CvmDropdownHelper
 
             #endregion
 
+            #region morphtarget
+
+            case MorphTargetMesh when (parent.Name is "baseMesh" && cvm.Name == "DepotPath") || cvm.Name is "baseMesh":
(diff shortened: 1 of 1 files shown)


notaspirit (changes requested):
the drop down with the options is empty by default, manually clicking the refresh button properly populates it though. 

also add this change to the unreleased section of the changelog.txt

manavortex:
because the dropdown should have worked for these types already, I just forget to add them. 

I'm calling it scope creep because this PR adds a few switch cases to already existing application logic which has been set up to support them already, and they were just missing. Adding auto-population is not yet implemented or supported. I had looked into it twice already, and haven't found a good way to make it happen. 

From my perspective, this PR Is "enable dropdown support for two things where it wasn't being triggered before (quick fix)". What you want is "add previously unsupported application logic to component". One thing is done and ready for merging, the other is something I don't even know how to add yet, and which will have to wait until either I have the mental capacity to do it (which will be Saturday at the earliest and might be "whenever"), or until somebody else does it.

PR #2826, opened 2026-02-11
refactor: cvm material tools

manavortex:
# refactor: cvm material tools

using a proper interface now and put everything in one place

### DIFF
--- a/WolvenKit.App/WolvenKit.App.csproj
+++ b/WolvenKit.App/WolvenKit.App.csproj
@@ -23,7 +23,6 @@
     <None Remove="ViewModels\Shared\**" />
     <Page Remove="ViewModels\Others\**" />
     <Page Remove="ViewModels\Shared\**" />
-    <None Remove="WolvenKit.App.csproj.DotSettings" />
   </ItemGroup>
 
   <!-- Suppress compiler warnings for generated files -->


WolvenKit.App/Helpers/ChunkViewModel/CvmTools.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System.Collections.Generic;
```
notaspirit: is the reasoning for having CvmTools be a wrapper to have an umbrella style api that consolidates multiple distinct implementation responsibilities into one clean end point?

WolvenKit.App/Helpers/ChunkViewModel/CvmDependencyTools.cs
```diff
@@ -0,0 +1,148 @@
...
+                    Regenerate(ary);
+                }
+
+                return;
+            default:
+                _notificationService.Warning(
+                    "Can't regenerate visual controllers. Please check the log for more detail.");
+                _loggerService.Warning(
+                    $"Failed to regenerate visual controllers on {cvm.ResolvedData.GetType().Name}. " +
+                    "Select one or more appearances, the appearances array, or the root node.");
```
notaspirit: should be error level

PR #2827, opened 2026-02-11
Feature: flatten mi chain

manavortex:
# flatten mi chain
 
will resolve all templates, putting all properties directly into the local material
requires https://github.com/WolvenKit/WolvenKit/pull/2826


WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -778,16 +781,35 @@ public static async Task<HashSet<ResourcePath>> GetMaterialRefsFromFile(ChunkVie
...
             return -1;
         }
 
-        return array.ToList()
-            .Where(m => m.IsLocalInstance == isLocalInstance)
-            .Max(m => m.Index);
+        return FindHighestMaterialIndex(array, isLocalInstance);
+    }
+
+    private int FindHighestMaterialIndex(CArray<CMeshMaterialEntry> array, bool isLocalInstance)
```
notaspirit: variable name ("array") should be more descriptive like in the public method

WolvenKit.RED4/Types/Helper/CKeyValuePairFactory.cs
```diff
@@ -0,0 +1,47 @@
+﻿namespace WolvenKit.RED4.Types;
+
+public static class CKeyValuePairFactory
+{
+    public static CKeyValuePair Create(CMaterialParameter param)
+    {
+        var parameterName = param.ParameterName.GetResolvedText() ?? "invalid";
+        ;
```
notaspirit: this semicolon doesn't do anything

WolvenKit.App/ViewModels/Documents/RedDocumentViewToolbarModel.cs
```diff
@@ -654,6 +660,25 @@ private bool CanConvertFromPreloadMaterials() => RootChunk?.ResolvedData is CMes
...
     private void ConvertFromPreloadMaterials() => _cvmTools.ConvertMaterialsFromPreload(RootChunk);
 
+    private bool CanFlattenMiChain() => RootChunk?.ResolvedData is (CMesh or CMaterialInstance);
+
+    [RelayCommand(CanExecute = nameof(CanFlattenMiChain))]
+    private void FlattenMiChain()
+    {
+        var chunk = SelectedChunk ?? RootChunk;
+
+        if (chunk?.Tab?.Parent.IsDirty != false)
```
notaspirit, line 670: using double negative is unnecessarily convoluted, just do `== true`

also why is this necessary? It's really annoying UX wise.

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -833,4 +855,229 @@ public void AddTagsToMeshAppearances(List<ChunkViewModel> chunks, List<string> t
 
         appearanceChunks.First().Tab?.Parent.SetIsDirty(true);
     }
+
+    private static readonly Dictionary<string, List<CKeyValuePair>> s_materialValuesByPath = [];
+
+    /// <summary>
+    /// Reads a material's properties into a map for quick lookup. This includes shader properties and parameters.
```
notaspirit: should clarify which map as it only returns a list of strings

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -833,4 +855,229 @@ public void AddTagsToMeshAppearances(List<ChunkViewModel> chunks, List<string> t
...
+    /// </summary>
+    /// <param name="relativeBasePath"></param>
+    /// <param name="archiveManager"></param>
+    /// <param name="previousPaths">Array of previous paths - will append current base material and return</param>
+    /// <returns></returns>
+    private List<string> ReadMaterialValuesRecursive(ResourcePath relativeBasePath, IAppArchiveManager archiveManager,
```
notaspirit: should be using ref for the resourcepath as it can be reassgined as part of this method

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -833,4 +855,229 @@ public void AddTagsToMeshAppearances(List<ChunkViewModel> chunks, List<string> t
...
+            switch (file.RootChunk)
+            {
+                case CMaterialInstance matInstance:
+                    s_materialValuesByPath[relPath] = matInstance.Values.ToList();
+                    relativeBasePath = matInstance.BaseMaterial.DepotPath;
+                    continue;
+
+                case CMaterialTemplate shader:
+                {
+                    var shaderParamList = shader.Parameters[^1];
```
notaspirit: what's the reasoning behind always getting the last entry without length checks?

notaspirit (changes requested):
doesn't seem to work for the entire mesh, only when selecting specific material instances despite the code implying that it should work for the entire mesh at some points and the positioning in the "global" drop down implying the same.
Tested on `ep1\characters\garment\npc_instances\inst_paradise_wa\t2_127_wa_jacket__evelyn_fur_inst.mesh`

I also don't quite understand why it needs to be so complex, wouldn't it conceptually just need to check if the material path points to an mi, if it is look that file up and set all the missing properties in the root material instance, then do that recursively until the current mi points to a non mi base file?

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
manavortex: there are no shaders with empty parameters, I hope.

WolvenKit.App/ViewModels/Documents/RedDocumentViewToolbarModel.cs
manavortex: But I am fine with it being undefined. That's why I wrote the check like this.

notaspirit: then add an explicit null check, would be a lot easier to read

manavortex: I find that perfectly readable and it's valid C# syntax ¯\_(ツ)_/¯

notaspirit: It takes more effort to read and fully understand this statement compared to explicitly checking for null and false like e.g. `chunk?.Tab?.Parent.IsDirty is not bool isDirty || !isDirty`.

You also didn't address the second question about it's necessity. In the future please let the person who opened the comment mark it as resolved as you can't know if the party bringing it up also considers it resolved.

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
notaspirit: There is at least one I found from clicking around `.mt` files: `base\fx\_shaders\invisible.mt`

notaspirit: Ah nvm this is indexing the first level of the nested arrays (the one that is / should(?) always have a length of 3), should be fine then.

WolvenKit.App/ViewModels/Documents/RedDocumentViewToolbarModel.cs
manavortex: Sorry, but I find your version less readable, as it requires me to parse two conditionals in my head rather than one.

As for the necessity: I don't understand the question. What is annoying about it?

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
manavortex: Can't do that because 
> Property 'BaseMaterial' access returns temporary value. Cannot modify struct member when accessed struct is not classified as a variable
Introduced a temporary variable instead

WolvenKit.App/ViewModels/Documents/RedDocumentViewToolbarModel.cs
notaspirit: It's annoying because if you e.g. want to just flatten two out of X materials (or if you have any other previous changes), you need to flatten one, then save then flatten the other when no other workflow requires that where you can stack the changes until you as the user decide it's time to save.

manavortex: I hadn't thought of that since I was mostly thinking about NPVs (and skin tones). Why'd a user want to flatten only one or two materials, and not everything?

notaspirit: For example you have some materials that are using your own custom mi files and some that use vanilla ones where you want to simplify the vanilla chain but keep your custom mi files. 
You wouldn't want to also flatten your custom mi files as you have made them for a reason, and the vanilla ones may be too convoluted / opaque for your use case.

manavortex: Seems convoluted to me, but you're super right - we're talking about users here, and somebody will come up with even edgier cases. I just need to repair the logic.

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -778,16 +781,35 @@ public static async Task<HashSet<ResourcePath>> GetMaterialRefsFromFile(ChunkVie
...
+        // ReSharper disable once LoopCanBeConvertedToQuery - will throw NoElementsException
+        for (var i = 0; i < matDefArray.Count; i++)
+        {
+            if (matDefArray[i].IsLocalInstance == isLocalInstance)
+            {
+                indices.Add(matDefArray[i].Index);
+            }
+        }
+
+        return indices.Count == 0 ? -1 : indices.Max();
```
notaspirit: the creation of the array for the indicies is redundent, you can init an int with -1 and have the loop overwrite the index if it's higher, that way you also remove the if statement in the return

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -833,4 +855,255 @@ public void AddTagsToMeshAppearances(List<ChunkViewModel> chunks, List<string> t
...
+
+        // failed to read file
+        if (archiveManager.GetCR2WFile(relPath, true, true) is not CR2WFile file)
+        {
+            return;
+        }
+
+        switch (file.RootChunk)
+        {
+            case CMaterialInstance matInstance:
```
notaspirit: why does one have braces and the other doesn't?

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -833,4 +855,255 @@ public void AddTagsToMeshAppearances(List<ChunkViewModel> chunks, List<string> t
...
+            }
+
+            materialProperties.AddRange(consolidatedValues.Where(v =>
+                !materialProperties.Contains(v)
+            ));
+
+            inst.BaseMaterial = new CResourceReference<IMaterial>(shaderPath);
+            inst.Values.Clear();
+
+            // can't LINQ here because CArray doesn't like it
```
notaspirit: use reasoning in the comment, "CArray doesn't like it" is very vague and creates more answers than it answers

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -833,4 +855,255 @@ public void AddTagsToMeshAppearances(List<ChunkViewModel> chunks, List<string> t
...
+        foreach (var relPath in baseMaterialChain.Where(relPath => relPath.EndsWith(".mi")))
+        {
+            var properties = s_materialValuesByRelPath.TryGetValue(relPath, out var value) ? value : [];
+            values.AddRange(properties.Where(v => !values.Contains(v)));
+        }
+
+        return values;
+    }
+
+    private void ConvertExternalMaterials(CMesh mesh)
```
notaspirit: what's this doing here as part of this pr if it's never used?

notaspirit (changes requested):
In addition to the comments, the point about the file needing to be saved is unresolved and you haven't explained the reasoning behind building the full tree of materials and values when you're discarding most of it afterwards anyway

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
manavortex: How, exactly? CArrays and linq do not play nice with each other, so I'd rather avoid that

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
manavortex: Because that's how our coding conventions want it. The braces around the switch param are mandatory, braces in the case would be redundant here

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
notaspirit: you wouldn't use any linq it'd just be 

```csharp
var maxIndex = -1;
// ReSharper disable once ForCanBeConvertedToForeach - can't LINQ here
// ReSharper disable once LoopCanBeConvertedToQuery - will throw NoElementsException
for (var i = 0; i < matDefArray.Count; i++)
{
    if (matDefArray[i].IsLocalInstance == isLocalInstance && matDefArray[i].Index > maxIndex)
    {
        maxIndex = matDefArray[i].Index;
    }
}

return maxIndex;
```

manavortex: .. ok, fair, that'd do the trick :D

manavortex:
At some point after the refactor:tm:, I want to expand file validation to check for invalid material properties - then I'll need this.

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
notaspirit: both paths braces are technically redundant except for scoping variables, but as both declare variables they should have equal treatment.

manavortex: They don't? The first one switches?

notaspirit: first case defines the variable `parentPath` in the second line of the case, second case defines the variable `shaderParamList` in it's first line.

manavortex: Oh, now I get what you were talking about, I'll fix

notaspirit:
Regarding the overall approach complexity point in the first review, I've had to implement material flattening in a different project where I implemented it like this. It doesn't need to keep track of where the material property came from as it's using `TryAdd` which will only add the entry if isn't already in the dict which is the correct approach given that the chain is read top to bottom. 

`CMaterialParameterScalar` when read from the `CMaterialTemplate` I'm not 100% sure about as that also has a min and max property but it doesn't seem like the `CMaterialInstance` can set those from it's value array. 

Feel free to base an implementation for wkit on this. 

```CSharp
 private record FlatMaterial(string BaseMaterial, Dictionary<string, object> Properties);
 private FlatMaterial GetFlattenedMaterial(IMaterial material, string basePath)
    {
        var baseMaterial = basePath;
        Dictionary<string, object> properties = new();
        
        var currentMat = material;
        
        while (true)
        {
            switch (currentMat)
            {
                case CMaterialInstance matInstance:
                {
                    foreach (var kvp in matInstance.Values)
                    {
                        string? name = kvp.Key;
                        if (name is not null)
                            properties.TryAdd(name, kvp.Value);
                    }
                    
                    if (matInstance.BaseMaterial.DepotPath.GetString() is not { } baseMaterialPath)
                    {
                        goto breakOuter;
                    }
                    
                    var baseMatRc = _archiveManager.GetCR2WFile(baseMaterialPath)?.RootChunk;
                    
                    if (baseMatRc is not IMaterial baseMat)
                    {
                        goto breakOuter;
                    }

                    currentMat = baseMat;
                    baseMaterial = baseMaterialPath;
                    break;
                }
                case CMaterialTemplate matTemplate:
                {
                    var values = matTemplate.Parameters[2];
                    foreach (var matParam in values)
                    {
                        if (matParam.Chunk is null)
                            continue;

                        string? name = matParam.Chunk.ParameterName;
                        object? value = matParam.Chunk switch
                        {
                            CMaterialParameterColor mpc => mpc.Color,
                            CMaterialParameterCpuNameU64 mpcnu => mpcnu.Name,
                            CMaterialParameterCube mpcu => mpcu.Texture,
                            CMaterialParameterDynamicTexture mpdt => mpdt.Texture,
                            CMaterialParameterFoliageParameters mpfp => mpfp.FoliageProfile,
                            CMaterialParameterGradient mpg => mpg.Gradient,
                            CMaterialParameterHairParameters mphp => mphp.HairProfile,
                            CMaterialParameterMultilayerMask mpml => mpml.Mask,
                            CMaterialParameterMultilayerSetup mpms => mpms.Setup,
                            CMaterialParameterScalar mps => mps.Scalar,
                            CMaterialParameterSkinParameters mpsp => mpsp.SkinProfile,
                            CMaterialParameterStructBuffer => null,
                            CMaterialParameterTerrainSetup mpts => mpts.Setup,
                            CMaterialParameterTexture mpt => mpt.Texture,
                            CMaterialParameterTextureArray mpta => mpta.Texture,
                            CMaterialParameterVector mpv => mpv.Vector,
                            _ => null
                        };
                        
                        if (name is not null && value is not null)
                            properties.TryAdd(name, value);
                    }
                    
                    goto breakOuter;
                }
                default:
                {
                    Console.WriteLine($"Material {currentMat} is not a material instance or material template!");
                    goto breakOuter;
                }
            }
            
            continue;
            
            breakOuter:
            break;
        }
        
        return new FlatMaterial(baseMaterial, properties);
    }
```

manavortex:
Yeah ok, caching makes no sense with this approach

changelog/unreleased/2827.yaml
```diff
@@ -0,0 +1,6 @@
+changes:
+    - type: "fixed"
```
notaspirit: why "fixed"?

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -908,4 +917,151 @@ public static IRedType UnEmbedResourceReference(IRedType cvpValue)
...
+    private void FlattenMaterial(CMaterialInstance material, IAppArchiveManager archiveManager)
+    {
+        var flatMaterial = GetFlattenedMaterial(material, material.BaseMaterial.DepotPath.GetString() ?? "", archiveManager);
+        material.BaseMaterial = new CResourceReference<IMaterial>(flatMaterial.BaseMaterial, InternalEnums.EImportFlags.Default);
+        material.Values.Clear();
+        foreach (var cKeyValuePair in flatMaterial.Properties.Values)
+        {
+            material.Values.Add(cKeyValuePair);
+        }
+    }
```
notaspirit: Why write a wrapper for the example method I provided? 

You don't need the FlatMaterial record at all, you can directly write everything inplace to the CMaterialInstance you're flattening.

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -908,4 +917,151 @@ public static IRedType UnEmbedResourceReference(IRedType cvpValue)
...
+                            continue;
+                        }
+                        properties.TryAdd(matParam.Chunk.ParameterName!, CKeyValuePairFactory.Create(matParam.Chunk));
+                    }
+
+                    goto breakOuter;
+                }
+                default:
+                {
+                    Console.WriteLine($"Material {currentMat} is not a material instance or material template!");
```
notaspirit: this should be using the logger not console

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -908,4 +917,151 @@ public static IRedType UnEmbedResourceReference(IRedType cvpValue)
...
+                case CMaterialTemplate matTemplate:
+                {
+                    var values = matTemplate.Parameters[2];
+                    foreach (var matParam in values)
+                    {
+                        if (matParam.Chunk?.ParameterName.GetResolvedText() is not string name || string.IsNullOrEmpty(name) )
+                        {
+                            continue;
+                        }
+                        properties.TryAdd(matParam.Chunk.ParameterName!, CKeyValuePairFactory.Create(matParam.Chunk));
```
notaspirit: since you're already checking with an out variable above use that directly.

WolvenKit.App/Helpers/ChunkViewModel/CvmTools.cs
```diff
@@ -40,6 +42,14 @@ public void AddMaterialAndDefinition(ChunkViewModel cvm, string newName) =>
     public void AddTagsToMeshAppearances(List<ChunkViewModel> chunks, List<string> tagList) =>
         _cvmMaterialTools.AddTagsToMeshAppearances(chunks, tagList);
 
+    public void FlattenMiChain(List<ChunkViewModel> cvmSelection, IAppArchiveManager archiveManager, Cp77Project? project) => throw new System.NotImplementedException();
```
notaspirit: what's this for if it just throws?

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -778,15 +777,25 @@ public static async Task<HashSet<ResourcePath>> GetMaterialRefsFromFile(ChunkVie
...
-        if (materialEntries.Count == 0)
+        if (materialDefinitionArray?.ResolvedData is not CArray<CMeshMaterialEntry> array || array.Count == 0 )
         {
             return -1;
         }
 
-        var existingIndices = materialEntries
-            .Select(m => m.IsLocalInstance == isLocalMaterial ? m.Index : -1)
+        return FindHighestMaterialIndex(array, isLocalMaterial);
+    }
```
notaspirit: this overload is never used

notaspirit (changes requested):
The core functionality works, but the ux can be a bit rough. If you don't have a fitting property selected nothing happens. 
It should either do it on the entire mesh when a specific CMaterialInstance isn't selected, be moved to the context menu where visibility is controlled by the availability of the action, or at the very least show an error notificiation.

What is the purpose of injecting the IAppArchiveManager as a method parameter everywhere? I don't see any obvious circular dependency or anything like that which would stop you from adding the archive manager as a dependency to cvm tools as a whole. 

For everything else I've left code comments.

WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
manavortex: because I found it way easier to read and to debug that way.

notaspirit:
Looks good, just the UI/UX changes haven't been addressed yet.

manavortex:
But I did, if there's any node selected in a CMesh that's not covered by one of the previous cases, it'll run the operation on parent

notaspirit:
Now I see it, though the display value doesn't correctly update when run on a different property until you actually click on it. 
Is that something that is easily fixable here?

notaspirit:
Noticed that it now it doesn't mark the tab as dirty after the changes, running it on an arbitrary property now does refresh. When the root CMesh is selected it doesn't seem to resolve the mi files

notaspirit (approved):

PR #2832, opened 2026-02-13
exclude .bin files from project wide validation

manavortex:
# exclude .bin files from project wide validation

These are temp files or something - and they're spamming the log


WolvenKit.App/ViewModels/Shell/AppViewModel.WScript.cs
```diff
@@ -161,7 +161,7 @@ private async Task RunFileValidationOnProject()
 
         foreach (var file in filesToValidate)
         {
-            var fileExtension = file.Extension.TrimStart('.').ToLower();
```
notaspirit: why change it to include the dot when in the same method in the next loop it excludes the dot? should be consistent within the method.

Issue #2813, opened 2026-02-03
"Index was outside the bounds of the array" error on exporting vanilla .anims


Maffaxxx:
Hi all. I tried searching for a solution to this on the modding discord, but either people try again some time later and it solved by itself or the solution provided doesnt apply to my situation. 

I am trying to export a vanilla .anims file to Blender, but WK wont let me (this is the 26/03/02 8.17.2 nightly verbose log, but the same happened on the 8.17.1 stable): 

```[03/02/2026 19:07:18] [Important] synced__mq055_female_average.anims: Found 34 animations to export, using rig base\characters\base_entities\woman_base\woman_base.rig```
```[03/02/2026 19:07:18] [Important] synced__mq055_female_average.anims: Root motion export: disabled```
```[03/02/2026 19:07:18] [Debug    ] Exporting SIMD animation synced__lean_wall_female_kiss_player__01_female_average with 13870 S/R/T transforms and 2470 tracks```
```[03/02/2026 19:07:18] [Error    ] base\animations\quest\minor_quests\mq055\synced\synced__mq055_female_average.anims - Index was outside the bounds of the array.```
```[03/02/2026 19:07:18] [Error    ] Message: Index was outside the bounds of the array.
Source: WolvenKit.Modkit
StackTrace:    at WolvenKit.Modkit.RED4.Animation.SIMD.DecodeSimdAnimationData(AnimationBufferData& result, ModelRoot& model, animAnimationBufferSimd& animBufSimd, List`1& animDataBuffers, String animName, animAnimation animAnimDes, ILoggerService _loggerService)
   at WolvenKit.Modkit.RED4.ModTools.GetAnimation(CR2WFile animsFile, CR2WFile rigFile, String animsFileName, ModelRoot& gltfModel, Boolean includeRig, Boolean additiveAddRelative, Boolean incRootMotion)
   at WolvenKit.Modkit.RED4.ModTools.ExportAnim(CR2WFile animsFile, FileInfo outfile, Boolean isGLBinary, Boolean additiveAddRelative, Boolean incRootMotion, ValidationMode vmode)
   at WolvenKit.Modkit.RED4.ModTools.InternalUncookBuffers(CR2WFile cr2wFile, String relPath, FileInfo outfile, GlobalExportArgs settings, DirectoryInfo rawOutDir)```
```[03/02/2026 19:07:18] [Error    ] The following items failed:
[...]

Issue #2631, opened 2025-09-07
Error Exporting .anims


lulzx-feilis:
I've encountered 2 different errors exporting .anims files occured in WolvenKit.Modkit.RED4.ModTools.GetAnimation and WolvenKit.Modkit.RED4.Animation.SIMD.DecodeSimdAnimationData, I've already tried the solution of https://github.com/WolvenKit/WolvenKit/commit/cf4e3737f7544433c5cffdb0554eff1703e29bac, but no luck

[2025-09-07 17:23:38] [Error    ] base\animations\npc\generic_characters\female_average\open_world\generic_average_female_work.anims - Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
[2025-09-07 17:23:39] [Error    ] Message: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
Source: System.Private.CoreLib
StackTrace:    at System.Collections.Generic.List`1.get_Item(Int32 index)
   at SharpGLTF.Schema2.Skin.GetJoint(Int32 idx)
   at WolvenKit.Modkit.RED4.ModTools.GetAnimation(CR2WFile animsFile, CR2WFile rigFile, String animsFileName, ModelRoot& gltfModel, Boolean includeRig, Boolean additiveAddRelative, Boolean incRootMotion)
   at WolvenKit.Modkit.RED4.ModTools.ExportAnim(CR2WFile animsFile, FileInfo outfile, Boolean isGLBinary, Boolean additiveAddRelative, Boolean incRootMotion, ValidationMode vmode)
   at WolvenKit.Modkit.RED4.ModTools.InternalUncookBuffers(CR2WFile cr2wFile, String relPath, FileInfo outfile, GlobalExportArgs settings, DirectoryInfo rawOutDir)

[2025-09-07 17:23:39] [Error    ] base\animations\player\interactive_scene\player_male__stand__2h_on_sides.anims - Index was outside the bounds of the array.
[2025-09-07 17:23:39] [Error    ] Message: Index was outside the bounds of the array.
Source: WolvenKit.Modkit
StackTrace:    at WolvenKit.Modkit.RED4.Animation.SIMD.DecodeSimdAnimationData(AnimationBufferData& result, ModelRoot& model, animAnimationBufferSimd& animBufSimd, List`1& animDataBuffers, String animName, animAnimation animAnimDes, ILoggerService _loggerService)
[...]

Issue #1453, opened 2023-11-27
Anims: SIMD Buffers with Odd NumExtraJoints Fail to Export


effs:
Something's off with the alignment or count calculation when the animation has (at least) an odd `NumExtraJoints`, so the seek position becomes incorrect at some point (TBD) and we're reading semantic garbage from there on out.

### Repro

1. Try to export `base\animations\cyberware\mantisblade\cw_mantisblade_finisher.anims` (fails on `finisher_mantis_blades_01`)

'Good first issue'

kmirea:
Heya

What is the total bonecount, extra joints, extra tracks (seperately).  Just off top of my head if u short on ur 4x4, maybe could populate the remainder with zero vals / empty quat, i had similar issues on a side project, very much unrelated xD

notaspirit:
fixed in #2833

PR #2833, opened 2026-02-13
New SIMD animation encoder and decoder fixes.

### LINKED ISSUE #1443: Anims: Import SIMD Animations
Currently originally SIMD anims are converted to CompressedBuffer on import. Should be reasonably straightforward with the correct format defined in the export.

yabab:
Hey, I am working on an HMG mod and noticed the overheat animation was breaking when I attempted to re-import all of the animations. Stumbled upon the SIMD animation bug and attempted to fix it and I seem to be able to now export and re-import that animation going through a Blender roundtrip (something issue #1443 seemed to point out was not working) with additional changes to the Blender plugin (https://github.com/WolvenKit/Cyberpunk-Blender-add-on/pull/352) as well. I went into the issues and looked up SIMD and came upon the decoder issues and also attempted a fix. This PR might not be ready for prime time, but could get the ball rolling. Would appreciate it if someone more experienced with the tool could look at it.
Description:

Fix SIMD animation decoder out-of-bounds crashes, add SIMD encoder and preserve original quantization bit depth through the pipeline

Implemented:

SIMD animation encoder (AnimSIMDEncoder.cs) — enables .anims import for SIMD-compressed animation buffers
ForceCompressedEncoding option in GltfImportArgs to select SIMD encoding on import
Import routing in [AnimationTools.cs](vscode-file://vscode-app/c:/Users/ronis_1xgq/AppData/Local/Programs/Microsoft%20VS%20Code/b6a47e94e3/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to dispatch to the new encoder
SimdQuantizationBits field threaded through AnimationBufferData, AnimationOptimizationHints, and glTF extras — decoder now preserves the original quantization bit depth (e.g. 12, 16, 20) from the game's animAnimationBufferSimd, and the encoder uses it on re-import instead of always assuming 16-bit
Fixed:

scalesForFrame array undersized when NumJoints exceeds SIMD-aligned count, causing IndexOutOfRangeException
Non-constant scale loop iterating by NumJoints instead of numJointsSimdAligned, reading past raw data boundaries
Scale write missing bounds guard for padding joints beyond NumJoints
evalIndices position write missing upper-bound check, crashing on animations with extra eval joints
copyIndices position write missing bounds check, crashing on animations with copy indices outside joint count

Additional notes:
Attempts to close #1443 , #1447 , #1453 , #2631 and #2813

### DIFF
--- a/WolvenKit.Common/Model/Arguments/ImportArgs.cs
+++ b/WolvenKit.Common/Model/Arguments/ImportArgs.cs
@@ -235,6 +235,17 @@ namespace WolvenKit.Common.Model.Arguments
         public bool AdditiveStripLocalTransform { get; set; } = true;
 
         /// <summary>
+        /// Force all animations to use Compressed encoding even if the source was SIMD.
+        /// When unchecked, animations exported from SIMD sources (or flagged as preferSIMD)
+        /// will be re-encoded as SIMD buffers, preserving the original format.
+        /// </summary>
+        [Category("Animation Settings")]
+        [Display(Name = "Force Compressed Encoding")]
+        [Description("When checked, all animations will use Compressed encoding regardless of their original format. Uncheck to preserve SIMD encoding for animations that were originally SIMD.")]
+        [UsedWith(nameof(ImportFormat), GltfImportAsFormat.Anims)]
+        public bool ForceCompressedEncoding { get; set; } = false;
+
+        /// <summary>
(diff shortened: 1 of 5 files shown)


manavortex (approved):
Presto says this is better than the current implementation, that's good enough for me

Issue #2791, opened 2026-01-19
feature: search in entire project


manavortex:
Would probably be easiest to read files as json and check if the match is in the file?

EDIT: notaspirit suggests recursive property search via reflection, which would be faster and also decouple things from the UI

PR #2867, opened 2026-03-15
feature: search in project

manavortex:
# feature: search in project

image

as per issue #2791 (which we should leave open until spirit can provide a better implementation)


WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -1399,4 +1400,61 @@ public List<string> GetAppearanceNames(string relativePath)
...
+        if (totalFiles == 0)
+        {
+            return [];
+        }
+
+        progressService.Report(0);
+        var processedFiles = 0;
+        var progressIncrement = totalFiles > 0 ? 100.0 / totalFiles : 100;
+
+        List<string> filesWithMatch = [];
```
gistya: Can we do this asynchronously on a background thread so the app doesn't freeze up for people that have 100,000 files etc.?

gistya (commented):
Code looks fine, but if you could move the CPU hot path to an asynchronous  background task that would be great

WolvenKit.App/Helpers/Cr2WTools.cs
```diff
@@ -109,6 +111,25 @@ public CR2WFile ReadCr2W(string absolutePath)
         return null;
     }
 
+    public string ReadJson(string absolutePath)
```
notaspirit: `ReadAsJson` would be more clear. As `ReadJson` implies it's reading a json file, not that it's reading a cr2w and returning the json serialized string.

WolvenKit.App/Helpers/StringHelper.cs
```diff
@@ -443,4 +443,70 @@ public static string Truncate(string text, int maxLength)
...
+    /// Check a string for a substring, considering search options regex and whole word
+    /// </summary>
+    /// <param name="input">String to be replaced in</param>
+    /// <param name="searchOrPattern">Search string or regex pattern</param>
+    /// <param name="isWholeWord">Search refinement: whole word only</param>
+    /// <param name="isRegex">Search refinement: regular expression</param>
+    /// <returns></returns>
+    public static bool StringContains(string? input, string searchOrPattern, bool isWholeWord, bool isRegex)
+    {
+        if (input is null)
```
notaspirit, line 492: the method should accept a non nullable string with the caller ensuring it's not null

WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -1399,4 +1401,64 @@ public List<string> GetAppearanceNames(string relativePath)
...
+            progressService.Report(currentProgress);
+
+            return ValueTask.CompletedTask;
+        });
+
+        await Parallel.ForEachAsync(project.ResourceFiles, (filePath, _) =>
+        {
+            try
+            {
+                var absolutePath = Path.Join(project.ResourcesDirectory, filePath);
```
notaspirit: handle the absolutepath variability consistently, either use a variable for both or for neither

WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -1399,4 +1401,64 @@ public List<string> GetAppearanceNames(string relativePath)
...
+            catch
+            {
+                _loggerService.Error($"Failed to open resources{Path.DirectorySeparatorChar}{filePath} for searching");
+            }
+            return ValueTask.CompletedTask;
+        });
+
+        progressService.IsIndeterminate = false;
+
+        return filesWithMatch.ToList().Distinct().ToList();
```
notaspirit: Pretty sure the distinct is redundant as the operating system already enforces that

WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -1399,4 +1401,64 @@ public List<string> GetAppearanceNames(string relativePath)
...
+                {
+                    filesWithMatch.Add($"resources{Path.DirectorySeparatorChar}{filePath}");
+                }
+
+                var currentProgress = Interlocked.Increment(ref processedFiles) * progressIncrement;
+                progressService.Report(currentProgress);
+            }
+            catch
+            {
+                _loggerService.Error($"Failed to open resources{Path.DirectorySeparatorChar}{filePath} for searching");
```
notaspirit, line 1474: What's the reason for the different error handling? 
Where for cr2w files you suppress the error and for the resource files you do explicitly handle and log them.

WolvenKit.App/Helpers/StringHelper.cs
```diff
@@ -443,4 +443,70 @@ public static string Truncate(string text, int maxLength)
...
+    /// <param name="replace">String to replace it with</param>
+    /// <param name="isWholeWord">Search refinement: whole word only</param>
+    /// <param name="isRegex">Search refinement: regular expression</param>
+    /// <returns></returns>
+    public static string ReplaceInString(string input, string searchOrPattern, string replace, bool isWholeWord,
+        bool isRegex)
+    {
+        if (isWholeWord)
+        {
+            if (input != searchOrPattern)
```
notaspirit, line 462: I don't understand this check, can you explain what it does and why searchOrPattern is a string?

WolvenKit.App/ViewModels/Shell/AppViewModel.cs
```diff
@@ -887,23 +887,39 @@ private void SaveAs()
...
+            {
+                if (_projectManager.ActiveProject is not { } project || string.IsNullOrEmpty(tab.FilePath))
+                {
+                    return true;
+                }
+
+                var filePath = project.GetRelativePath(tab.FilePath) ?? tab.FilePath;
+                return project.ModFiles.Contains(filePath, StringComparer.InvariantCultureIgnoreCase);
+            }).ToList();
+
```
notaspirit, line 926: how is this change related to the search feature? And what problem does it solve?

notaspirit (changes requested):
The way the search presents the results is a point of improvement, but it'll do for now. It'll probably need to be addressed together with the way file validation presents it's results as it does it in the same manner. 

Other than the comments looks good.

WolvenKit.App/Helpers/StringHelper.cs
manavortex: Why?

WolvenKit.App/Helpers/ProjectResourceTools.cs
manavortex: How sure is pretty in percentage? if it's <100, I'd leave it in.

WolvenKit.App/Helpers/StringHelper.cs
manavortex: "search for whole word" disables the regex checkbox (regex has its own "whole word" implementation). So if it's not equal, it's not a match.

WolvenKit.App/ViewModels/Shell/AppViewModel.cs
manavortex: Is related to "save all" before launching the project. I ran into an exception, so I figured I might as well fix it.

WolvenKit.App/Helpers/ProjectResourceTools.cs
notaspirit: It can't contain duplicates as the fs enforces that and Mod/Resource/RawFiles uses `Directory.EnumerateFiles` directly as the source of truth.

WolvenKit.App/Helpers/StringHelper.cs
notaspirit: But input is the entire string, so if you had e.g. input as "The quick brown fox jumps over the lazy dog" and searchOrPattern be "quick" with isWholeWord set to true it wouldn't match it despite it being a whole separated word. 
When `StringContains` in comparison builds a regex querry for that which will match the example above and thus create different results.

WolvenKit.App/Helpers/StringHelper.cs
notaspirit: It's not an optional parameter as it is essential to the function of the method and is likely to hide issues when a value is unexpectedly null.

The caller should have the responsibility of handling the null case, to keep the same behavior (false on null) the consumer simply needs to use `?? """` on the parameter which explicitly handles it.

notaspirit:
Another thing I noticed after sleeping on it is that for `StringContains` and `ReplaceInString` enums should be used instead of two bools as they are independent and mutually exclusive paths which an enum communicates well.

WolvenKit.App/Helpers/StringHelper.cs
manavortex: yeah, but then this has to be handled by every consumer. This way, you can directly run it on e.g. `DepotPath.GetResolvedText`, which is exactly the way I want it (if no match, then no need to do anything).

WolvenKit.App/Helpers/StringHelper.cs
manavortex: I wrote this with cnames and depotpaths in mind, so ok, the naming of the parameter is misleading

manavortex:
I have no idea what you mean, can you rephrase or elaborate?

notaspirit:
Take `ReplaceInString` for example, the three exclusive paths are 
whole world replacement
regex replacement
and regular replacement

but there are 4 isWholeWord isRegex combinations 
both true => wholeword path, regex is ignored
isWholeWord true, isRegex false =>  wholeword path
isWholeWord false, isRegex true => regex path
neither true => regular replace

As such an enum with three options like "Regex, WholeWord, Simple" would communicate the relationship between the paths much better.

manavortex:
true, but I don't think I have the brain to get that in right now (fever). Feel free to do it, or maybe open a ticket?

WolvenKit.App/Helpers/StringHelper.cs
notaspirit: Ig it's alright in this case, though imo explicit handling by the consumer would still be better.

notaspirit:
I don't have the time to work on it today, and it should be addressed as part of this pr and not as a future separate issue as while that method structure existed before it was scoped to be private in the chunkviewmodel whereas now it's a public helper method.

WolvenKit.App/Helpers/StringHelper.cs
```diff
@@ -443,4 +443,103 @@ public static string Truncate(string text, int maxLength)
...
+        {
+            return false;
+        }
+
+        switch (searchType)
+        {
+            case SearchType.Contains:
+                return input.Contains(searchOrPattern);
+            case SearchType.Regex:
+                return Regex.IsMatch(input, searchOrPattern);
```
notaspirit, line 502: the ignore case isn't consistent accross all 4 cases

WolvenKit.App/Helpers/ProjectResourceTools.cs
```diff
@@ -1412,4 +1413,71 @@ public List<string> GetAppearanceNames(string relativePath)
     }
 
     #endregion
+
+    public async Task<List<string>> GetFilesContainingString(string searchText, bool isRegex, bool isWholeWord,
+        IProgressService<double> progressService)
```
notaspirit, line 1418: why not add progress service to project resource tools as a private field (like the archive manager already is) instead of passing it to the method?

WolvenKit.App/Helpers/StringHelper.cs
notaspirit: Should be adjusted to be consistent with `StringContains` either way since that uses \b which will match words that aren't the entire string, while both methods are containing the "wholeword" keyword.

notaspirit (changes requested):
To expand on the case sensitivity in the search: it's not just inconsistent within the search method but also between the search method and the replace method. 
The user should be able to choose whether they want it to be case sensitive or not, just like you can in any other dev tool like rider or vsc.

If you arrange the checkboxes like in the screenshot it'll be a lot more compact while being equally usable. 
Ultimately these kinds of things should imo be icon only with a tool tip but that is a style discussion for another time. 

Example of how the grid could look like in comparison to the vertical list: 

Image

notaspirit:
and this is a change big enough to warrant a changelog.

manavortex:
Feel free to implement it yourself. It's good enough for me, I'll keep a personal fork.

PR #2882, opened 2026-03-30
made "find broken references" consider resource files

manavortex:
# made "find broken references" consider resource files

### DIFF
--- a/WolvenKit.App/Models/ProjectManagement/Project/Cp77Project.cs
+++ b/WolvenKit.App/Models/ProjectManagement/Project/Cp77Project.cs
@@ -737,6 +737,7 @@ public sealed partial class Cp77Project : IEquatable<Cp77Project>, ICloneable
         if (filePaths.Count == 0)
         {
             filePaths.AddRange(ModFiles);
+            filePaths.AddRange(ResourceFiles);
         }
 
         SortedDictionary<string, List<string>> references = new();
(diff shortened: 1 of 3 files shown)


WolvenKit.App/Models/ProjectManagement/Project/Cp77Project.cs
notaspirit, line 748: this is missing a check to ensure it's not trying to read files in the resources or matching it against known cr2w extensions (which fail as they're not cr2w files)

notaspirit (changes requested):
doesn't seem to process files in resources (possibly mixing up relative and absolute path in the check before the parallel task?) as completely made up filepaths in `.txt` files don't get flagged as incorrect

WolvenKit.App/ViewModels/Shell/AppViewModel.cs
```diff
@@ -946,8 +946,8 @@ private async Task ScanForBrokenReferencePaths()
     {
         var userInteraction = Interactions.ShowQuestionYesNoCancel((
             $"This will scan {ActiveProject!.ModFiles.Count} files and can take a moment.\n\n"
-            + "Should files in other mods be treated as valid references?\n"
-            + "If you don't know what that means, click 'no'."
+            + "Flag files in other mods?\n"
+            + "If you don't know what that means, click 'yes'."
```
notaspirit: this seems like a regression in communication, the previous sentence was fairly clear while the new one is ambigious. First thing that comes to mind when reading it is "flag how?"

notaspirit (changes requested):
other than that one communication degradation in the dialog it's good

notaspirit (approved):

PR #2883, opened 2026-03-31
fix: copy materials from other mesh

### LINKED ISSUE #2879: materials -> convert material preload state should calculate isDefault of all nodes
After converting preloadMaterials to regular, isDefault property does not refresh:
Image

### LINKED ISSUE #2880: materials -> copy materials from/to other mesh needs to unify preloadLocalMaterial state
**Describe the bug**
Copying materials from/to another mesh will result in cheerfully mixing `localMaterialBuffer.materials` and `preloadLocalMaterials`.

- Fix that
- Improve user feedback
- Optional: Check material name collision and include that in user feedback

### LINKED ISSUE #2881: Make "copy material from other mesh" support multi select
It should be possible to copy materials from multiple meshes - instead of the dropdown menu, there should be a multiselect as in the "copy to" dialogue.

manavortex:
# fix: copy materials from other mesh

### General changes
"Copy materials to/from other mesh(es)" has gotten some love: 
- Added better user feedback via notification service
- Will now consider name collisions (if your mesh already has a material `default`, the material definition will be renamed to `default_1`, and the name will be replaced in all appearances)

#### Copying from: 
- can do multiple meshes now (rather than just one)
- if copying from multiple files, will generate separator appearances 
image

**Additional notes:**
closes #2880
closes #2881 
closes #2879

### DIFF
--- a/WolvenKit.App/Helpers/ArchiveXlHelper.cs
+++ b/WolvenKit.App/Helpers/ArchiveXlHelper.cs
@@ -480,4 +480,6 @@ public static partial class ArchiveXlHelper
         };
         // @formatter:on
     }
+
+
 }
--- a/WolvenKit.App/Helpers/ArchiveXlHelper.cs
+++ b/WolvenKit.App/Helpers/ArchiveXlHelper.cs
@@ -480,6 +480,4 @@ public static partial class ArchiveXlHelper
         };
         // @formatter:on
     }
-
-
 }
--- a/WolvenKit.App/Helpers/DocumentTools.cs
+++ b/WolvenKit.App/Helpers/DocumentTools.cs
@@ -12,6 +12,7 @@ using System.Threading.Tasks;
 using System;
 using WolvenKit.App.Interaction;
 using WolvenKit.App.ViewModels.GraphEditor.Nodes;
+using WolvenKit.Common.Services;
 using WolvenKit.Modkit.RED4.Tools;
 using WolvenKit.Modkit.Scripting;
 using CollectionExtensions = HelixToolkit.SharpDX.Core.CollectionExtensions;
@@ -41,6 +42,7 @@ internal class MultilayerProperties
 public class DocumentTools
 {
     private readonly ILoggerService _loggerService;
+    private readonly INotificationService _notificationService;
     private readonly Cr2WTools _cr2WTools;
     private readonly MeshTools _meshTools;
     private readonly IArchiveManager _archiveManager;
@@ -65,7 +67,7 @@ public class DocumentTools
 
     public DocumentTools(ILoggerService loggerService, Cr2WTools cr2WTools, IArchiveManager archiveManager,
         IProjectManager projectManager, MeshTools meshTools, ProjectResourceTools projectResourceTools,
-        ISettingsManager settingsManager)
+        ISettingsManager settingsManager, INotificationService notificationService)
     {
         _loggerService = loggerService;
         _cr2WTools = cr2WTools;
@@ -74,6 +76,7 @@ public class DocumentTools
         _meshTools = meshTools;
         _projectResourceTools = projectResourceTools;
         _settingsManager = settingsManager;
+        _notificationService = notificationService;
     }
 
 
@@ -1139,10 +1142,10 @@ public class DocumentTools
         mesh.PreloadLocalMaterialInstances.Clear();
     }
 
-    public bool AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string sourcePath = "",
+    public int AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string sourcePath = "",
         string destPath = "")
     {
-        var wasChanged = false;
+        var wasChanged = 0;
 
         if (sourceCr2W is null)
         {
@@ -1167,7 +1170,7 @@ public class DocumentTools
         if (!hasMaterials)
         {
             _loggerService.Error($"source file {sourcePath} does not have materials!");
-            return false;
+            return 0;
         }
 
         // ReSharper disable ForCanBeConvertedToForeach
@@ -1182,7 +1185,7 @@ public class DocumentTools
             appearances.Add(sourceMesh.Appearances[i]);
         }
 
-        wasChanged = wasChanged || destMesh.Appearances.Count != appearances.Count;
+        wasChanged += appearances.Count - destMesh.Appearances.Count;
         destMesh.Appearances = appearances;
 
         CArray<CMeshMaterialEntry> materialEntries = [];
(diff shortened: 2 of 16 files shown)


WolvenKit.App/Helpers/ArchiveXlHelper.cs
```diff
@@ -480,4 +480,6 @@ public static IRedType UnDynamifyResourceReference(IRedType cvpValue, string new
         };
         // @formatter:on
     }
+
+
 }
```
notaspirit: please remove the unnecessary whitespace lines

WolvenKit.App/Helpers/DocumentTools.cs
```diff
@@ -1139,25 +1143,32 @@ public void ClearMeshMaterials(CR2WFile? meshCr2W)
...
-        string destPath = "")
+    private int AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string sourcePath = "",
+        string destPath = "", bool appendSeparator = false)
     {
-        var wasChanged = false;
+        var wasChanged = 0;
+
+        // append separators to appearances, materialNames, materials
+        var separatorString =
+            $"----- merged_from_{Path.GetFileNameWithoutExtension(sourcePath).ToFileName()}";
```
notaspirit: what's the ToFileName for? This is never used as a file name and the source path is already expected to be a valid filename regardless

WolvenKit.App/Helpers/DocumentTools.cs
```diff
@@ -1166,35 +1177,14 @@ public bool AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string
 
         if (!hasMaterials)
         {
+            _notificationService.Error($"source file {sourcePath} does not have materials!");
             _loggerService.Error($"source file {sourcePath} does not have materials!");
-            return false;
+            return 0;
```
notaspirit, line 1188: why break the pattern here by returning 0 explicitly compared to wasChanged like the previous if statements?

WolvenKit.App/Helpers/DocumentTools.cs
```diff
@@ -1139,25 +1143,32 @@ public void ClearMeshMaterials(CR2WFile? meshCr2W)
         mesh.PreloadLocalMaterialInstances.Clear();
     }
 
-    public bool AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string sourcePath = "",
-        string destPath = "")
+    private int AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string sourcePath = "",
+        string destPath = "", bool appendSeparator = false)
     {
-        var wasChanged = false;
+        var wasChanged = 0;
```
notaspirit: variable name should accurately describe it's content and function. wasChanged makes no sense for an int. Consider changedCount or similar.

WolvenKit.App/Helpers/DocumentTools.cs
```diff
@@ -1214,72 +1204,241 @@ public bool AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string
...
+        wasChanged += materialEntries.Count - destMesh.MaterialEntries.Count;
 
-        CArray<IMaterial> localMaterials = [];
-        for (var i = 0; i < destMesh.LocalMaterialBuffer.Materials.Count; i++)
+        destMesh.MaterialEntries.Clear();
+        materialEntries.ForEach(e => destMesh.MaterialEntries.Add(e));
+
+        var appearances = CopyMeshAppearances();
+
+        wasChanged += appearances.Count - destMesh.Appearances.Count;
```
notaspirit: seems overloaded, the same variable is used for both materials and appearances, what is it intended to signal to the caller?

WolvenKit.App/Helpers/DocumentTools.cs
```diff
@@ -1139,25 +1143,32 @@ public void ClearMeshMaterials(CR2WFile? meshCr2W)
         mesh.PreloadLocalMaterialInstances.Clear();
     }
 
-    public bool AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string sourcePath = "",
-        string destPath = "")
+    private int AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string sourcePath = "",
```
notaspirit, line 1155: even for a private method if the return type / value meaning is not immediately clear it should have annotations

WolvenKit.App/Helpers/DocumentTools.cs
```diff
@@ -1214,72 +1204,241 @@ public bool AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string
...
+                    IsLocalInstance = false,
+                    Index =
+                        (CUInt16)(CvmMaterialTools.FindHighestMaterialIndex(destMesh.MaterialEntries, false) + 1),
+                });
+            }
+
+            for (var i = 0; i < sourceMesh.MaterialEntries.Count; i++)
+            {
+                var materialEntry = sourceMesh.MaterialEntries[i];
+                var materialName = materialEntry.Name.GetResolvedText() ?? "destMaterial" + i;
```
notaspirit: $"destMaterial{i}" has better readability

WolvenKit.App/Helpers/DocumentTools.cs
```diff
@@ -1139,25 +1143,32 @@ public void ClearMeshMaterials(CR2WFile? meshCr2W)
...
+        var wasChanged = 0;
+
+        // append separators to appearances, materialNames, materials
+        var separatorString =
+            $"----- merged_from_{Path.GetFileNameWithoutExtension(sourcePath).ToFileName()}";
 
         if (sourceCr2W is null)
         {
+            _notificationService.Error($"source file {sourcePath} not found. Can't copy...");
             _loggerService.Error($"source file {sourcePath} not found. Can't copy...");
```
notaspirit: those are new sentences and should start with a capital letter, also why the triple dots?

WolvenKit.App/Helpers/DocumentTools.cs
```diff
@@ -1214,72 +1204,241 @@ public bool AppendMeshMaterials(CR2WFile? sourceCr2W, CR2WFile? destCr2W, string
...
     }
 
-    public bool CopyMeshMaterials(string? sourcePath, string destPath, bool? append)
+    /// <summary>
+    /// Copies materials from a source mesh to a dest mesh. Can replace or append.
+    /// </summary>
+    /// <param name="sourcePath">source path (file to copy FROM)</param>
+    /// <param name="destPath">dest path (file to copy TO)</param>
+    /// <param name="append">append materials from source mesh instead of replacing them?</param>
+    /// <param name="addSeparator">Add separators? (Will only be set when copying _from_, not when copying _to_)</param>
```
notaspirit: addSeparator isn't in the method signature

notaspirit (changes requested):
Image

- The mesh list should be expanding to fill the dialog when it's resized so you're not stuck scrolling such a small window.
- Writing a mesh path manually clears all previously selected checkboxes without updating the ui.
  - The expectation from the user perspective is that they are not mutually exclusive. Also since the checkboxes have multimesh support the manually typed paths should as well. 
- The dialog remembers the last manually provided mesh path, but despite it being valid the finish button is disabled until the text box content is edited
- As well as the missing changelog entry

WolvenKit.App/Helpers/DocumentTools.cs
manavortex: Since this uses any project file, users can end up with text that triggers file validation because it has uppercase characters or whitespaces. I know at least three different people who absolutely INSIST on copying mesh files around by hand so they can have a nice `- Copy` suffix, and at least two of them panic at the sight of file validation.

WolvenKit.App/Helpers/DocumentTools.cs
manavortex: how many changes were made in the file

WolvenKit.App/Helpers/DocumentTools.cs
notaspirit: Why would file validation complain about a field that is not used as a filepath?

manavortex: because CNames are not supposed to have non-alphanumeric chars either.

WolvenKit.App/Helpers/DocumentTools.cs
```diff
@@ -1139,62 +1143,55 @@ public void ClearMeshMaterials(CR2WFile? meshCr2W)
...
-            _loggerService.Error($"source file {sourcePath} not found. Can't copy...");
-            return wasChanged;
+            _notificationService.Error($"Source file {sourcePath} not found. Can't copy.");
+            _loggerService.Error($"Source file {sourcePath} not found. Can't copy.");
+            return 0;
         }
 
         if (destCr2W is null)
         {
+            _notificationService.Error($"target file {destCr2W} not found. Can't copy...");
```
notaspirit: All these messages start with a lower case, should be upper like the first one

notaspirit (changes requested):
Image

1: It's very unclear that that's a button, especially not one that completes the dialog. 
You could have it be e.g. a checkbox instead that then disables all other inputs
2: The main list still doesn't expand to fill the space when the dialog is resized

The changelog entry got lost in the main merge, the one that was there before was very vague though. If I was a user I wouldn't really know what changed as "logic" means anything. 

And the whitespace in ArchiveXlHelper is back / still there.

manavortex:
No clue how to make it resize when the dialogue does, and tbh, I'm not motivated to find out.
The button has been like that for as long as the feature exists, and it uses the default WKit button style, which I was told that I'm not supposed to deviate from.
As for the capitalization and the changelog entry, fixed that.
The whitespace/ extra line is not there in my version of the code, presumably you're using an old diff?

notaspirit:
okay, then that button is a separate issue I'll open a ticket for it. 
For the selection resizing I'm pretty sure you just need to set that row height to `*` so it takes up all available space. 

Odd, the whitespace is gone now from just reloading the page, must be github having issues again or something. 

Then that's just the selection element resizing and it can be merged.

notaspirit (approved):

Issue #2432, opened 2025-06-26
Issue with mlmasks export/import


Ametis81:
After mlmasks extraction we are getting mask images with an offset (its noticeable with scrolling layer mask preview in WK). So packing them back to game will cause an issue with material layering

1. Create a project
2. Find _notell_room_wall_h400_l200_vending_a.mesh_ in WK
3. Find used files
4. Add _ml_notell_room_set_b_masksset.mlmask_ to project
5. Export mlmask file
6. Import mlmask back
7. Install project to mod folder
8. Run the game
9. Spawn mesh in game via World Builder (or AMM)

You can export mesh to Blender and see problem for the default appearance there. If you select for instance _ml_notell_room_set_b_pink_zebra_masksset_ you may see correct layer representation

**Expected behavior**
Same material layering after importing masks back

WK Version:

![Image](https://github.com/user-attachments/assets/fb5025ee-e76b-48bc-896a-0bc8c88b64e2)

**Screenshots**

Default materal:

![Image](https://github.com/user-attachments/assets/1ffbffb5-d88f-4e09-a150-90a889d89261)

Same material with same masks, but after importing them back:

![Image](https://github.com/user-attachments/assets/099dbd0e-329e-47fa-a95a-5e48c1d2b70f)

We are getting same wrong result in Blender:

![Image](https://github.com/user-attachments/assets/21111d62-2153-421c-ba17-a0e50d882783)

Wrong material layering in game:

![Image](https://github.com/user-attachments/assets/586792f0-f7b9-4249-83e4-39f8e845e19f)

poirierlouis:
So based on the screenshot and your feedback, it seems UV values lost precision, hence textures bleeding on edges.

@seberoth I'm not familiar yet with mlmask and multilayer implementation, but is my guess correct?

poirierlouis:
I'll be able to take a look this week. Thanks for all the details. Regarding the technique, the best would be to use the one implemented by the engine (if not already the case) to avoid any surprise. I actually wonder why the resolution is not preserved when going from WKit to Blender's plugin, I'll ask around.

Ametis81:
> I'll be able to take a look this week. Thanks for all the details. Regarding the technique, the best would be to use the one implemented by the engine (if not already the case) to avoid any surprise. I actually wonder why the resolution is not preserved when going from WKit to Blender's plugin, I'll ask around.

[...]

PR #2892, opened 2026-04-13
MlmaskTools.cs / MlmaskImportTools.cs update - ML masks export / import

Ametis81:
Fixed:

Fixed multilayered masks after export: now they have the required resolution from specified parameters in the .mlmask file; there is no pixels shift due to bilinear interpolation (it is replaced by Nearest Neighbor method)

Result:  We have masks with various resolution, not only 256x256, no more pixel shift. Perhaps there is a better solution to this, since in Blender we can still notice some artifacts in some cases of layer blending (maybe this is a feature of the application itself), but they are not as noticeable as when using the current export method.

(fixes by MrChedda)

### DIFF
--- a/WolvenKit.App/WolvenKit.App.csproj
+++ b/WolvenKit.App/WolvenKit.App.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net8.0-windows</TargetFramework>
-    <Version>8.18.0</Version>
+    <Version>8.18.1</Version>
     <Platforms>x64</Platforms>
     <SatelliteResourceLanguages>en</SatelliteResourceLanguages>
     <Nullable>enable</Nullable>
--- a/WolvenKit.CLI/WolvenKit.CLI.csproj
+++ b/WolvenKit.CLI/WolvenKit.CLI.csproj
@@ -4,7 +4,7 @@
     <Platforms>x64</Platforms>
     <OutputType>Exe</OutputType>
     <TargetFramework>net8.0</TargetFramework>
-    <VersionPrefix>8.18.0</VersionPrefix>
+    <VersionPrefix>8.18.1</VersionPrefix>
     <VersionSuffix>$(VersionSuffix)</VersionSuffix>
     <IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
 
--- a/WolvenKit.Common/WolvenKit.Common.csproj
+++ b/WolvenKit.Common/WolvenKit.Common.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
-    <Version>8.18.0</Version>
+    <Version>8.18.1</Version>
     <Platforms>x64</Platforms>
     <SatelliteResourceLanguages>en</SatelliteResourceLanguages>
     <Nullable>enable</Nullable>
--- a/WolvenKit.Core/WolvenKit.Core.csproj
+++ b/WolvenKit.Core/WolvenKit.Core.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
-    <Version>8.18.0</Version>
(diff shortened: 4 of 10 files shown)


Ametis81:
https://github.com/WolvenKit/WolvenKit/issues/2432

notaspirit:
I'll mark it as a draft since the discussion about the import side is ongoing (as mentioned in #2432). If that is not the case and the matter is already resolved feel free to mark it as ready for review at any time.

Ametis81:
updated both files, dunno if it gets new versions of them here, i hope

Ametis81:
## Summary
- Export now always outputs masks in their **native per-layer resolutions**
- Added rich `.masklist` metadata export with full header information (`# LayerResolutions=...`, original atlas/mask sizes, etc.)
- Import now correctly reads `LayerResolutions` from `.masklist` and restores proper per-layer sizes during round-trip

## Changes

### MlmaskTools.cs (Export)
- `GetRedImages()` now always returns layers in their original native resolution
- `UncookMlmask()` now writes a detailed `.masklist` header containing:
  - `# LayerResolutions=...` (comma-separated list of all layer sizes)
  - Original header values (`AtlasWidth`, `MaskWidth`, `MaskWidthLow`, etc.)
- Added/updated `ConvertMultilayerMaskToDdsStreams()` for internal preview compatibility
- Improved `DownscaleNearest()` with boundary safety

### MlmaskImportTools.cs (Import)
- Parses `# LayerResolutions=` from `.masklist` header
- Uses per-layer resolutions when importing (auto-upscales content to correct target size)
- Preserves original `MaskWidthLow` / `MaskHeightLow` from header for correct round-trip

## Benefits
- True lossless round-trip workflow (export → edit → import → re-export)
- Correct handling of all layer sizes (4×4, 128×128, 256×256, 1024×1024, etc.)
- Much cleaner and more informative `.masklist` files for modders
- No more pixel shifting issues

## Compatibility / Notes
- Old `.masklist` files without `LayerResolutions` header will still work (fallback to uniform high-res)

Ametis81:
update for masks import / export so we use image metadata about resolution instead of reading resolutions from .masklist file (i accidently edited MeshImportTools, so it shouldn't be here)

notaspirit:
If it is possible to derive all the new values written to the exported file I'd be faster to merge and release if the file contents didn't change as that is a breaking change given that old versions will fail to read the new format. 

If some data can't be derived and has to be written to the file at least use json (with a version number field), but the merge / release time frame will be months as the breaking change would need to be bundled with other breaking changes and deprecation time frames for a single major release.

Ametis81:
Basically old masklist file just getting new commented lines in header, then we use only these commented lines in import:

```
# AtlasWidth=...
# AtlasHeight=...
# MaskTileSize=...
```

Currently after we get new low res and high res masks, if we try to pack them back atlas size can change (mostly noticable for masks 1024 and more). The whole point of some header values is to try and pack mlmask file to its default values, so it will be the same size as before we exported masks.  If not, there is a fallback and we will get new atlas size.

I'm not sure if there is a need to edit old masks and pack them back (often we do changes in layers with color, microblend patterns and so on), so the initial use of them is mostly for export to edit materials in Blender.  The import stuff is mostly for adding new masks for your material of your mesh you created from scratch, so import to mlmask file will do it without any problems 

We still use only paths to masks same as in old masklist, so if you have old masklist file without new header you also be able to import stuff, and you'll get new atlas size. I tried to keep compatibility with old masklist file import

BTW, I can also remove that header creation after exporting, so file will be the same old masklist file, just the atlas size will be different after we try import back our extracted masks and that's all

Ametis81:
Updated: Default masklist file without any additions, atlas size could be different after importing masks compared to original mlmask file data, but working in game pretty fine (we got low res / high res masks after export and able to import them back, no more issues after testing described in https://github.com/WolvenKit/WolvenKit/issues/2432

Ametis81:
Code cleaning / formatting, better logging MlmaskTools.cs / MlmaskImportTools.cs

notaspirit:
when this gets merged it should be squashed and merged given the commit messages

PR #2895, opened 2026-04-15
context menu import export

### LINKED ISSUE #2629: feat: right click export import
the import / export tools take up a lot of space and are quite disconnected from the other actions in the explorer.
If you hide them you constantly need to show them again and if you don't hide them they take up a lot of space for no good reason.

Add import / export to the context menu of files in the project explorer, for export you can have an option for import with defaults and one where you can change the config now, potentially differentiated via shift (=> without shift asks for config, with uses default).
There should also be an option to change the default values.

notaspirit:
# context menu import export

**Implemented:**
- added import export to context menu in file browser for both singular files and bulk 
   - only shows up when relevant files are selected
   - sub context menu for action with default config and dialog for modifying the config 

- moved some checks from Import / Export view models to import export helper for global access
- added button styles to styles xml (matching the update dialog buttons) 

**Additional notes:**
closes #2629

### DIFF
--- a/WolvenKit.App/Factories/PaneViewModelFactory.cs
+++ b/WolvenKit.App/Factories/PaneViewModelFactory.cs
@@ -79,8 +79,8 @@ public class PaneViewModelFactory : IPaneViewModelFactory
 
     public LogViewModel LogViewModel() => new(_loggerService, _appScriptService, _settingsManager);
     public ProjectExplorerViewModel ProjectExplorerViewModel(AppViewModel appViewModel)
-        => new(appViewModel, _projectManager, _loggerService, _progressService, _modTools,
-            _gameController, _pluginService, _settingsManager, _modifierSvc, _archiveManager, _projectResourceTools);
+        => new(appViewModel, _projectManager, _loggerService, _notificationService, _progressService, _modTools,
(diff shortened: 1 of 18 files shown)


manavortex:
image
Should "to JSON" also be in the export menu?

notaspirit:
no, since the import and export options introduced in this pr are analogous to the import and export tools and json conversion is separate from that. Not to mention it would change the placement that users are used to with no clear gain.

manavortex:
It does export the file, though ¯\_(ツ)_/¯ this will lead to confusion either way

notaspirit:
I don't see how that is confusing, can you elaborate?

manavortex:
The context menu now has two ways of exporting, and they are not in the same place. People will absolutely mix them up no matter what because reading is hard

notaspirit:
It's not any different from the export and import tools not handling conversion to and from json files, that has historically been a separate action. 
If anything adding it under the import and export submenu would only lead to more confusion as it would imply that the import and export tools have that functionality as well when they don't, likely leading someone to look for that functionality where it is not. 
The wording clearly separates them as well, "Convert to JSON" / "Convert from JSON" vs "Export" / "Import".

manavortex (approved):

PR #2897, opened 2026-04-16
Expand mesh appearances

manavortex:
# Expand mesh appearances

**Implemented:**
- Mesh: added "expand mesh appearances" to ArchiveXL menu


WolvenKit.App/Helpers/ChunkViewModel/CvmMaterialTools.cs
```diff
@@ -565,34 +565,54 @@ int GetIndexOfFirstNonRepeatingMaterial(List<string> materials)
         }
     }
 
-    public void UnDynamifyMaterials(ChunkViewModel? cvm)
+    public Dictionary<string, List<string>> ExpandMeshAppearances(ChunkViewModel? cvm,
+        out Dictionary<string, List<string>> templatesAndValues, bool preserveDynamicChunks = false)
     {
+        templatesAndValues = [];
```
notaspirit: what's the reasoning for having the same dictionary as both the return type and an out param?

notaspirit:
Do you have a mesh to test this with?

manavortex:
Anything from https://www.nexusmods.com/cyberpunk2077/mods/10516

notaspirit (changes requested):
sorry for the review taking so long, other than the return value it looks fine to me, also please add a changelog since that's become a requirement after this pr has been made

notaspirit (approved):

PR #2907, opened 2026-05-03
file validation yaml parsing: better error message

manavortex:
# file validation yaml parsing: better error message

### DIFF
--- a/WolvenKit.Modkit/Scripting/ScriptFunctions.cs
+++ b/WolvenKit.Modkit/Scripting/ScriptFunctions.cs
@@ -197,7 +197,9 @@ public partial class ScriptFunctions
         catch (Exception e)
         {
             _loggerService.Error("Failed to read YAML: " + e.Message);
-            _loggerService.Error(e);
+            _loggerService.Error("");
+            _loggerService.Error("Please run the file through https://yamllint.com and fix all errors!");
+            _loggerService.Error("");
         }
 
         return "";


notaspirit (changes requested):
I can see what this pr attempts to address, the full stacktrace is definitely not useful here. However the approach to solving it doesn't address the real issue which is the odd error handling. 
As this is a method called from the wscripts it doesn't make sense to instruct the user to run the file through a yaml linter as there is no expectation that the yaml content run through this method is even user generated content that they can affect.
Instead this method should log nothing and pass on the error to the scripting environment so it can handle it however it sees fit. 

The proper solution would be to adjust the error handling to propagate to the scripting layer so that the script can handle it. This is of course outside of the scope of this pr so I will open a separate issue for it. 

If you want to just remove the logging of the entire exception including stack trace, only keeping the logging of the exception message that can be merged as part of this pr.

notaspirit:
The issue mentioned above is #2913

manavortex:
But this method is literally used to verify yaml files inside the project?

notaspirit:
What matters is what this method is designed to do, which is convert a yaml string to a json string. Like mentioned previously error handling should be done on the scripting side so it can handle the context correctly.
This method may be used in some instances for validating yaml files, but adding the additional log message about linting the file would create instructions which do not make sense in a lot of scenarios where the method is used as designed.

manavortex:
What other use case could there be other than "parsing the contents of a yaml file from script", though?

notaspirit:
I don't have a different common usecase from the top of my mind, however whether I can come up with one is not the point. 

My previous points still stand.

manavortex:
The current implementation dumps the error - if you want the method to throw it instead, that might break existing user scripts. I can change "file" to "yaml" if you prefer that, but I'd rather not introduce new behaviour. If you want, we can add a `YamlToJsonThrowsException` or so.

notaspirit:
Yeah, like I initially said changing the error handling will be done in a separate pr, hence the separate issue. 

Since this pr is just about adjusting the error messaging the suggestion was to just remove the stack trace logging and keep everything else the same, which would already be an immediate improvement as the stack trace isn't useful and is just polluting the log.

manavortex:
I initially touched this because I was given a full error message by a confused user. If I implement your change, I will have to add error checking in every place where I use this across file validation or implement the extra method. So instead of improving UX, this will make it worse and add significant implementation overhead for me to make it work again.
Not worth the spoons, sorry.

PR #2917, opened 2026-05-12
fix "create item codes from yaml"

manavortex:
# fix "create item codes from yaml"

Yaml paths will now point at the correct folder

### DIFF
--- a/WolvenKit.App/ViewModels/Shell/MenuBarViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/MenuBarViewModel.cs
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.IO;
 using CommunityToolkit.Mvvm.ComponentModel;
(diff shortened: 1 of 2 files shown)


notaspirit (changes requested):
code wise looks good, but it's missing the changelog entry

manavortex:
I thought the changelog was being generated?

notaspirit:
The user facing changelog is getting generated from the yaml source file (`changelog\!unreleased.yaml`). Each pr with relevant changes needs to add an entry otherwise there is nothing to generate the user facing changelog from. 
For the last release I had written most of them myself as the changelog requirement was introduced mid release cycle. 
The relevant info for contributing is on this [dev wiki page](<https://wolvenkit.github.io/WolvenKit/contributing/keep-changelog.html>)

manavortex:
I thought that file was populated on pull request 😭 guess I'll do it by hand later

notaspirit (approved):

PR #2919, opened 2026-05-13
FIx for MlmaskTools.cs / RDTMeshViewModel.cs

Ametis81:
MlmaskTools.cs - R8G8B8A8 to R8 (Greyscale with one channel for masks export)
RDTMeshViewModel.cs - Fixed material loading with Generate Materials button


Ametis81:
added fix for material preview

PR #2921, opened 2026-05-14
Adjust release actions to no longer push to main 

notaspirit:
# Adjust release actions to no longer push to main 

**Implemented:**
- the stage changelog action now generates a pull request targeting the main branch which contains all the changes needed as part of the release (appended main changelog, separate changelog files for this specific release in markdown and bbcode as well as bumping the yaml changelog version) 
- the release action now simply reads those files and does not make any changes to the upstream
- Adjusted the docs to reflect new workflow

This avoids the issue of providing the action with permissions to push to a protected branch as well as providing a more transparent and boundary setting process.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,28 +52,20 @@ jobs:
         id: get_version
         with:
           proj-path: ${{env.PROJ}}
+
       - name: change version
         shell: bash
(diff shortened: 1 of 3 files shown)


notaspirit:
@manavortex can you review this? The adjusting of the way the changelog files are handled to avoid constant conflicts would be build on top of these changes.

manavortex (approved):

PR #2922, opened 2026-05-15
Masks export / import overhaul (low res and high res masks / no pixel shift/offset)

Ametis81:
Update for exporting / importing masks at their native resolution (low and high res masks with no pixel shift / offset due to scaling) to solve the problem in https://github.com/WolvenKit/WolvenKit/issues/2432

Fixed export  - greyscale with one channel instead of RGB

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
@@ -52,7 +52,6 @@ public class MLMASK
 
         if (files.Count == 0)
         {
-            // 0x2006 = MLMask: No layer files specified in masklist
             throw new ArgumentException("No layer files specified in masklist.", nameof(txtImageList));
         }
 
--- a/WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
@@ -14,12 +14,10 @@ using WolvenKit.RED4.CR2W;
 using WolvenKit.RED4.Types;
 
 /// <summary>
-/// Error code ranges used in multilayer mask import:
-/// 0x2000–0x20FF : Import / masklist parsing and image loading errors
+/// MLMask import error codes (registered in LogCodeHelper):
 ///   0x2001 = Invalid color space (must be R8/grayscale)
-///   0x2002 = Reserved (power-of-2 requirement)
-///   0x2004 = No layers found in masklist
-///   0x2005 = General image loading failure
+///   0x2006 = No layer files specified in .masklist
+///   0x2007 = Failed to load layer image
 /// </summary>
 
 namespace WolvenKit.Modkit.RED4.MLMask
@@ -62,8 +60,8 @@ namespace WolvenKit.Modkit.RED4.MLMask
 
             if (files.Count == 0)
             {
-                // 0x2004 = No layers found in masklist (0x2002 is reserved for power-of-2 requirement errors)
-                throw new WolvenKitException(0x2004, "No layer files specified in masklist.");
+                // 0x2006 = MLMask: No layer files specified in masklist
+                throw new WolvenKitException(0x2006, "No layer files specified in masklist.");
             }
 
             _mlmask = new MlMaskContainer();
@@ -80,7 +78,7 @@ namespace WolvenKit.Modkit.RED4.MLMask
 
                 try
                 {
-                    using var image = RedImage.LoadFromFile(f) ?? throw new WolvenKitException(0x2005, $"Could not load image: {f}"); // 0x2005 = General image loading failure
+                    using var image = RedImage.LoadFromFile(f) ?? throw new WolvenKitException(0x2007, $"Could not load image: {f}"); // 0x2007 = MLMask: General image loading failure
 
                     if (image.Metadata.Format != DXGI_FORMAT.DXGI_FORMAT_R8_UNORM)
                     {
--- a/WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
@@ -5,7 +5,6 @@ using System.Linq;
 using System.Runtime.InteropServices;
 using WolvenKit.Common;
 using WolvenKit.Common.DDS;
-using WolvenKit.Common.Exceptions;
 using WolvenKit.Core.Exceptions;
 using WolvenKit.Core.Extensions;
 using WolvenKit.Core.Interfaces;
@@ -16,14 +15,9 @@ using WolvenKit.RED4.Types;
 
 namespace WolvenKit.Modkit.RED4.MLMask
 {
-    // TODO refactor all this, it's completely unsafe
-
     public class MLMASK
     {
         private const uint s_headerLength = 148;
-        //assuming DDSUtils.ConvertToDdsMemory always creates a dx10 dds,
-        //if it creates a dx9 or both we will have to check for it, headerlength = 128, if(dx10) headerlength += 20
-
         private MlMaskContainer _mlmask;
         private readonly ILoggerService _logger;
 
@@ -33,107 +27,117 @@ namespace WolvenKit.Modkit.RED4.MLMask
             _logger = logger;
         }
 
+        /// <summary>
+        /// Imports a .masklist file into a .mlmask file.
+        /// Supports both single-resolution and multi-resolution layer sets.
+        /// The .masklist now contains only layer file paths (no header).
+        /// </summary>
         public void Import(FileInfo txtImageList, FileInfo outFile)
         {
-            // relative and absolute paths
-            var paths = File.ReadAllLines(txtImageList.FullName);
+            var lines = File.ReadAllLines(txtImageList.FullName);
             var baseDir = txtImageList.Directory.NotNull();
-            var files = paths.Select(x => Path.Combine(baseDir.FullName, x)).ToList();
 
-            #region InitandVerify
+            var filePaths = new List<string>();
 
-            _mlmask = new MlMaskContainer();
-            var textures = new List<RawTexContainer>();
-
-            var firstLayerName = Path.GetFileNameWithoutExtension(files[0]);
-            if (!firstLayerName.EndsWith("_0"))
+            foreach (var line in lines)
             {
-                var white = new RawTexContainer
+                var trimmed = line.Trim();
+                if (string.IsNullOrWhiteSpace(trimmed) || trimmed.StartsWith("#"))
                 {
-                    Pixels = new byte[16],
-                    Width = 4,
-                    Height = 4
-                };
-                Array.Fill<byte>(white.Pixels, 255);
-                textures.Add(white);
-            }
-
-            (uint imageDimensionX, uint imageDimensionY) = (0, 0);
-            foreach (var f in files)
-            {
-                if (!File.Exists(f))
-                {
-                    throw new FileNotFoundException($"Line{{lineIdx}}: \"{f}\" Make sure the file path is valid and exists (paths are specified line by line in ascending layer order in masklist)");
-                }
-
-                RedImage? image = null;
-                try
-                {
-                    image = RedImage.LoadFromFile(f);
-                }
-                catch (WolvenKitException e)
-                {
-                    throw new WolvenKitException(0x2001, $"{e.Message} (.mlmask images need to be in color space black and white)");
-                }
-
-                if (image == null)
-                {
-                    _logger.Error($"\"{f}\" could not be loaded!");
                     continue;
                 }
 
-                if (image.Metadata.Format != DXGI_FORMAT.DXGI_FORMAT_R8_UNORM)
-                {
-                    image.Convert(DXGI_FORMAT.DXGI_FORMAT_R8_UNORM);
-                }
-
-                if (image.Metadata.Width % 2 != 0 || image.Metadata.Height % 2 != 0)
-                {
-                    throw new WolvenKitException(0x2002,
-                        $"Texture {f}: width={image.Metadata.Width},height={image.Metadata.Height} must have dimensions divisible by 2");
-                }
-
-                using var ms = new MemoryStream(image.SaveToDDSMemory());
-                using var br = new BinaryReader(ms);
-                ms.Seek(s_headerLength, SeekOrigin.Begin);
-                var bytes = br.ReadBytes(image.Metadata.Width * image.Metadata.Height);
-
-                // #1865: check the the image dimensions are the same for all pngs
-                var imageWidth = (uint)image.Metadata.Width;
-                var imageHeight = (uint)image.Metadata.Height;
-
-                if (imageDimensionX == 0)
-                {
-                    imageDimensionX = imageWidth;
-                    imageDimensionY = imageHeight;
-                }
-                else if (imageDimensionX != imageWidth || imageDimensionY != imageHeight)
-                {
-                    throw new WolvenKitException(0x2003, $"Texture {f} should be of size {imageDimensionX}x{imageDimensionY}");
-                }
-
-                var tex = new RawTexContainer { Width = imageWidth, Height = imageHeight, Pixels = bytes };
-                textures.Add(tex);
+                filePaths.Add(trimmed);
             }
+
+            if (filePaths.Count == 0)
+                throw new WolvenKitException(0x2002, "No layer files specified in masklist.");
+
+            var files = filePaths.Select(x => Path.Combine(baseDir.FullName, x)).ToList();
+
+            _mlmask = new MlMaskContainer();
+
+            var textures = new List<RawTexContainer>();
+
+            // Add white base layer 0 (if the first layer does not end with _0)
+            var firstLayerName = Path.GetFileNameWithoutExtension(files[0]);
+            var needsLayer0 = !firstLayerName.EndsWith("_0") && !firstLayerName.Equals("0");
+
+            if (needsLayer0)
+            {
+                uint whiteW = 1024u;
+                uint whiteH = 1024u;
+
+                if (files.Count > 0 && File.Exists(files[0]))
+                {
+                    try
+                    {
+                        using var probe = RedImage.LoadFromFile(files[0]);
+                        if (probe != null)
+                        {
+                            whiteW = (uint)probe.Metadata.Width;
+                            whiteH = (uint)probe.Metadata.Height;
+                        }
+                    }
+                    catch { }
+                }
+
+                var whitePixels = new byte[whiteW * whiteH];
+                Array.Fill(whitePixels, (byte)255);
+                textures.Add(new RawTexContainer { Width = whiteW, Height = whiteH, Pixels = whitePixels });
+            }
+
+            // Load all layers from files
+            foreach (var f in files)
+            {
+                if (!File.Exists(f))
+                    throw new FileNotFoundException($"File not found: {f}");
+
+                try
+                {
+                    using var image = RedImage.LoadFromFile(f) ?? throw new WolvenKitException(0x2001, $"Could not load image: {f}");
+
+                    if (image.Metadata.Format != DXGI_FORMAT.DXGI_FORMAT_R8_UNORM)
+                        image.Convert(DXGI_FORMAT.DXGI_FORMAT_R8_UNORM);
+
+                    uint imgW = (uint)image.Metadata.Width;
+                    uint imgH = (uint)image.Metadata.Height;
+
+                    using var ms = new MemoryStream(image.SaveToDDSMemory());
+                    using var br = new BinaryReader(ms);
+                    ms.Seek(s_headerLength, SeekOrigin.Begin);
+                    var pixels = br.ReadBytes((int)(imgW * imgH));
+
+                    textures.Add(new RawTexContainer { Width = imgW, Height = imgH, Pixels = pixels });
+                }
+                catch (WolvenKitException e)
+                {
+                    throw new WolvenKitException(0x2001, $"{e.Message} (must be grayscale R8)");
+                }
+            }
+
             _mlmask.Layers = textures.ToArray();
-            #endregion
 
             Create();
             Write(outFile);
         }
 
-        #region Methods
+        internal static bool IsBlankLayer(byte[] pixels)
+        {
+            if (pixels == null || pixels.Length == 0) return true;
+            var first = pixels[0];
+            for (int i = 1; i < pixels.Length; i++)
+                if (pixels[i] != first) return false;
+            return true;
+        }
+
+        #region Core Methods
 
         private void Write(FileInfo f)
         {
             var cr2w = new CR2WFile();
-            var mask = new Multilayer_Mask
-            {
-                CookingPlatform = Enums.ECookingPlatform.PLATFORM_PC
-            };
+            var mask = new Multilayer_Mask { CookingPlatform = Enums.ECookingPlatform.PLATFORM_PC };
             cr2w.RootChunk = mask;
-
-            ArgumentNullException.ThrowIfNull(_mlmask.Layers, nameof(_mlmask.Layers));
 
             var blob = new rendRenderMultilayerMaskBlobPC
             {
@@ -142,7 +146,7 @@ namespace WolvenKit.Modkit.RED4.MLMask
                     Version = 3,
                     AtlasWidth = _mlmask.AtlasWidth,
                     AtlasHeight = _mlmask.AtlasHeight,
-                    NumLayers = (uint)_mlmask.Layers.Length,
+                    NumLayers = (uint)_mlmask.Layers!.Length,
                     MaskWidth = _mlmask.WidthHigh,
                     MaskHeight = _mlmask.HeightHigh,
                     MaskWidthLow = _mlmask.WidthLow,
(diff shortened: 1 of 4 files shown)


WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+    // Decode each grayscale layer to RedImage (always outputs linear R8G8B8A8_UNORM)
+    private static IEnumerable<RedImage> GetRedImages(rendRenderMultilayerMaskBlobPC blob)
+    {
+        uint atlasWidth = blob.Header.AtlasWidth;
+        uint atlasHeight = blob.Header.AtlasHeight;
+        uint maskWidth = blob.Header.MaskWidth;
+        uint maskHeight = blob.Header.MaskHeight;
+        uint maskWidthLow = blob.Header.MaskWidthLow;
+        uint maskHeightLow = blob.Header.MaskHeightLow;
+        uint maskTileSize = blob.Header.MaskTileSize;
```
notaspirit: All of these are unused.

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+        using (var ms = new MemoryStream(tileBuffer.GetBytes()))
+        using (var br = new BinaryReader(ms))
+        {
+            ms.Seek(0, SeekOrigin.Begin);
+            for (var i = 0; i < tiles.Length; i++)
             {
-                throw new Exception();
+                tiles[i] = br.ReadUInt32();
             }
+        }
```
notaspirit: this writes to tiles, even though tiles is never read.

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+        uint maskHeight = blob.Header.MaskHeight;
+        uint maskWidthLow = blob.Header.MaskWidthLow;
+        uint maskHeightLow = blob.Header.MaskHeightLow;
+        uint maskTileSize = blob.Header.MaskTileSize;
+
+        var atlasRaw = new byte[atlasWidth * atlasHeight];
+        if (!BlockCompression.DecodeBC(blob.AtlasData.Buffer.GetBytes(), ref atlasRaw, atlasWidth, atlasHeight, BlockCompression.BlockCompressionType.BC4))
+        {
+            throw new WolvenKitException(0x3001, "BC4 decode failed for multilayer mask atlas.");
+        }
```
notaspirit: it's decoding into atlasRaw but then is never using it

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+        uint maskHeight = blob.Header.MaskHeight;
+        uint maskWidthLow = blob.Header.MaskWidthLow;
+        uint maskHeightLow = blob.Header.MaskHeightLow;
+        uint maskTileSize = blob.Header.MaskTileSize;
+        uint maskCount = blob.Header.NumLayers;
+
+        var atlasRaw = new byte[atlasWidth * atlasHeight];
+        if (!BlockCompression.DecodeBC(blob.AtlasData.Buffer.GetBytes(), ref atlasRaw, atlasWidth, atlasHeight, BlockCompression.BlockCompressionType.BC4))
+        {
+            throw new WolvenKitException(0x3001, "BC4 decode failed for multilayer mask atlas.");
```
notaspirit: how is this error related to archive xl patching? (which 0x3001 links to)

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
-                };
-
-                yield return RedImage.Create(info, maskData);
-            }
+    /// <summary>
+    /// Decodes each layer from the multilayer mask.
+    /// Layers are exported either at full high resolution or at the reduced low resolution,
+    /// depending on whether the layer has dedicated high-resolution tile data.
+    /// </summary>
+    private static IEnumerable<(byte[] buffer, uint width, uint height)> DecodeLayerBuffers(rendRenderMultilayerMaskBlobPC blob)
```
notaspirit: use a record for this instead of returning a tuple

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
         }
 
-        private static byte BilinearInterpolation(byte q00, byte q10, byte q01, byte q11, int x, int x1, int y, int y1)
-        {
-            const int sc = 256;
+        return false;
+    }
 
-            if (x1 == 0 || y1 == 0)
+    private static void Decode(ref byte[] maskData, uint maskWidth, uint maskHeight, uint maskWidthLow, uint maskHeightLow, byte[] atlasData, uint atlasWidth, uint atlasHeight, uint[] tileData, uint maskTileSize, int maskIndex)
```
notaspirit: if you need that many parameters just make it a nested method within the parent method then it has access to all the parent variables, or use a struct to pass the data.

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+        var widthInTilesScaled = DivCeil(maskWidth / smallScale, maskTileSize);
+        var widthInTilesFull = DivCeil(maskWidth, maskTileSize);
+
+        for (uint y = 0; y < maskHeight; y++)
+        {
+            for (uint x = 0; x < maskWidth; x++)
             {
-                return q00;
+                DecodeSingle(ref maskData, maskWidth, maskHeight, atlasData, atlasWidth, atlasHeight, x, y, tileData, maskTileSize, maskIndex, smallOffset, smallScale, widthInTilesScaled);
+                DecodeSingle(ref maskData, maskWidth, maskHeight, atlasData, atlasWidth, atlasHeight, x, y, tileData, maskTileSize, maskIndex, 0, 1, widthInTilesFull);
```
notaspirit: Aren't both writing to the same index in maskData?

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+        var fullResBuffer = new byte[maskWidth * maskHeight];
 
-            if ((args.AsList && subDir is null) || (!args.AsList && outfile.Directory is null))
-            {
-                _loggerService.Error("directory was null");
-                return false;
-            }
+        for (var i = 0; i < maskCount; i++)
+        {
+            Array.Clear(fullResBuffer, 0, fullResBuffer.Length);
```
notaspirit: The clear is redundent, as Decode already overwrite the entire buffer

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+
+    private static uint DivCeil(uint l, uint r)
+    {
+        return (l + r - 1) / r;
+    }
+
+    private static uint CountBits(uint v)
+    {
+        return (uint)System.Numerics.BitOperations.PopCount(v);
+    }
```
notaspirit: as these are oneliners they should have expression bodies according to our style rules

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+        }
+
+        DirectoryInfo? subDir = null;
+        if (args.AsList)
+        {
+            subDir = new DirectoryInfo(Path.ChangeExtension(outfile.FullName, null) + "_layers");
+            if (!subDir.Exists)
+            {
+                Directory.CreateDirectory(subDir.FullName);
             }
```
notaspirit: You don't need all of this, just call `Directory.CreateDirectory(Path.ChangeExtension...)` directly. 
The method only creates a new directory if it doesn't already exist. It also returns a directory info so you don't need to manually create one.

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
         }
 
-        private static void DecodeSingle(ref byte[] maskData, uint maskWidth, uint maskHeight, byte[] atlasData, uint atlasWidth, uint atlasHeight, uint x, uint y, uint[] tilesData, uint maskTileSize, int maskIndex, uint tilesOffset, uint smallScale)
+        if ((args.AsList && subDir is null) || (!args.AsList && outfile.Directory is null))
         {
-            var widthInTiles = DivCeil(maskWidth / smallScale, maskTileSize);
+            _loggerService.Error("directory was null");
+            return false;
+        }
```
notaspirit: What's up with this check? You're checking the same condition as above and checking for non null on a variable that is guaranteed to be not null if that path is hit.

But also implementing an argument check that should be done at the very beginning of the method.

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+
+                dst[rowOffset + x] = src[srcX + srcY * srcW];
             }
-            return count;
         }
+        return dst;
+    }
 
-        private static void Decode(ref byte[] maskData, uint maskWidth, uint maskHeight, uint mWidthLow, uint mHeightLow, byte[] atlasData, uint atlasWidth, uint atlasHeight, uint[] tileData, uint maskTileSize, int maskIndex)
+    private static byte[] CreateWkPreviewPng(byte[] gray, int width, int height)
```
notaspirit: Just `gray` isn't very descriptive for a parameter, especially as it's just a `byte[]` and the contents could have an arbitrary shape

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
-            var xi = (int)(x & x1);
-            var y1 = (1 << (int)sy) - 1;
-            var yi = (int)(y & y1);
+            // Simplified header - only basic info, no original atlas/mask dimensions
+            var headerLines = new List<string>
+            {
+                "# MLMask Export",
+                "# Layer image files (must be in correct order)",
+                ""
+            };
```
notaspirit: Adding anything to the file that isn't path to an image layer is not backwards compatible as the previous version of wkit can't read that. 
Not to mention that this added content isn't useful.

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+        if ((args.AsList && subDir is null) || (!args.AsList && outfile.Directory is null))
         {
-            var widthInTiles = DivCeil(maskWidth / smallScale, maskTileSize);
+            _loggerService.Error("directory was null");
+            return false;
+        }
+
+        var cnt = 0;
+        var masks = new List<string>();
+        var layerResolutions = new List<string>();
```
notaspirit: this is only ever updated without being read

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,481 @@
...
+    private const uint ATLAS_TILE_PADDING = 2;
 
-            uint maskTileSize = blob.Header.MaskTileSize;
+    // Bit layout of tileDecl (from multilayer mask tiles data)
+    private const uint TILE_PARAM_MASK = 0x3FF;   // 10 bits (0-9)
+    private const int  TILE_DX_SHIFT   = 0;       // bits 0-9  → X offset in atlas
+    private const int  TILE_DY_SHIFT   = 10;      // bits 10-19 → Y offset in atlas
+    private const int  TILE_SX_SHIFT   = 20;      // bits 20-23 → X scale/shift
+    private const int  TILE_SY_SHIFT   = 24;      // bits 24-27 → Y scale/shift
+    private const uint TILE_S_MASK     = 0xF;     // 4 bits
```
notaspirit: the naming convention for private conts is s_ + camel case so e.g. `s_tileSMask`

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
```diff
@@ -33,283 +27,220 @@ public MLMASK(MlMaskContainer mlMask, ILoggerService logger)
...
+                if (string.IsNullOrWhiteSpace(trimmed) || trimmed.StartsWith("#"))
+                {
+                    continue;
+                }
+
+                filePaths.Add(trimmed);
+            }
+
+            if (filePaths.Count == 0)
+                throw new WolvenKitException(0x2002, "No layer files specified in masklist.");
```
notaspirit: all blocks should have braces according to your style

notaspirit: the error code is wrong, 0x2002 is for the power of 2 requirement not layer files missing entirely

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
```diff
@@ -33,283 +27,220 @@ public MLMASK(MlMaskContainer mlMask, ILoggerService logger)
...
+
+            var files = filePaths.Select(x => Path.Combine(baseDir.FullName, x)).ToList();
```
notaspirit: it'd be cleaner if all the file path resolution was done in one place (meaning get the rooted path when reading it instead of iterating over a second time)

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
```diff
@@ -33,283 +27,220 @@ public MLMASK(MlMaskContainer mlMask, ILoggerService logger)
...
 
-            (uint imageDimensionX, uint imageDimensionY) = (0, 0);
+            // Load all layers from files
             foreach (var f in files)
             {
                 if (!File.Exists(f))
-                {
-                    throw new FileNotFoundException($"Line{{lineIdx}}: \"{f}\" Make sure the file path is valid and exists (paths are specified line by line in ascending layer order in masklist)");
-                }
+                    throw new FileNotFoundException($"File not found: {f}");
```
notaspirit: blocks should have braces here too

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
```diff
@@ -33,283 +27,220 @@ public MLMASK(MlMaskContainer mlMask, ILoggerService logger)
...
-                {
-                    _logger.Error($"\"{f}\" could not be loaded!");
-                    continue;
-                }
-
-                if (image.Metadata.Format != DXGI_FORMAT.DXGI_FORMAT_R8_UNORM)
-                {
-                    image.Convert(DXGI_FORMAT.DXGI_FORMAT_R8_UNORM);
-                }
+                    using var image = RedImage.LoadFromFile(f) ?? throw new WolvenKitException(0x2001, $"Could not load image: {f}");
```
notaspirit: misusing error codes here too, 0x2001 is specifically for an invalid color space

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
```diff
@@ -33,283 +27,220 @@ public MLMASK(MlMaskContainer mlMask, ILoggerService logger)
...
 
-        #region Methods
+        internal static bool IsBlankLayer(byte[] pixels)
+        {
+            if (pixels == null || pixels.Length == 0) return true;
+            var first = pixels[0];
+            for (int i = 1; i < pixels.Length; i++)
+                if (pixels[i] != first) return false;
+            return true;
+        }
```
notaspirit: method is unused

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
```diff
@@ -33,283 +27,220 @@ public MLMASK(MlMaskContainer mlMask, ILoggerService logger)
...
-                }
+                    using var image = RedImage.LoadFromFile(f) ?? throw new WolvenKitException(0x2001, $"Could not load image: {f}");
 
-                if (image.Metadata.Width % 2 != 0 || image.Metadata.Height % 2 != 0)
-                {
-                    throw new WolvenKitException(0x2002,
-                        $"Texture {f}: width={image.Metadata.Width},height={image.Metadata.Height} must have dimensions divisible by 2");
-                }
+                    if (image.Metadata.Format != DXGI_FORMAT.DXGI_FORMAT_R8_UNORM)
+                        image.Convert(DXGI_FORMAT.DXGI_FORMAT_R8_UNORM);
```
notaspirit: braces style again

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
```diff
@@ -33,283 +27,220 @@ public MLMASK(MlMaskContainer mlMask, ILoggerService logger)
...
+                            whiteH = (uint)probe.Metadata.Height;
+                        }
+                    }
+                    catch { }
+                }
+
+                var whitePixels = new byte[whiteW * whiteH];
+                Array.Fill(whitePixels, (byte)255);
+                textures.Add(new RawTexContainer { Width = whiteW, Height = whiteH, Pixels = whitePixels });
             }
```
notaspirit: I'd be better to use WidthLow and HeightLow for the creation of the white layer so it's only as big as it needs to be

notaspirit (changes requested):
There are a couple of points that need to be adjusted in the logic (like not writing any additional data to the list file, using correct error codes etc.) The vast majority of issues are style related or misc. I've annoted some of them but there are just too many, assuming your ide is correctly set up it should already tell you about them (like missing braces, wrong variable name style etc).

Ametis81:
Refactor MLMask import/export: clean .masklist format, fix error codes, remove dead code, enforce style, and improve white layer handling

**`WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs`**

- Removed significant dead code in GetRedImages (unused atlas/tile decoding).
- Fixed redundant Array.Clear + restored correct zero-initialization of layer buffers (fixes white background bug).
- Changed error code 0x3001 → 0x3101 (avoid ArchiveXL collision) + added documentation.
- Converted one-liner helpers to expression-bodied members.
- Renamed private constants to follow s_ + camelCase convention.
- Simplified directory creation and removed convoluted null checks.
- Improved white base layer creation: now created at the actual low-resolution size (minimum layer dimensions).
- Added clarifying comments for double DecodeSingle calls.

**`WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs`**

- Fixed incorrect error code usage (0x2001 and 0x2002).
- Added proper error code documentation and ranges.
- Enforced consistent brace style across the file (aggressive cleanup in packing logic).
- Refactored file path resolution to single-pass.
- Removed unused IsBlankLayer method.
- Renamed vague parameter gray → grayData.
- Improved white layer handling logic.

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: removed

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: removed

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: removed

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: changed to 0x3101

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: Now:

```
private readonly record struct LayerBuffer(byte[] Buffer, uint Width, uint Height);

private static IEnumerable<LayerBuffer> DecodeLayerBuffers(...)
{
    ...
    yield return new LayerBuffer(layerBuffer, outWidth, outHeight);
}
```

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81:
```
// Decode using low-res tile map first (fallback)
DecodeSingle(..., smallOffset, smallScale, widthInTilesScaled);

// Then decode using full-res tile map — this can overwrite with higher quality data
DecodeSingle(..., 0, 1, widthInTilesFull);
```

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81:
```
for (var i = 0; i < maskCount; i++)
{
    Array.Clear(fullResBuffer, 0, fullResBuffer.Length); // Ensure pixels without tile data are black (0)
    Decode(...);
}
```
At first, I removed it. After i did that I was getting white background for masks (empty spaces just filled with white pixels), so I put it back because Decode() only writes pixels that have data. The rest should be black (0).

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: fixed

```
private static uint DivCeil(uint l, uint r) => (l + r - 1) / r;
private static uint CountBits(uint v) => (uint)System.Numerics.BitOperations.PopCount(v);
```

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: This code wasn't mine (c)

I can change it to:

```
if (args.AsList)
{
    var layersDir = Path.ChangeExtension(outfile.FullName, null) + "_layers";
    subDir = Directory.CreateDirectory(layersDir);
}
```

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: This code wasn't mine (c)

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: Now:
```
private const uint s_atlasTilePadding = 2;
private const uint s_tileParamMask = 0x3FF;
private const int  s_tileDxShift   = 0;
```

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: removed

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: removed, though i thought it was there all the time

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: Renamed the parameter from _gray_ to _grayData_

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: done

Ametis81: Right?

0x2001 —  Invalid color space (R8)
0x2004 — No layers found in masklist
0x2005 — General image loading failure

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81:
```
foreach (var f in files) { textures.Add(...); }

uint targetLowW = textures.Min(t => t.Width);
uint targetLowH = textures.Min(t => t.Height);

if (needsLayer0)
{
    textures.Insert(0, new RawTexContainer 
    { 
        Width = targetLowW, 
        Height = targetLowH, 
        Pixels = whitePixels 
    });
}
```

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: done

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: removed

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: ?
image

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: done

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: Now we resolve the full path immediately while reading:

```
var files = new List<string>();

foreach (var line in lines)
{
    var trimmed = line.Trim();
    if (string.IsNullOrWhiteSpace(trimmed) || trimmed.StartsWith("#"))
        continue;

    // Resolve to full path immediately (single pass)
    files.Add(Path.Combine(baseDir.FullName, trimmed));
}

if (files.Count == 0)
{
    throw new WolvenKitException(0x2004, "No layer files specified in masklist.");
}
```

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: Make Decode a nested method (local function)?  its easier but is it ok if it becomes less reusable? With `struct` I need to write more code lines :)

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
notaspirit: Where is that table from? The error codes are registered in `LogCodeHelper` where these are the mappings for the codes in your screenshot.

```C#

 s_mapping.Add(0x2001,
            "https://wiki.redmodding.org/cyberpunk-2077-modding/modding-guides/textures-and-luts/images-importing-editing-exporting#invalid-color-space");
s_mapping.Add(0x2002,
    "https://wiki.redmodding.org/cyberpunk-2077-modding/modding-guides/textures-and-luts/images-importing-editing-exporting#must-have-dimensions-in-powers-of-2");
s_mapping.Add(0x2003,
    "https://wiki.redmodding.org/cyberpunk-2077-modding/modding-guides/textures-and-luts/images-importing-editing-exporting#all-images-must-be-the-same-size");
s_mapping.Add(0x2004, // morphtarget: no chunks
    "https://wiki.redmodding.org/cyberpunk-2077-modding/modding-guides/npcs/a-new-head-for-v#step-6-optional-disabling-the-character-creator");
s_mapping.Add(0x2005, // mesh: bone mismatch
    "https://wiki.redmodding.org/cyberpunk-2077-modding/for-mod-creators-theory/3d-modelling/troubleshooting-your-mesh-edits#bones-not-found-in-import-mesh-es");

```

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
```diff
@@ -14,16 +13,20 @@
...
 using WolvenKit.RED4.Types;
 
+/// <summary>
+/// Error code ranges used in multilayer mask import:
+/// 0x2000–0x20FF : Import / masklist parsing and image loading errors
+///   0x2001 = Invalid color space (must be R8/grayscale)
+///   0x2002 = Reserved (power-of-2 requirement)
+///   0x2004 = No layers found in masklist
+///   0x2005 = General image loading failure
+/// </summary>
```
notaspirit: This summary block isn't relevant, nor accurate as it does not match the codes defined in `LogCodeHelper`

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
```diff
@@ -33,283 +36,206 @@ public MLMASK(MlMaskContainer mlMask, ILoggerService logger)
...
+
+                // Resolve to full path immediately (single pass)
+                files.Add(Path.Combine(baseDir.FullName, trimmed));
             }
 
-            (uint imageDimensionX, uint imageDimensionY) = (0, 0);
+            if (files.Count == 0)
+            {
+                // 0x2004 = No layers found in masklist (0x2002 is reserved for power-of-2 requirement errors)
+                throw new WolvenKitException(0x2004, "No layer files specified in masklist.");
```
notaspirit: still incorrect error code, 0x2004 is about no morth target chunks as defined in the LogCodeHelper

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,464 @@
...
 using WolvenKit.Common.Model.Arguments;
 using WolvenKit.RED4.Archive.CR2W;
 using WolvenKit.RED4.CR2W;
 using WolvenKit.RED4.Types;
 
-namespace WolvenKit.Modkit.RED4
+/// <summary>
+/// Error code ranges used in multilayer mask processing:
+/// 0x3100–0x31FF : Core decoding, BC4 unpacking and layer processing errors
+/// </summary>
```
notaspirit: this summary isn't relevant

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
notaspirit: The method is already private and only used in one place, with that many parameters it's already not very reusable anyway. 
Having that many parameters just makes it more difficult to read.

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
notaspirit: It'd be better if `DecodeSingle` returned an explicit boolean indicating the success state, that way the logic is more explicit and you avoid decoding twice when you only need to decode once.

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
notaspirit: As mentioned in the other comment all the codes are in `LogCodeHelper`

notaspirit (changes requested):
You don't have to use the wolvenkit exception if you don't have an associated error code or wiki page to link to, all the currently available codes are in `LogCodeHelper`, if you need additional ones you need to register them there. 

There are still lots of style issues all over, with riders refactoring tools it should be simple to apply the style consistently across the entire files.

Also this pr should have a changelog for the app and nuget packages.

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: part of them were imagined :) so i need to edit `LogCodeHelper` too to add my new errors?

notaspirit: yeah if you want additional error codes you need to register them

Ametis81: all in modkit section? (beginning with 3003)? or leave two of them  related to image import inside 2000 (import image failure and masklist has no layers specified)

Ametis81:
i will add error codes later, still don't know all of them should be under 3000 numbers, or two of them should be under 2000 (image related errors)

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81:
```
private static bool DecodeSingle(...)   // now returns bool
{
    ...
    if (...) return false;
    if (...) return false;
    
    ...write pixel ...
    return true;
}
```

done

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: Removed

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: changed error code (added to LogCodeHelper)

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: Removed

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: added them to LogCodeHelper

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: E,g. without Decode:

image

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81:
```
var fullResBuffer = new byte[maskWidth * maskHeight];
// Local function: decodes one layer into fullResBuffer.
void Decode(int maskIndex)
{
    var widthInTiles0 = DivCeil(maskWidth, maskTileSize);
    var heightInTiles0 = DivCeil(maskHeight, maskTileSize);
    var smallOffset = widthInTiles0 * heightInTiles0;
    var smallScale = (maskWidthLow == 0 || maskWidth < maskWidthLow) ? 1u : maskWidth / maskWidthLow;

    var widthInTilesScaled = DivCeil(maskWidth / smallScale, maskTileSize);
    var widthInTilesFull = DivCeil(maskWidth, maskTileSize);

    for (uint y = 0; y < maskHeight; y++)
    {
        for (uint x = 0; x < maskWidth; x++)
        {
            // Decode using low-res tile map first (fallback)
            DecodeSingle(ref fullResBuffer, maskWidth, maskHeight, atlasRaw, atlasWidth, atlasHeight, x, y, tiles, maskTileSize, maskIndex, smallOffset, smallScale, widthInTilesScaled);

            // Then decode using full-res tile map — this can overwrite with higher quality data
            // if the layer has dedicated high-resolution tiles for this pixel.
            DecodeSingle(ref fullResBuffer, maskWidth, maskHeight, atlasRaw, atlasWidth, atlasHeight, x, y, tiles, maskTileSize, maskIndex, 0, 1, widthInTilesFull);
        }
    }
}

for (var i = 0; i < maskCount; i++)
{
    Array.Clear(fullResBuffer, 0, fullResBuffer.Length); // Ensure pixels without tile data are black (0)
    Decode(i);

    var hasHighRes = HasLayerHighResolutionData(tiles, maskWidth, maskHeight, maskTileSize, i);
    ...
}
```

changelog/unreleased/2922.yaml
```diff
@@ -0,0 +1,6 @@
+changes:
+    - type: "fixed | changed"
```
notaspirit: you can only pick one

changelog/unreleased/2922.yaml
```diff
@@ -0,0 +1,6 @@
...
+      packages: ["App", "Nuget Packages"]
```
notaspirit: It does affect the CLI as well as it does use ModKit

WolvenKit/Helpers/LogCodeHelper.cs
```diff
@@ -26,12 +26,15 @@ static LogCodeHelper()
             "https://wiki.redmodding.org/cyberpunk-2077-modding/modding-guides/npcs/a-new-head-for-v#step-6-optional-disabling-the-character-creator");
         s_mapping.Add(0x2005, // mesh: bone mismatch
             "https://wiki.redmodding.org/cyberpunk-2077-modding/for-mod-creators-theory/3d-modelling/troubleshooting-your-mesh-edits#bones-not-found-in-import-mesh-es");
+        s_mapping.Add(0x2006, "https://wiki.redmodding.org/wolvenkit/wolvenkit-app/error-codes"); // MLMask: No layer files specified in masklist
+        s_mapping.Add(0x2007, "https://wiki.redmodding.org/wolvenkit/wolvenkit-app/error-codes"); // MLMask: General image loading failure
```
notaspirit: it doesn't really make sense to redirect the user to the generic error codes wiki page for these specific issues, as the error codes page primarily functions under the assumption that it is a bug while this is about an invalid / broken file.

If there isn't a specific page to link to about the issue I would prefer a regular exception without an error code.

notaspirit:
lmk when it's ready for the next review round, preferably via the request rereview button in the top right in the reviewers section.

Ametis81:
> lmk when it's ready for the next review round, preferably via the request rereview button in the top right in the reviewers section.

done

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,482 @@
...
                 {
-                    Directory.CreateDirectory(subDir.FullName);
+                    // Decode using low-res tile map first (fallback)
+                    DecodeSingle(ref fullResBuffer, maskWidth, maskHeight, atlasRaw, atlasWidth, atlasHeight, x, y,
+                        tiles, maskTileSize, maskIndex, smallOffset, smallScale, widthInTilesScaled);
+
+                    // Then decode using full-res tile map — this can overwrite with higher quality data
+                    // if the layer has dedicated high-resolution tiles for this pixel.
+                    DecodeSingle(ref fullResBuffer, maskWidth, maskHeight, atlasRaw, atlasWidth, atlasHeight, x, y,
+                        tiles, maskTileSize, maskIndex, 0, 1, widthInTilesFull);
```
notaspirit: this still double decodes instead of only using the fallback when the primarily failed (using the boolean return value as the signal)

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,482 @@
...
+            File.WriteAllLines(maskListPath, masks);
+        }
 
-            var q00 = atlasData[ux + 0 + ((uy + 0) * atlasWidth)];
-            var q10 = atlasData[ux + 1 + ((uy + 0) * atlasWidth)];
-            var q01 = atlasData[ux + 0 + ((uy + 1) * atlasWidth)];
-            var q11 = atlasData[ux + 1 + ((uy + 1) * atlasWidth)];
+        return true;
+    }
+    // The import side ignores lines starting with '#' or containing '='.
```
notaspirit: what does this comment belong to?

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
```diff
@@ -1,288 +1,482 @@
...
+                return true;
             }
+        }
+
+        return false;
+    }
+
+    private static bool DecodeSingle(ref byte[] maskData, uint maskWidth, uint maskHeight, byte[] atlasData,
+        uint atlasWidth, uint atlasHeight, uint x, uint y, uint[] tilesData, uint maskTileSize, int maskIndex,
+        uint tilesOffset, uint smallScale, uint widthInTiles)
```
notaspirit: again this should either be one or multiple structs / records or be a local function that uses the parent context directly.

notaspirit (changes requested):

WolvenKit/Helpers/LogCodeHelper.cs
Ametis81: removed

changelog/unreleased/2922.yaml
Ametis81: done

changelog/unreleased/2922.yaml
Ametis81: done

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: better?

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: removed

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: changed

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81:
```c#
bool DecodeSingle(uint x, uint y, int maskIndex, uint tilesOffset, uint smallScale, uint widthInTiles)
{
    var xTile = x / maskTileSize / smallScale;
    var yTile = y / maskTileSize / smallScale;

    var tileIndex = (widthInTiles * yTile) + xTile + tilesOffset;

    if ((tileIndex * 2) + 1 >= tiles.Length)
    {
        return false;
    }

    var paramOffset = tiles[tileIndex * 2];
    var paramBits = tiles[(tileIndex * 2) + 1];

    if ((uint)(paramBits & (1 << maskIndex)) == 0U)
    {
        return false;
    }

    var extraAdd = CountBits((uint)(paramBits & ((1 << maskIndex) - 1)));

    uint tileDecl = 0;
    if (paramOffset + extraAdd < tiles.Length)
    {
        tileDecl = tiles[paramOffset + extraAdd];
    }

    // Extract atlas position and scaling factors from the packed tile declaration.
    var dx = (tileDecl >> s_tileDxShift) & s_tileParamMask;
    var dy = (tileDecl >> s_tileDyShift) & s_tileParamMask;
    var sx = (tileDecl >> s_tileSxShift) & s_tileSMask;
    var sy = (tileDecl >> s_tileSyShift) & s_tileSMask;

    var atlasTileSize = maskTileSize + s_atlasTilePadding;

    // Use Min to avoid reading into the right/bottom padding area of atlas tiles
    var localX = Math.Min((x >> (int)sx) % maskTileSize, maskTileSize - 1);
    var localY = Math.Min((y >> (int)sy) % maskTileSize, maskTileSize - 1);

    var ux = localX + 1 + (dx * atlasTileSize);
    var uy = localY + 1 + (dy * atlasTileSize);

    var p = atlasRaw[ux + (uy * atlasWidth)];
    fullResBuffer[x + (y * maskWidth)] = p;

    return true;
}

```

```c#
void Decode(int maskIndex)
{
    var widthInTiles0 = DivCeil(maskWidth, maskTileSize);
    var heightInTiles0 = DivCeil(maskHeight, maskTileSize);
    var smallOffset = widthInTiles0 * heightInTiles0;
    var smallScale = (maskWidthLow == 0 || maskWidth < maskWidthLow) ? 1u : maskWidth / maskWidthLow;

    var widthInTilesScaled = DivCeil(maskWidth / smallScale, maskTileSize);
    var widthInTilesFull = DivCeil(maskWidth, maskTileSize);

    for (uint y = 0; y < maskHeight; y++)
    {
        for (uint x = 0; x < maskWidth; x++)
        {
            // First try to decode using full-res tile map (primary)
            bool decoded = DecodeSingle(x, y, maskIndex, 0, 1, widthInTilesFull);

            // If full-res didn't have data for this layer/pixel, fall back to low-res tile map
            if (!decoded)
            {
                DecodeSingle(x, y, maskIndex, smallOffset, smallScale, widthInTilesScaled);
            }
        }
    }
}
```

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: removed 2005 / 2006

WolvenKit/Helpers/LogCodeHelper.cs
Ametis81: still left. OK?

image

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: changed

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: changed

WolvenKit.Modkit/RED4/Tools/MlmaskImportTools.cs
Ametis81: removed

WolvenKit.Modkit/RED4/Tools/MlmaskTools.cs
Ametis81: also removed doubling of `DecodeSingle`

WolvenKit/Helpers/LogCodeHelper.cs
notaspirit: as it doesn't have a dedicated section / page it should also just be a regular exception

notaspirit (changes requested):
internal methods should be at the end of the method after the final return statement (or the last line or block of code that is in the root of the parent method). 
Other than that and the error code not having a dedicated section that gave an answer to in that thread it's good to merge.

WolvenKit/Helpers/LogCodeHelper.cs
Ametis81: done

notaspirit (approved):

Ametis81:
> internal methods should be at the end of the method after the final return statement (or the last line or block of code that is in the root of the parent method). Other than that and the error code not having a dedicated section that gave an answer to in that thread it's good to merge.

done

PR #2923, opened 2026-05-15
Update RDTMeshViewModel.cs 

Ametis81:
FIxed material loading for meshes in preview window with Generate materials button

Merge to dev


WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
```diff
@@ -1710,47 +1710,33 @@ public void LoadMaterials()
     {
         if (SelectedAppearance == null)
         {
-            Parent.GetLoggerService().NotNull().Warning($"No material selected!");
```
manavortex, line 1713: Instead of `.NotNull().Warning`, you might rather want to use `?.Warning`. First will throw an exception if the object is null, second will just do nothing.

manavortex (changes requested):

WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
Ametis81: should i use this in all `NotNull()` lines?

manavortex: Yeah :)

Ametis81: image
and here?

manavortex: Yeah, unless you actually want the exception, but it makes no sense here

Ametis81: ok, only warnings

Ametis81: Done

manavortex: No - the `.NotNull()` will cause the C# code to throw an exception when the logger is null. Unless that is what you want, you should replace it with a `?` in every instance where it's currently used.

Ametis81: better? :)

manavortex: Wayy better thanks :)

WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
```diff
@@ -1900,415 +1914,309 @@ public async ValueTask LoadMaterial(WolvenKit.App.Models.Material? material)
...
-                destBitmap.Save(filename_b, ImageFormat.Png);
-            }
-            catch (Exception e)
-            {
-                Parent.GetLoggerService().Error(e.Message);
-            }
-            finally
-            {
-                destBitmap.Dispose();
+                try { destBitmap.Save(filename_b, ImageFormat.Png); } catch (Exception e) { Parent.GetLoggerService().Error(e.Message); } finally { destBitmap.Dispose(); }
```
manavortex: Can you please use the same formatting as the rest of the code? Makes it easier to actually spot changes. I've seen if statements above that aren't in Allman style, too.

WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
```diff
@@ -1900,415 +1914,309 @@ public async ValueTask LoadMaterial(WolvenKit.App.Models.Material? material)
...
-                Parent.GetLoggerService().Error(e.Message);
-            }
-            finally
-            {
-                rmBitmap.Dispose();
-            }
+            #endregion
+            #region diffuseMaps
+        DiffuseMaps:
+            if (File.Exists(filename_d)) goto NormalMaps;
```
manavortex: That's what I mean, your formatter has changed this for the worse :)

WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
```diff
@@ -1900,415 +1914,309 @@ public async ValueTask LoadMaterial(WolvenKit.App.Models.Material? material)
...
-        }
-
-        if (dictionary.TryGetValue("DiffuseTexture", out var diffuse) && diffuse is CResourceReference<ITexture> crrd)
-        {
-            var xbm = Parent.GetFileFromDepotPathOrCache(crrd.DepotPath);
-
-            if (xbm?.RootChunk is not ITexture it)
+            if (dictionary.TryGetValue("ParalaxTexture", out var paralax) && paralax is CResourceReference<ITexture> crrp)
             {
-                goto NormalMaps;
```
manavortex, line 2134: What's the reasoning behind making the code run through all the extra checks instead of jumping straight to normal? Did you find a new edge case? :)

WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
Ametis81: Better? Formatting / Styling

manavortex: Yesn't - now it has the newlines again, but it's missing the braces. This makes it hard to compare what has actually changed and what hasn't. There are try-catch blocks around L2000 somewhere that have been squished into a single line, too.

Ametis81: Damn :) VS code cleaner is playing tricks with me

Ametis81: :3

WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
```diff
@@ -1772,543 +1766,541 @@ public async ValueTask LoadMaterial(WolvenKit.App.Models.Material? material)
...
+            }
         #endregion
-        return;
+
+        SkipNormals:
+            DispatcherHelper.RunOnMainThread(() => SetupPBRMaterial(material.Name, true));
+        }
+        catch (Exception ex)
+        {
+            Parent.GetLoggerService().Error($"Failed to load material '{material?.Name}': {ex.Message}");
```
manavortex, line 2256: Maybe it's worth putting the try/catch block inside the loop? As it is right now, a single error when processing anything will cause the entire material generation to fail. (Still better than the exception not being caught, though, so if you want to leave it like this, I'm fine with it)

WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
Ametis81: Done

manavortex (approved):

PR #2924, opened 2026-05-19
Fix: inkatlas generator overwriting

manavortex:
# Inkatlas generator user feedback

**Fixed:**
- Inkatlas generator will now delete file before overwriting and provide better feedback

### DIFF
--- a/WolvenKit.App/Helpers/InkatlasImageGenerator.cs
+++ b/WolvenKit.App/Helpers/InkatlasImageGenerator.cs
@@ -14,6 +14,11 @@ namespace WolvenKit.App.Helpers;
 
 public static class InkatlasImageGenerator
 {
+    /// <summary>
+    /// Generates an inkatlas from input parameters.
+    /// </summary>
+    ///
+    /// <returns>A boolean if the write operation to the original path was successful</returns>
     public static bool GenerateAtlas(
         string pngFolder,
         string relativeSourcePath,
@@ -33,7 +38,7 @@ public static class InkatlasImageGenerator
         // Ensure output directory exists
         Directory.CreateDirectory(absoluteSourcePath);
         var destFileName = Path.Combine(absoluteSourcePath, $"{atlasFileName.Replace(".inkatlas", "")}.inkatlas");
-        var fileWasWritten = true;
+        var writtenToOriginalPath = true;
 
         if (File.Exists(destFileName))
         {
@@ -44,12 +49,12 @@ public static class InkatlasImageGenerator
             catch
             {
                 destFileName = destFileName.Replace(".inkatlas", "_new.inkatlas");
-                fileWasWritten = false;
+                writtenToOriginalPath = false;
             }
         }
         var cr2WFile = new CR2WFile() { RootChunk = inkatlas };
         cr2WTools.WriteCr2W(cr2WFile, destFileName);
-        return fileWasWritten;
+        return writtenToOriginalPath;
     }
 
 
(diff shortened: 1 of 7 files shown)


changelog/!unreleased.yaml
```diff
@@ -7,6 +7,16 @@ Template changelog entry:
       description: ""
 
 changes:
+    - type: "fixed"
+      packages: ["App"]
+      pr-number: 2917
+      author: "manavortex"
+      description: "Fixed 'create item codes from yaml'"
```
notaspirit: it's readding the entry for the other pr which is already long merged

WolvenKit.App/Helpers/InkatlasImageGenerator.cs
```diff
@@ -32,10 +32,24 @@ public static void GenerateAtlas(
...
+            {
+                destFileName = destFileName.Replace(".inkatlas", "_new.inkatlas");
+                fileWasWritten = false;
+            }
+        }
         var cr2WFile = new CR2WFile() { RootChunk = inkatlas };
-        cr2WTools.WriteCr2W(cr2WFile,
-            Path.Combine(absoluteSourcePath, $"{atlasFileName.Replace(".inkatlas", "")}.inkatlas"));
+        cr2WTools.WriteCr2W(cr2WFile, destFileName);
+        return fileWasWritten;
```
notaspirit: the naming of fileWasWritten is a bit misleading as it does write the file either way with only the location being different

WolvenKit.App/Helpers/InkatlasImageGenerator.cs
```diff
@@ -14,7 +14,7 @@ namespace WolvenKit.App.Helpers;
 
 public static class InkatlasImageGenerator
 {
-    public static void GenerateAtlas(
+    public static bool GenerateAtlas(
```
notaspirit, line 22: the concern I have with this is that bool as a return value of a method that generates something usually signals success / failure of the whole process rather than it writing to a different path.

It'd be clearer if it returned an enum, or perhaps even the path it saved it to (which would also have the added benefit of downstream systems being able to use it directly instead of having to rebuild it if they need to read the file).
As the return value is written to a variable before being used in an evaluation it shouldn't negatively impact the readability / complexity for existing consumers. 

But it'd also like to hear your thoughts on this.

notaspirit (changes requested):
The button is a bit off

Image

WolvenKit.App/Helpers/InkatlasImageGenerator.cs
manavortex: Pretty much every other solution seemed like complete overkill. And the boolean does signal success, if it's false, then the file wasn't written to where the user wanted it (but elsewhere, so that you don't have to repeat all your inputs). IMO working around this will make the code less maintainable.

notaspirit: fair enough, but as it is a public method it should then be annotated in the xml tags so it's easy to figure out without needing to read the method implementation

manavortex: I did that?

notaspirit: I don't see it in the version that github shows for this pr

notaspirit (changes requested):
the new button has different spacing compared to the existing one below that has the same layout

Image

merging main into this branch removed the changelog

notaspirit (approved):

PR #2927, opened 2026-05-23
Performance improvements

gistya:
Batch JSON loading & much faster project load/refresh.

This is a draft PR that will get split into more and thinned down. I have done a lot of local testing with it to iron it out. If anyone wants to get on a discord call to discuss any of the changes or you have any feedback, please reach out.

**Implemented:**
- Reworks project loading to first arrange all files into the data structures and then add them all to the UI in one update, so that very large projects load in seconds rather than hours/days. 
- Implements batch adding of files to a project, dramatically speeding up adding many files to a project. 
- Implements batch conversion of files to JSON, dramatically speeding up conversion of many files to JSON. They go in parallel and it flies.
- Adds a setting to turn off automatic expansion of all folders in ProjectExplorer when opening a mod. This speeds up loading of very large projects by a few seconds.
- Implements a new way of loading files into ProjectExplorer that bypasses File System Event monitoring entirely, for improved reliability and greatly enhanced speed.
- Implements better concurrency in the app for performance
- Improved timing of tasks to not impact key transitions framerates

**Fixes**
- Addresses inconsistencies with project loading and switching. Now the UI should accurately indicate when a project is loading or reloading. 

**New Tests**
- Adds integration and UI tests to cover impacted areas

**Additional notes:**
- Batching was not implemented for file changes or removals, though we could do this in a future PR. 
- A _fileProcessing dictionary was added in WatcherService to prevent conflicts between the Update and BatchUpdate threads, due to the fact that adding a bunch of files also triggers a bunch of change events. Ask me on discord if you need a more detailed explanation for this. (This was further mitigated with switch to non-file-system-based file adding.)
- Shaved ~3 seconds off loading a 349-file project.
- Now loads my 104,386-file project (with also that many JSONs) in 19 seconds instead of something like 19+ hours.


WolvenKit.App/Controllers/RED4Controller.cs
```diff
@@ -1071,7 +1095,7 @@ public bool AddToMod(IGameFile file, ArchiveManagerScope searchScope)
             {
                 using FileStream fs = new(diskPathInfo.FullName, FileMode.Create);
                 file.Extract(fs);
-                _loggerService.Info($"Added game file to project: {file.Name}");
+                //_loggerService.Info($"Added game file to project: {file.Name}");
```
gistya: Forgot to remove this line. It was causing major slowdowns. Now we do the logging elsewhere in a batch. Will remove before merge.

WolvenKit.App/Services/IProjectEvents.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.IO;
+using WolvenKit.Core.Interfaces;
+
+namespace WolvenKit.App.Services;
+
+/// <summary>
+/// Service for published events related to file operations.
+/// Provides an alternative to monitoring of file system events.
+/// </summary>
+public interface IProjectEvents
```
gistya, line 12: New type for publishing file additions to the Watcher so it does not have to get them from Windows.

WolvenKit.App/Services/ISettingsDto.cs
```diff
@@ -51,6 +51,7 @@ public interface ISettingsDto
     public uint TreeViewGroupSize { get; set; }
     public EditorDifficultyLevel DefaultEditorDifficultyLevel { get; set; }
     public string? TreeViewIgnoredExtensions { get; set; }
+    public bool AutoExpandAllFoldersOnLaunch { get; set; }
```
gistya, line 54: New setting to disable autoexpanding folders so that very big mods can load faster.

WolvenKit.App/Services/ProjectManager.cs
```diff
@@ -81,7 +81,7 @@ await ReadFromLocationAsync(location).ContinueWith(x =>
                 return;
             }
 
-
+            // Triggers OnPropertyChanged in ProjectExplorerViewModel, which in turn triggers a RefreshProjectData call.
```
gistya: Can remove this comment if needed.

WolvenKit.App/ViewModels/Tools/AssetBrowserViewModel.cs
```diff
@@ -579,24 +579,12 @@ private void GetFilesRecursive(RedFileSystemModel directory, Dictionary<ulong, I
...
-            await Task.Run(() => { _gameController.GetController().AddToMod(file); }, token);
-
-            Interlocked.Increment(ref progress);
-            _progressService.Report(progress / (float)files.Count);
-        });
-
-        _progressService.Completed();
-
+        _watcherService.Suspend();
+        await _gameController.GetController().AddToModAsync(files);
```
gistya, line 613: Now the gameController adds files as a batch asyncronously.

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -1004,53 +1009,103 @@ private async Task ConvertArchiveFile()
...
+                    finally
+                    {
+                        var currentProgress = Interlocked.Increment(ref progress);
+                        _progressService.Report(currentProgress / (float)allFiles.Count);
+                    }
+                });
+
+            _progressService.Completed();
+            // Return list of created JSON files
+            _projectEvents.PublishFilesImported(new FilesImportedMessage([],[.. createdJsonFiles.ToList()]));
```
gistya, line 1158: Publishes the processed files list to WatcherSevice so it can add them to the ViewModel's FileList and FileTree.

WolvenKit.Modkit/RED4/RedConverter.cs
```diff
@@ -92,9 +92,6 @@ public async Task<bool> ConvertToJsonAndWriteAsync(string infile, DirectoryInfo
                 var outpath = Path.Combine(outputDirInfo.FullName, $"{Path.GetFileName(infile)}.json");
                 using var fileStream = File.Open(outpath, FileMode.Create, FileAccess.Write);
                 await ConvertToJsonAsync(fileStream, infile);
-
-                _loggerService.Success($"Converted {infile} to {outpath}");
```
gistya, line 96: These logs slow down the app an incredible amount so I am batching them now.

WolvenKit.App/Models/FileSystemModel.cs
```diff
@@ -128,6 +135,10 @@ private void GetMetadata()
 
         GameRelativePath = string.Join(ResourcePath.DirectorySeparatorChar, hashParts);
         FullName = Path.Combine(root.RawRelativePath, RawRelativePath);
+        if (!FullName.StartsWith("D"))
+        {
+            Console.WriteLine("WTF");
+        }
```
gistya: Debugging code to be removed

WolvenKit.App/Helpers/DispatcherHelper.cs
```diff
@@ -33,4 +35,15 @@ public static void RunOnUIThread(this DispatcherObject d, Action action, Dispatc
...
+    /// Runs `action` on the main thread after specified delay, without blocking.
+    /// </summary>
+    /// <param name="action"></param>
+    /// <param name="millisecondsDelay"></param>
+    public static void DelayOnMainThread(Action action, int millisecondsDelay)
+    {
+        Task.Delay(millisecondsDelay)
+            .ContinueWith(_ => RunOnMainThread(action), TaskScheduler.Default);
+    }
+}
```
gistya, line 142: This `DelayOnMainThread` is used to get the progress indicator to start before a refresh operation in the ProjectExplorerViewModel. The progress bar is on low priority so it won't start before a higher priority task that it's supposed to be indicating the progress of, unless you give the thread some air with a non-blocking delay.

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -155,13 +163,19 @@ private void Svc_ThreadIdleTenSeconds(object? sender, EventArgs e)
...
+
+    private void InternalRefresh()
+    {
+        _progressService.IsIndeterminate = true;
+        _progressService.Status = EStatus.Running;
+        OnSetLoading?.Invoke(this, true);
+
+        // Give time for progress bar to update.
+        DispatcherHelper.DelayOnMainThread(() => RefreshProjectData(), 50);
     }
```
gistya: Consolidated refreshing of the ProjectExplorer to this method. It will first indicate a reload is occurring then dispatch the `RefreshProjectData` function. 

Here `OnSetLoading` tells the view to turn on the "Loading" indication and hide the tree views.

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -177,56 +191,47 @@ private void AppViewModel_OnInitialProjectLoaded(object? sender, EventArgs e)
     private void AppViewModelOnPropertyChanged(object? sender, PropertyChangedEventArgs e) =>
         CanScrollToOpenFile = HasSelectedItem && _appViewModel.ActiveDocument is not null;
 
-    private bool _loading;
-
```
gistya, line 181: This prior Loading property did not appear to be used anywhere at all so I removed it to avoid confusion.

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -177,56 +191,47 @@ private void AppViewModel_OnInitialProjectLoaded(object? sender, EventArgs e)
...
-    private void ProjectManager_OnPropertyChanged(object? sender, PropertyChangedEventArgs e)
-    {
-        if (e.PropertyName != nameof(ProjectManager.ActiveProject))
-        {
-            return;
-        }
-
-        RefreshProjectData();
-    }
```
gistya, line 201: Previously there was contention between OnProjectChange action and ProjectManager property change for Active project, leading to multiple triggers of `RefreshProjectData`, which in turn had several repeated calls to various methods that would result in multiple reloads of the project.  Getting rid of this.

WolvenKit.App/Controllers/RED4Controller.cs
```diff
@@ -74,20 +77,21 @@ public RED4Controller(
         _pluginService = pluginService;
         _modifierService = modifierService;
         _parserService = parserService;
+        _projectEvents = projectEvents;
     }
 
     public async Task HandleStartup()
     {
         if (!_initialized)
```
manavortex: Save the early return!!
```suggestion
        if (_initialized)
        {
          return;
        }
```

WolvenKit.Modkit/RED4/RedConverter.cs
seberoth: Please keep in mind, that .Modkit is also used by CLI and the nuget packages.
While I understand the problem its causing in the GUI, now it also won't get logged for the other projects.

Tests/WolvenKit.IntegrationTests/App/ViewModels/Tools/ProjectExplorerConvertToJsonUITests.cs
```diff
@@ -0,0 +1,162 @@
...
+        // Here, animMotionDb.Files should have lots of members, but only has 1.
+
+        var rightItems = vm.RightItems;
+        Assert.True(rightItems.Count> 5);
+
+        vm.RightItems.ToList().ForEach(item => item.IsChecked = true);
+        await vm.AddSelectedAsync();
+
+        // This assertion stil failing, will only have 1 file. Why?
+        Assert.True(_projectExplorerVm!.FileList.Count > 5);
```
gistya: This is where `animMotionDb` RedFileSystemModel should have a Files property with a bunch of files in it, but for some reason it always just loads with 1 file in there. If anyone can demystify what the test is not doing that would be in the real app that would be great.

Tests/WolvenKit.IntegrationTests/App/ViewModels/Tools/ProjectExplorerConvertToJsonUITests.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            state = progress.Status;
+            await Task.Delay(100);
+        }
+
+        var archives = vm._boundRootNodes.First();
+        AddDirs(archives, vm);
+
+        var baseDir = vm.LeftItems.First().Directories[Const.BaseDirName];
+        AddDirs(baseDir, vm);
```
gistya, line 147: Properties were made internal to allow this

WolvenKit.Modkit/RED4/RedConverter.cs
gistya: When i looked it seems it's already logging the file names (prolly happening twice now). Can you check?

PR #2931, opened 2026-05-30
better error msg for corrupt project file

manavortex:
# Better error msg for corrupt project file

**Implemented:**
- There will now be a better error message if project fails to load because of bad xml

### DIFF
--- a/WolvenKit.App/Services/ProjectManager.cs
+++ b/WolvenKit.App/Services/ProjectManager.cs
@@ -189,24 +189,6 @@ public partial class ProjectManager : ObservableObject, IProjectManager
             _loggerService.Error(e);
             return null;
         }
-
-        static bool DeleteLayoutFile(string layoutFilePath)
-        {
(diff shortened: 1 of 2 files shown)


WolvenKit.App/Services/ProjectManager.cs
```diff
@@ -131,7 +131,7 @@ await ReadFromLocationAsync(location).ContinueWith(x =>
         return null;
     }
 
-    private async Task<Cp77Project?> Load(string path)
+    private async Task<Cp77Project?> Load(string path, bool isRetry = false)
```
notaspirit: how is this related to the error handling change?

notaspirit (changes requested):

WolvenKit.App/Services/ProjectManager.cs
manavortex: it isn't, I took that code out. Thanks for spotting it

notaspirit (approved):

PR #2937, opened 2026-06-02
UI Automation Tests

gistya:
# UI Automation Tests

**Implemented:**
- Adds UI Automation tests based on [FLaUI framework](https://github.com/FlaUI/FlaUI) that controls the app through [Microsoft UIA metadata](https://learn.microsoft.com/en-us/dotnet/framework/ui-automation/ui-automation-overview) on views (which you can inspect at runtime using the free [FLaUInspect app](https://github.com/FlaUI/FlaUInspect)
- Adds `InspectableDataGrid` and `InspectableTreeGrid` subclass wrappers for SfDataGrid and SfTreeGrid to provide customized [AutomationPeer](https://learn.microsoft.com/en-us/dotnet/api/system.windows.automation.peers.automationpeer?view=windowsdesktop-10.0&viewFallbackFrom=net-8.0) objects. This solves the issue where Syncfusion views normally lack adequate metadata for UI automation tests. Now, the UI tests can get key info like number of elements in the tree/list view (not just the visible ones). 
- Adds `AutomationProperties.AutomationId` tags to certain View.xaml files to support the tests.
- Updates certain View.xaml files to use the new `Inspectable` types. (Note: this required adding `x:` in front of the `Name` property to resolve an xmls namespace issue... don't ask me how long that took to figure out)

**Fixed:**
- A lack of UI Automation tests 🗡️ 👯 

**Additional notes:**

The tests are structured in such a way that these test cases can be branched off from to cover additional use cases and user behaviors. Helper methods are available for tests to get key info from the views. Future tests can use the base test case as a branching-off point to load a new project without having to duplicate code.

Additionally, a README has been provided giving some basic guidance on how to approach writing these tests.


manavortex (approved):

changelog/!unreleased.yaml
```diff
@@ -27,3 +27,13 @@ changes:
...
+    - type: "fixed"
+      packages: ["App"]
+      pr-number: 2934
+      author: "manavortex"
+      description: "When generating a radio, the folder in resources will now be created"
+    - type: "added"
+      packages: ["Tests"]
+      pr-number: 2937
+      author: "vitamin arrr"
+      description: "Added UI Automation tests"
```
notaspirit: We keep the changelog only for changes that are relevant for the end user of the app and the nuget packages, for all the dev only changes (like docs, actions, tests etc) we don't have an explicit changelog. 
Also the author field should be the github username as that's what ends up being tagged in the release.

notaspirit (changes requested):
The ui tests currently look up `repoRoot\bin\x64\Debug\net8.0-windows10.0.17763\win-x64\WolvenKit.exe` by default when the build executable is under  `repoRoot\WolvenKit\bin\x64\Debug\net8.0-windows10.0.17763\win-x64\WolvenKit.exe` instead. 

Other than that looks good.

notaspirit (changes requested):
actually there is a regression in the ui visual related to the inspect able tree grid where the background color doesn't get applied the same way
with this pr:

Image

on main:

Image

changelog/!unreleased.yaml
gistya: ok will remove and fix the author name in the other pr.

notaspirit:
Regarding the file path lookup: it only fails when running it from the cli (and then it's fixed by setting the env variable) while working fine when run from an IDE. 
So not worth fixing now as all the tests have to be run locally anyway due to the game install requirement.

gistya:
> actually there is a regression in the ui visual related to the inspect able tree grid where the background color doesn't get applied the same way
> with this pr:

Turns out Syncfusion uses type-specific registration for some style metadata. So the subclasses didn't get all of it. 

The fix is to set some of the key DPs (at minimum Background color) in App.xml, so that any usage of <others:InspectableTreeGrid> or <others:InspectableDataGrid> gets the correct app-themed background. 

If we find any other regressions that's addressable in the same global setting.

In the actual C# classes I also added DefaultStyleKeyProperty.OverrideMetadata to the base type to help ensure the template/other theme parts come through. WPF doesn't like this so I put a pragma mark to suppress the complaint. (WPF thinks you'd never want a custom subclass to inherit style data I guess?)

changelog/!unreleased.yaml
```diff
@@ -27,3 +27,8 @@ changes:
       pr-number: 2934
       author: "manavortex"
       description: "When generating a radio, the folder in resources will now be created"
+    - type: "fixed"
+      packages: ["App"]
+      pr-number: 2934
+      author: "manavortex"
+      description: "When generating a radio, the folder in resources will now be created"
```
notaspirit: it's duplicating the changelog entry for a different pr

notaspirit (changes requested):
just this one minor thing with the changelog that I previous thought was the diff being weird and it's good to merge

notaspirit (approved):

PR #2939, opened 2026-06-02
Parallelize JSON Conversion

gistya:
# Parallelize JSON Conversion

**Implemented:**
- Parallelizes JSON conversion for much faster adding of JSON files to a project.
- Defers refresh of the views until all the files are added to avoid issues 
- WolvenKit.IntegrationTests added to cover this change
- Various minor changes that allow the Integration tests to work and not crash etc.
- Directly supplying a WatcherService in a few places to avoid a crash in testing where it's not ready yet

**Fixed:**
- The slowness of JSON conversion

**Additional notes:**
- As promised, this is one of the aspects that's being made into a smaller PR from the "Performance Improvements" PR


WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -991,7 +993,14 @@ private async Task ConvertArchiveFile()
 
         if (!IsShiftKeyPressed)
         {
-            await ConvertToJsonInternal(selection);
+            _deferredRefreshCts = new CancellationTokenSource();
+            if (BeginDeferredRefreshContext == null)
+            {
+                throw new Exception("Rendering context does not exist.");
```
manavortex: Would it make sense here to throw a WolvenkitException with "please restart WKit"?

manavortex (approved):

notaspirit (changes requested):
tbh this should have a changelog entry, speeding up json conversion considerably is noteworthy, as well as the change to the logging behavior in Core which is both a nuget package and part of the cli. 
Not like we have massive changelogs where we need to drop things in favor of brevity anyway.

gistya:
> tbh this should have a changelog entry, speeding up json conversion considerably is noteworthy, as well as the change to the logging behavior in Core which is both a nuget package and part of the cli. Not like we have massive changelogs where we need to drop things in favor of brevity anyway.

Agreed. I meant to add one but it got omitted, will resolve now.

notaspirit (approved):

PR #2942, opened 2026-06-04
Group recent projects on the welcome page

The-Destructor95:
# Group recent projects on the welcome page

**Implemented:**
- Project groups on the welcome page to keep the recent projects list tidy. Projects with no group stay under a default **"Ungrouped"** section, so nothing changes for existing users.
- Right-click a project to create a group, move it to an existing group, or remove it from its group.
- Collapse / expand each group.
- Delete a group (with confirmation) — its projects move back to "Ungrouped", nothing is deleted.
- Pick a group when creating a new project (optional, defaults to "Ungrouped").

**Additional notes:**
Groups are saved per project in `recentItems.json` (backward compatible, no migration). No change to the `.cpmodproj` format.

[PREVIEW](https://youtu.be/Q8iRER19ci4)

https://github.com/user-attachments/assets/3790851d-2089-4f3c-b2e2-8a53889a23af

### DIFF
--- a/WolvenKit.App/Models/ProjectManagement/RecentlyUsedItemModel.cs
+++ b/WolvenKit.App/Models/ProjectManagement/RecentlyUsedItemModel.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using CommunityToolkit.Mvvm.ComponentModel;
 
 namespace WolvenKit.App.Models.ProjectManagement;
@@ -16,6 +16,11 @@ public partial class RecentlyUsedItemModel : ObservableObject
 
     [ObservableProperty]
     private bool _isPinned;
+
+    // NOUVEAU : groupe du projet. null ou vide = "Sans groupe".
+    // Sérialisé automatiquement dans recentItems.json (rétrocompatible : absent = null).
+    [ObservableProperty]
+    private string? _group;
 
     public RecentlyUsedItemModel(string name, DateTime dateTime, DateTime lastOpened)
     {
--- a/WolvenKit.App/Models/ProjectManagement/RecentlyUsedItemModel.cs
+++ b/WolvenKit.App/Models/ProjectManagement/RecentlyUsedItemModel.cs
@@ -20,7 +20,7 @@ public partial class RecentlyUsedItemModel : ObservableObject
     // Project group. null or empty = "Ungrouped".
     // Serialized automatically in recentItems.json (backward compatible: absent = null).
     [ObservableProperty]
-    private string? _group;
+    private string _group = "";
 
     public RecentlyUsedItemModel(string name, DateTime dateTime, DateTime lastOpened)
     {
--- a/WolvenKit.App/Models/ProjectManagement/RecentlyUsedItemModel.cs
+++ b/WolvenKit.App/Models/ProjectManagement/RecentlyUsedItemModel.cs
@@ -17,8 +17,8 @@ public partial class RecentlyUsedItemModel : ObservableObject
     [ObservableProperty]
     private bool _isPinned;
 
-    // NOUVEAU : groupe du projet. null ou vide = "Sans groupe".
-    // Sérialisé automatiquement dans recentItems.json (rétrocompatible : absent = null).
+    // Project group. null or empty = "Ungrouped".
+    // Serialized automatically in recentItems.json (backward compatible: absent = null).
     [ObservableProperty]
     private string? _group;
 
--- a/WolvenKit.App/ViewModels/Dialogs/ProjectWizardViewModel.cs
+++ b/WolvenKit.App/ViewModels/Dialogs/ProjectWizardViewModel.cs
@@ -5,7 +5,6 @@ using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
-using System.Linq;
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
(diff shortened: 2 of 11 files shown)


WolvenKit.App/ViewModels/HomePage/Pages/WelcomePageViewModel.cs
```diff
@@ -269,8 +355,67 @@ private void RefreshRecentProjects() =>
...
+            var items = GetSortedItems(false, SelectedRecentOrder, RecentFilter);
+
             FancyProjects.Clear();
-            FancyProjects.AddRange(GetSortedItems(false, SelectedRecentOrder, RecentFilter));
+            FancyProjects.AddRange(items);
+
+            // Group by "Group". Projects without a group go into "Ungrouped",
+            // placed last; the other groups are sorted by name.
+            var groups = items
+                .GroupBy(p => string.IsNullOrWhiteSpace(p.Group) ? UngroupedName : p.Group!)
```
gistya: Here you use `Group` in a case-sensitive manner, but in ProjectWizardViewModel.cs you use:                 `.Distinct(StringComparer.InvariantCultureIgnoreCase)`

Should we be consistent about case sensitivity?

WolvenKit.App/Models/ProjectManagement/RecentlyUsedItemModel.cs
```diff
@@ -17,6 +17,11 @@ public partial class RecentlyUsedItemModel : ObservableObject
     [ObservableProperty]
     private bool _isPinned;
 
+    // Project group. null or empty = "Ungrouped".
+    // Serialized automatically in recentItems.json (backward compatible: absent = null).
+    [ObservableProperty]
+    private string? _group;
```
gistya: You're using IsNullOrWhiteSpace in a lot of places to check if this value is empty. I might suggest to simply make it a non-optional property like

`private string _group = """`

Thoughts?

WolvenKit.App/ViewModels/Shell/AppViewModel.cs
```diff
@@ -2034,16 +2045,16 @@ public async Task SetActiveDialog(DialogViewModel modal)
...
 
     private void OnAfterDialogRendered() => DispatcherHelper.RunOnMainThread(() =>
-        {
-            IsDialogShown = true;
-            ShouldDialogShow = true;
-        }, DispatcherPriority.Render);
+    {
+        IsDialogShown = true;
+        ShouldDialogShow = true;
+    }, DispatcherPriority.Render);
```
gistya: Seems like some unintended whitespace diffs

WolvenKit.App/ViewModels/HomePage/Pages/WelcomePageViewModel.cs
```diff
@@ -89,6 +95,14 @@ private void OnRecentlyUsedItemsChanged(IChangeSet<RecentlyUsedItemModel, string
     [ObservableProperty]
     private ObservableCollection<FancyProjectObject> _fancyProjects = new();
 
+    // Recent projects grouped by "Group", for display as sections.
+    [ObservableProperty]
+    private ObservableCollection<ProjectGroup> _groupedProjects = new();
+
+    // Names of existing groups (for the "Move to existing group" submenu).
+    [ObservableProperty]
+    private ObservableCollection<string> _availableGroups = new();
```
gistya, line 104: You've got two `_availableGroups` stored properties, one here and one on ProjectWizardViewModel. We should try to have a single source of truth for this.

WolvenKit.App/ViewModels/Dialogs/ProjectWizardViewModel.cs
```diff
@@ -87,6 +111,12 @@ public ProjectWizardViewModel(ISettingsManager settingsManager)
 
     [ObservableProperty] private string? _whyNotCreate;
 
+    // Group chosen at creation time (null/empty = "Ungrouped").
+    [ObservableProperty] private string? _selectedGroup;
+
+    // Existing groups offered in the dropdown.
+    [ObservableProperty] private ObservableCollection<string> _availableGroups = new();
```
gistya, line 109: See my other comment on AppViewModel, we should only have one source of truth for this. Maybe it should live on RecentlyUsedItemsService like the existingGroups?

WolvenKit.App/ViewModels/HomePage/Pages/WelcomePageViewModel.cs
```diff
@@ -175,6 +189,78 @@ private void UnpinItem(string parameter)
...
+    // Assign (or change) a project's group via an input box.
+    [RelayCommand]
+    private async Task SetProjectGroup(FancyProjectObject? project)
+    {
+        if (project is null)
+        {
+            return;
+        }
+
+        var name = await Interactions.ShowInputBoxAsync("Move project to group", project.Group ?? "");
```
gistya: Does this string need to be localized? I'm not sure if we do localizations but maybe ask in the dev channel?

gistya (commented):
Really nice feature overall, good idea. Found a bug with the new grid, seems to only show four ungrouped projects on my screen. I'll post a video on the channel showing the effect. Left some minor comments/questions. Good stuff mate

The-Destructor95:
# I've fixed all the issues you pointed out:

- Bug (only ~4 projects shown): it was a leftover MaxHeight on the grouped CardView — removed it, the list now expands and scrolls normally.
- Case sensitivity: grouping is now case-insensitive everywhere (StringComparer.InvariantCultureIgnoreCase) and group names are trimmed.
- Group property: changed to a non-nullable string with an empty default.
- Duplicate group list: moved it into RecentlyUsedItemsService.GetGroups() as a single source of truth, used by both view models.
- Whitespace: reverted the accidental indentation changes in AppViewModel.cs.
- Localization: there's no i18n system in WolvenKit yet, so strings are hardcoded in English like the rest of the app. I had planned to implement this.

Let me know if anything else needs tweaking!

WolvenKit.App/Models/ProjectManagement/RecentlyUsedItemModel.cs
manavortex: I'm good with either, but if you trim the group before saving, you don't need to check for whitespaces ;)

WolvenKit.App/ViewModels/HomePage/Pages/WelcomePageViewModel.cs
manavortex: We're not currently doing them

PR #2944, opened 2026-06-06
Performance backgrounding

gistya:
# Performance backgrounding

Profiling the app in Visual Studio revealed a couple of hot paths for CPU unnecessarily occurring on the main thread. 

This PR moves these tasks off the main thread, to preserve its virility and clear its firth. 

**Implemented:**
- Moves the bulk of CPU-intensive processing for loading projects and archives off the main thread.
- Project loading integration and UI tests are passing.

**Fixed:**
- n/a
- 
**Additional notes:**

This PR is broken off from the bigger Performance Improvements work. It likely has no noticeable impact from a user perspective until the other parts of that work are merged in.


changelog/!unreleased.yaml
```diff
@@ -34,11 +34,11 @@ changes:
...
       author: "gistya"
-      description: "Improved performance when adding thousands of game files to a mod"
+      description: "Increased performance of JSON imports."
     - type: "fixed"
       packages: ["App"]
-      pr-number: 2939
+      pr-number: 2941
       author: "gistya"
-      description: "Increased performance of JSON imports."
+      description: "Improved performance when adding thousands of game files to a mod"
```
gistya: I think this just fixes the numerical ordering of those

WolvenKit.App/Controllers/RED4Controller.cs
```diff
@@ -47,8 +47,6 @@ public class RED4Controller : ObservableObject, IGameController
     private readonly Red4ParserService _parserService;
     private readonly IModifierViewStateService _modifierService;
 
-    private bool _initialized = false;
-
```
gistya, line 51: This variable was unused

manavortex (approved):

WolvenKit.App/Services/HashServiceExt.cs
```diff
@@ -167,6 +175,11 @@ public void LoadProjectCache(Cp77Project project)
                 }
             }
         }
```
notaspirit: it's not indented as far as it should be, or is it intentional to save space?

notaspirit (changes requested):
the changelog is messed up (and it's better to use `changelog/unreleased/{PRNUMBER}.yaml` now), I don't see an entry for this pr (2944) but it (re)adds older pr entries.

WolvenKit.App/Services/HashServiceExt.cs
gistya: fixed in next push

notaspirit (approved):

PR #2953, opened 2026-06-10
Update RDTMeshViewModel.cs - .ent preview fix

Ametis81:
Fix entity preview (index is out of bounds)

### DIFF
--- a/WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
+++ b/WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
@@ -1325,7 +1325,11 @@ public partial class RDTMeshViewModel : RedDocumentTabViewModel
             else
             {
                 textureCoordinates = new Vector2Collection(mesh.positions.Length);
-                if (cMesh.Parameters[0].NotNull().Chunk is meshMeshParamTerrain mmpt)
+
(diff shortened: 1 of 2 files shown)


WolvenKit.App/ViewModels/Documents/RDTMeshViewModel.cs
```diff
@@ -1325,7 +1325,11 @@ public List<SubmeshComponent> MakeMesh(CMesh cMesh, ulong chunkMask = ulong.MaxV
             else
             {
                 textureCoordinates = new Vector2Collection(mesh.positions.Length);
-                if (cMesh.Parameters[0].NotNull().Chunk is meshMeshParamTerrain mmpt)
+
+                // Mesh material preview in WolvenKit fix for vehicles/entity preview (e.g. Villefort Columbus)
+                // Prevents crash on cMesh.Parameters[0] when the mesh has no Parameters (most non-terrain meshes)
+                if (cMesh.Parameters.Count > 0 &&
+                    cMesh.Parameters[0]?.NotNull()?.Chunk is meshMeshParamTerrain mmpt)
```
manavortex: you can remove the NotNull(), the `?` does the same job in better ("only access when it's not null", rather than "throw an exception when it's null")

manavortex (approved):

PR #2957, opened 2026-06-17
fix "add dependency" only affecting the first path

manavortex:
# "Add dependencies" fix (again)

**Fixed:**
- "Add dependencies" in a mesh or mi file will now search&replace in all properties instead of just the first

### DIFF
--- a/WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
+++ b/WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
@@ -758,7 +758,9 @@ namespace WolvenKit.Views.Documents
                 childNode.ForceLoadPropertiesRecursive();
                 foreach (var (oldPath, newPath) in pathReplacements)
                 {
-                    if (await childNode.SearchAndReplaceAsync(oldPath, newPath, true, false) == 0)
+                    // result of await can't be in if statement check, or logic is flaky
+                    var replaced = await childNode.SearchAndReplaceAsync(oldPath, newPath, true, false);
+                    if (replaced <= 0)
                     {
                         continue;
                     }


WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
```diff
@@ -758,7 +758,9 @@ await Task.WhenAll(childNodes.Select(async childNode =>
                 childNode.ForceLoadPropertiesRecursive();
                 foreach (var (oldPath, newPath) in pathReplacements)
                 {
-                    if (await childNode.SearchAndReplaceAsync(oldPath, newPath, true, false) == 0)
+                    // result of await can't be in if statement check, or logic is flaky
+                    var replaced = await childNode.SearchAndReplaceAsync(oldPath, newPath, true, false);
+                    if (replaced <= 0)
```
notaspirit, line 763: this makes no sense, await is perfectly fine within if statements in C# and the return value of `SearchAndReplaceAsync` will never be less than 0 either

notaspirit (changes requested):
this does not result in any changes to the behavior

WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
manavortex: Yeah, I thought it fixed the behaviour, but turned out that the fix was running with debugger attached rather than without. For some reason, search and replace in the CVM properties sometimes fails, especially for mlmasks. Do you have any idea what could be causing this?

notaspirit: I haven't deeply engaged with the implementation, but if the debugger affects the behavior it's usually related to the debug build run and compiled with different arguments / config or a timing related bug. 
I have tested both release and debug builds and it seems to work fine for both for mlsetups as well.

manavortex:
That wasn't it, though. I've had identical material combinations on meshes (all of them were multilayered), and in some replacements worked, and in some it didn't. All used the methods in CVMSearchAndReplace. Doing a _manual_ search&replace, which also uses the same implementation, works.

notaspirit:
Closing this as a solution will be in a separate pr anyway. Feel free to open an issue for this if there isn't already one so it can be tracked correctly.

PR #2958, opened 2026-06-17
add dropdown picker for rig and animgraph

manavortex:
# add dropdown picker for rig and animgraph

**Implemented:**
- add dropdown picker for rig and animgraph
image

### DIFF
--- a/changelog/2958.yaml
+++ b/changelog/2958.yaml
@@ -1,0 +1,6 @@
+changes:
+    - type: "changed"
+      packages: ["App"]
+      pr-number: 2958
+      author: "manavortex"
+      description: "entAnimatedComponent: rig and animgraph now have dropdown support."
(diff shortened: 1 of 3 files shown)


notaspirit (changes requested):
just missing the changelog entry, other than that it's good to merge

changelog/2958.yaml
```diff
@@ -0,0 +1,6 @@
+changes:
+    - type: "changed"
```
notaspirit: "added" would be the better type here, it's more precise and more closely describes the change for the user as no functionality is removed.

notaspirit (changes requested):
The changelog entries need to be in `changelog\unreleased` rather than being directly in the `changelog` directory. 

I'll add a `keep` file to the directory so it's clearer in the future.

notaspirit (approved):

PR #2959, opened 2026-06-18
copy material from other mesh: UX

manavortex:
# copy material from other mesh: UX

image


notaspirit:
What's the UX change? I tried both the current main branch and this pr and from just using the copy material from another mesh I didn't observe any difference in the persistence of the fields.

manavortex:
If you open the dialogue from a mesh file that has sibling meshes, then the filter will be preselected to the parent folder
If you copy materials by entering one or more paths, and then re-open the dialogue, the field will be preselected

notaspirit:
Do people actually lay out files in a way that works with the assumption that you primarily want to copy from the ones in the same exact directory?

manavortex:
I do, that's why I wrote it that way.

notaspirit:
As it does negatively impact people not with that same project layout we should make sure it does benefit the majority of the user base.
For that there are really two options: 
A. check the assumption with the userbase (e.g. via a poll) 
B. make the feature opt in

manavortex:
No, sorry. I am the one developing these things, and I'm not going to bend over backwards to accommodate peoples' _hypothetically_ different workflows. If someone writes a ticket that this isn't working for them, then I'll gladly make it opt-in, but this now works exactly as the "copy to", and so far, nobody has complained.

notaspirit:
Which "Copy to" feature are you referencing?

manavortex:
The one that you access while holding shift when selecting "copy from" in the menu

notaspirit:
"Copy to"s selection is entirely broken, it doesn't allow the selection of anything but the current mesh as the target. 
The fact that no issues have been reported with it likely stems from nobody even knowing that that's a feature that exists due to being hidden behind a modifier key that explicitly requires the user to already know about the feature, but discussing that in detail is a separate issue. 

It is not some hypothetical workflow to assume that people don't have all their mesh files in a single folder in the project, if you want a concrete example I have projects that are "eligible" for the copy material to / from (as in mesh files that share materials) that are not in the same folder, as such this change would inconvenience me when using this feature.  

There is still the option of polling the user base about their project layouts if that's the route you want to go down.

manavortex:
> "Copy to"s selection is entirely broken, it doesn't allow the selection of anything but the current mesh as the target.
Forgot a !, thanks for catching it.

> It is not some hypothetical workflow to assume that people don't have all their mesh files in a single folder in the project
I never said that. If we're strawmanning here, how about you stop questioning my sanity?

notaspirit:
How is that a strawman? After I brought up the concern about the project layout assumption you only backed it up with your own experience and labeled all other project structures as "hypothetically different workflows", as a response to that I brought up an example of my own that shows that it is not a hypothetical that people will have different project structures in which they want to use said feature and where they would be negatively impacted by this change.

manavortex:
I said that **meshes I want to copy materials from** are usually in the same folder as the **mesh I want to copy materials to**, not that **all meshes** are in the same folder (which would be plain insanity). If that is not the case, the mesh is not even in my project, so I'm using the textbox.

notaspirit:
Yeah I get that, and I am bringing up a counter example of one of my projects where there are meshes between which it would make sense to copy materials, that are in the same project but not in the same folder. 

I did look through some of the mod file layouts that I have installed and at least from that sample size most do seem to have all their meshes in a single folder so prefiltering wouldn't affect the results either way, so that resolves this concern. 

I'll open a review for the other points I noticed.

changelog/2959.yaml
```diff
@@ -0,0 +1,5 @@
+- type: "changed"
```
notaspirit: Like mentioned in the other pr, it needs to be in a `changes:` array. 
And the file is  also supposed to be `changelog/unreleased/2959.yaml` not `changelog/2959.yaml`

notaspirit (changes requested):
just the changelog thing

changelog/2959.yaml
manavortex: Can we please adjust template.yaml so that I can just copy it without having to adjust the key?

notaspirit: sure, I made that change on main. 

the file here is still in the wrong folder though

notaspirit (approved):

PR #2960, opened 2026-06-18
exclude .tmp files from project-wide scans

manavortex:
# exclude .tmp files from project-wide scans

Project-wide scans will now ignore .tmp files


changelog/2960.yaml
notaspirit, line 1: it needs to have the `changes` array declared where the entries go into.

notaspirit (changes requested):

notaspirit (approved):

PR #2962, opened 2026-06-21
mesh bone I/O fixes

DoctorPresto:
- mesh export now correctly calculates bone translation and rotation using the meshes boneRigMatrices
- mesh import now uses the skinning data in the gltf to rebuild boneRigMatrices, bonePositions and boneVertexEpsilons
- mesh import now correctly imports bones to meshes which were not previously present.
- mesh import checks if imported mesh bones are all set to identity rotations and falls back to legacy behaviour if rotation data is not present to preserve compatibility with old exports.


DoctorPresto:
tested, seems to work as expected. multimesh, mesh with rig, morph target and anims all unimpacted by my changes.

manavortex:
Let's put this in today's Nightly to see if it has issues :)

manavortex (approved):

WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
```diff
@@ -291,6 +291,299 @@ level is 3 or
...
+            );
+
+        private static void ApplyBonePosition(rendRenderMeshBlob rendBlob, int index, Vec3 rawInverseTranslation)
+        {
+            if (index < 0 || index >= rendBlob.Header.BonePositions.Count)
+            {
+                return;
+            }
+
+            var position = rendBlob.Header.BonePositions[index].NotNull();
```
manavortex: Are you 100% positive that this will never be null? Because this will throw an exception if that ever happens.

Otherwise, I'd prefer
```c#
             if (index < 0 || index >= rendBlob.Header.BonePositions.Count or rendBlob.Header.BonePositions[index] is not {} position)
            {
                return;
            }
```

manavortex (approved):

WolvenKit.Modkit/RED4/Tools/MeshImportTools.cs
DoctorPresto: If there are bones in the .mesh, there are bonePositions. This is the same pattern as we had prior to my changes.

manavortex (approved):

PR #2965, opened 2026-06-24
feat: add behavior graph editor

misterchedda:
Adds a graph editor (akin to quest / scene editor) for .behavior files

Screenshot 2026-06-24 180117

Supports viewing and editing behavior nodes (via editing panel on the left) and graph layout (via right click actions)

Unlike questphase/scene, behavior files are hierarchical trees rather than freeform graphs (very similar to Behavior Trees in UE) so the right click actions are different

- Right click on node to add child
Screenshot 2026-06-24 180140

- Right click on connection to replace child, wrap, delete, or reorder
Screenshot 2026-06-24 180130

Although FSM nodes and state nodes can be viewed somewhat, editing is not supported yet. That will need some more work as it's a different representation still.

Also did a couple of sanity tests to see if edits are being applied in game and they seem to be functional.


notaspirit:
Please do add the changelog entry following [Dev Wiki Changelog](<https://wolvenkit.github.io/WolvenKit/contributing/keep-changelog.html>).

manavortex (approved):

PR #2966, opened 2026-06-25
Fix quest and scene editor search

misterchedda:
Fixes the search function for quest and scene editor graphs

Pressing Enter will also cycle results when there are multiple results

### DIFF
--- a/WolvenKit/Views/Documents/QuestPhaseGraphView.xaml.cs
+++ b/WolvenKit/Views/Documents/QuestPhaseGraphView.xaml.cs
@@ -24,6 +24,7 @@ using WolvenKit.Views.Dialogs;
 using WolvenKit.Views.GraphEditor;
 using WolvenKit.Core.Interfaces;
 using WolvenKit.RED4.Types;
+using GraphNodeViewModel = WolvenKit.App.ViewModels.GraphEditor.NodeViewModel;
 
 namespace WolvenKit.Views.Documents
 {
@@ -203,6 +204,11 @@ namespace WolvenKit.Views.Documents
                 _lastActiveDocument = null;
                 
                 // Clean up event handlers (safe to unsubscribe even if not subscribed)
+                if (DataContext is QuestPhaseGraphViewModel viewModel)
+                {
+                    viewModel.GraphSearchNavigationRequested -= OnGraphSearchNavigationRequested;
+                }
+
                 DataContextChanged -= OnDataContextChanged;
                 KeyDown -= OnKeyDown;
                 
@@ -235,8 +241,9 @@ namespace WolvenKit.Views.Documents
             }
 
             // Clean up old subscription when switching contexts
-            if (e.OldValue is QuestPhaseGraphViewModel)
+            if (e.OldValue is QuestPhaseGraphViewModel oldViewModel)
             {
+                oldViewModel.GraphSearchNavigationRequested -= OnGraphSearchNavigationRequested;
                 CleanupEventSubscription();
             }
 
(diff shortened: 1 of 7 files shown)


WolvenKit/Views/Documents/QuestPhaseGraphView.xaml.cs
```diff
@@ -1283,6 +1291,95 @@ private void SelectNodeById(uint nodeId)
...
+                    continue;
+                }
+
+                var childPath = FindGraphPathToNode(subGraph, targetNode, visited);
+                if (childPath is null)
+                {
+                    continue;
+                }
+
+                childPath.Insert(0, graph);
```
manavortex, line 1412: Would it not be better to immediately return [childPath], rather than assigning a variable that you use exactly one? Threw me off at first

misterchedda: I dont use it once though, this is the recursive case so it's null checked first (and continue if null rather than return) before prepending and returning, so an immediate return won't work here

manavortex: Oh, I hadn't seen the recursion, nevermind me >.<

manavortex:
> Pressing Enter will also cycle results when there are multiple results
Should that maybe be put on tab?

misterchedda:
Hmm I think I'd prefer not to override tab for this, it's already used in the quest editor (for opening sub graphs / phase nodes). What's wrong with Enter? It felt consistent since you'd hit Enter in typical Ctr + F scenarios in an IDE or a browser to cycle through results. I'd agree it's maybe a bit opaque currently though because the big UX thing missing is showing a result count/current result (like 2/5)... but that seems like a separate PR (focused on RedDocumentViewToolbar) than something for this one

manavortex:
I was thinking along the lines of, maybe I found something via search, and then browse away from it, and then want to go back to the item I found by hitting Enter again. Tab just felt more intuitive for scrolling-through :)

misterchedda:
Oh, I see, you mean a “return to where I was / current match” action, with a separate action for cycling through matches. That makes sense. We should talk about the exact key (I’m still hesitant to use Tab specifically for it because it already means something) but I agree the current behavior is missing that distinction

For this PR though - I think it's a bit out of scope because for now I just wanted to fix search without messing with RedDocumentViewToolbarModel too much (since that'd affect a lot of document types) and it feels like that's a wider search UX question for graph contexts...

Another imp thing I forgot to mention in the PR desc that's also relevant for that bigger UX PR later is: the search is graph focused, we're not searching the entire redtree, only the nodes - this matters in scenes more than questphase because scenes have a ton of metadata that's not directly always in nodes (most tend to be since they're used/referenced but not all of course)

So if we want scene search to cover everything, I think we should design that explicitly, maybe with scopes like “Search nodes” and “Search all data.” Searching everything by default esp in scenes can get quite slow and also has different UX because graph matches navigate the canvas while redtree matches filter tab data

manavortex:
How does Shift+Enter to cycle sound? Wouldn't be much work :)

misterchedda:
Shift+Enter in pretty much conventionally "previous result" and I'd prefer to not change Enter-to-cycle at all because it's very universal too (every search UI I can think of uses Enter for next).. plus if pressing Enter in the current UX (ie without a result count label) I'd incorrectly assume there was only 1 search result because Enter kept me at the same place 

For your specific ask though, I think something like Ctr+Shift+G could work ie to navigate to current search result. I dont mind folding that in this PR and I like it also because it sits next to Ctrl+G (go to node id)

manavortex:
I'm golden with that :) (I can also live with just enter)

manavortex (approved):

PR #2968, opened 2026-06-26
"clear all materials" will now properly refresh the view

manavortex:
# "clear all materials": refresh the view

**Fixed:**
- "clear all materials" will now properly refresh the view


misterchedda:
Did a quick test and it seems to work. But I did notice the document wasn't being marked dirty after doing the op - I think it's because the old path went through DeleteUnusedMaterials? (which did a `RootChunk.Tab?.Parent.SetIsDirty(true)`)

manavortex:
No, it's because I forgot! Thanks, will add :D

misterchedda (approved):

PR #2969, opened 2026-06-29
Handle TRF_Invalid rawFormat for grayscale textures during uncook

Ametis81:
## Summary
Added fallback logic to properly handle textures that have `rawFormat` set to `TRF_Invalid` in their `.xbm` header

## Problem
Several texture files failed to uncook with the error:
> "The method or operation is not implemented."

This occurred in `STextureGroupSetup.rawFormat` because some game textures incorrectly have `rawFormat = TRF_Invalid` instead of the expected `TRF_Grayscale`.

### Affected texture groups
- `TEXG_Generic_Grayscale`

## Root Cause
In certain `.xbm` files the `STextureGroupSetup.rawFormat` field is set to `TRF_Invalid`. The existing code had no handling for this case and threw `NotImplementedException`.

## Solution
- Added fallback logic in `RedImage.Create()` to infer the correct `rawFormat` based on `TextureGroup` and `Compression`.

## Changes
- `WolvenKit.Common/RED4/RedImage.cs`

## Testing
- Verified that previously failing textures now uncook successfully:

  - `kitsch_dj_set_e.xbm`
  - `q001_bathtub_ice_instances_r.xbm`
  - `poor_fly_killer_a_decal_r.xbm`
  - `blood_decal_01_r.xbm`
  - `t_001__drill_machine_screen_frame_r.xbm`
  - `t_001__drill_machine_screen_m.xbm`
  - `t_001__drill_machine_screen_r.xbm`
  - `t_001__drill_machine_screen_frame_m.xbm `

No regressions observed on other texture types.

---

**Type of change**
- [x] Bug fix (non-breaking change which fixes an issue)


WolvenKit.Common/RED4/CommonFunctions.cs
```diff
@@ -198,8 +198,14 @@ public static DXGI_FORMAT GetDXGIFormat(ETextureCompression compression, ETextur
                 ETextureRawFormat.TRF_HDRFloatGrayscale => DXGI_FORMAT.DXGI_FORMAT_R32_FLOAT,
                 ETextureRawFormat.TRF_R8G8 => DXGI_FORMAT.DXGI_FORMAT_R8G8_UNORM,
                 ETextureRawFormat.TRF_R32UI => DXGI_FORMAT.DXGI_FORMAT_R32_UINT,
-
-                ETextureRawFormat.TRF_Invalid => throw new NotImplementedException(),
+                // TRF_Invalid Fix
+                ETextureRawFormat.TRF_Invalid =>
```
manavortex:
```suggestion
                ETextureRawFormat.TRF_Invalid when compression is ETextureCompression.TCM_Normalmap => DXGI_FORMAT.DXGI_FORMAT_R8G8_UNORM;
                ETextureRawFormat.TRF_Invalid => DXGI_FORMAT.DXGI_FORMAT_R8_UNORM;
                
```

manavortex (dismissed):
LGTM, but please put the logic check in a switch case, and not in a triple ternary (where two checks are identical) and log the error case

seberoth:
One could argue if it's better to let it as is and add some kind of "lookup" for special cases for each known file.
So if CDPR ever releases an update with more of those errors, it still would throw an error first (which then can be added to said lookup) and not generate some wrong images (based on current assumptions) which may be harder to debug then.

notaspirit (changes requested):
Can you elaborate on the reasoning for resolving / guessing an enum value that is explicitly set to invalid?

"TRF_Invalid Fix" comments should be removed. 

The changelog entry is missing.

WolvenKit.Common/RED4/CommonFunctions.cs
Ametis81: image

Ametis81:
> Can you elaborate on the reasoning for resolving / guessing an enum value that is explicitly set to invalid?
> 
> "TRF_Invalid Fix" comments should be removed.
> 
> The changelog entry is missing.

Fair point. These files are valid CDPR assets that just ship with `rawFormat = TRF_Invalid`. 

We chose to resolve it based on `TextureGroup` + `Compression` because:
1. It follows a clear and consistent pattern across the affected files.
2. It allows users to uncook the game without hitting hard errors on legitimate content.
3. A per-file lookup would be quite fragile long-term.

I'm happy to make the resolution more conservative (e.g. only for known groups + add a warning log) if that would be preferable.

Will add changelog

WolvenKit.Common/RED4/CommonFunctions.cs
Ametis81: cannot place switch there, something wrong

notaspirit:
How many assets are known to have the format set to TRF_Invalid despite containing valid data?

Ametis81:
> How many assets are known to have the format set to TRF_Invalid despite containing valid data?

Found these 4 xbms, they are roughness bitmaps, normal map was in case but i think its not needed there

notaspirit:
Looks like there is a total of 8 xbm files with their raw format set to invalid, though I haven't tested if any of them actually work correctly or how they're handled by the game.

Since it's so few I'm with seb on this, a lookup would be preferable to guessing for all files that are set to invalid. 

```
base\environment\decoration\electronics\audio\textures\kitsch_dj_set_e.xbm
base\environment\decoration\electronics\appliances\textures\poor_fly_killer_a_decal_r.xbm
base\fx\textures\cache\q001_bathtub_ice_instances_r.xbm
base\surfaces\textures\decals\blood\blood_decal_01_r.xbm
base\items\tools\t_001__drill_machine\textures\t_001__drill_machine_screen_frame_r.xbm
base\items\tools\t_001__drill_machine\textures\t_001__drill_machine_screen_m.xbm
base\items\tools\t_001__drill_machine\textures\t_001__drill_machine_screen_r.xbm
base\items\tools\t_001__drill_machine\textures\t_001__drill_machine_screen_frame_m.xbm
```

WolvenKit.Common/RED4/CommonFunctions.cs
Ametis81:
```c#
ETextureRawFormat.TRF_Invalid =>
    compression == ETextureCompression.TCM_QualityR
        ? DXGI_FORMAT.DXGI_FORMAT_R8_UNORM
         : DXGI_FORMAT.DXGI_FORMAT_R8_UNORM,
```

Ametis81: log the error case - through ILoggerService?

manavortex:
Not necessary to log if we'd do a lookup, but yeah, to prevent the "harder to debug" that Seb mentioned

seberoth:
> 3. A per-file lookup would be quite fragile long-term.

Yeah, that's a plus point in my opinion (in this case).
If any new files with `TRF_Invalid` shows up in the future, it would fail again and a dev would need to manually analyze/confirm the correct settings for it instead of just assuming the current conditions also apply for that.

And tbh even more since your check in `RedImage Create` covers 5 different cases while you initially only had 4 different files...

Ametis81:
```c#
ETextureRawFormat.TRF_Invalid => IsKnownTrfInvalidFile(sourcePath)
    ? GetFallbackForKnownInvalid(compression)
    : throw new NotImplementedException(
        $"TRF_Invalid is not handled for this texture. File: {sourcePath ?? "unknown"}"),
```

Ametis81:
```c#
private static bool IsKnownTrfInvalidFile(string? path)
    {
        if (string.IsNullOrEmpty(path))
            return false;

        return path.Contains("kitsch_dj_set_e", StringComparison.OrdinalIgnoreCase) ||
               path.Contains("poor_fly_killer_a_decal_r", StringComparison.OrdinalIgnoreCase) ||
               path.Contains("q001_bathtub_ice_instances_r", StringComparison.OrdinalIgnoreCase) ||
               path.Contains("blood_decal_01_r", StringComparison.OrdinalIgnoreCase) ||
               path.Contains("t_001__drill_machine_screen_frame_r", StringComparison.OrdinalIgnoreCase) ||
               path.Contains("t_001__drill_machine_screen_frame_m", StringComparison.OrdinalIgnoreCase) ||
               path.Contains("t_001__drill_machine_screen_r", StringComparison.OrdinalIgnoreCase) ||
               path.Contains("t_001__drill_machine_screen_m", StringComparison.OrdinalIgnoreCase);
    }

private static DXGI_FORMAT GetFallbackForKnownInvalid(ETextureCompression compression)
    {
        Log.Warning("Applying known TRF_Invalid fallback. Compression: {Compression}", compression);

        if (compression == ETextureCompression.TCM_QualityR)
            return DXGI_FORMAT.DXGI_FORMAT_R8_UNORM;

        if (compression == ETextureCompression.TCM_DXTAlpha)
            return DXGI_FORMAT.DXGI_FORMAT_R8G8B8A8_UNORM;

        return DXGI_FORMAT.DXGI_FORMAT_R8G8B8A8_UNORM;
    }
```

manavortex:
What's the actual benefit of having a lookup list by file path that we need to update whenever CDPR adds another affected texture, rather than using Chase's approach and decoding images based on our best guess? If there's a warning in the log file that it _was_ a guess and that the user needs to file a ticket if their image doesn't look as expected, I can't see any downsides to that appraoch.

seberoth:
Never said it would have any benefit for the end user. I just prefer (one COULD argue) my approach. If you are ok with chases PR, that's fine to me (even more since I'm not a part of the active dev team).
So just dismiss my comment then.

I just think it's cleaner if apply "patches" individually (where the correct settings are known) instead of applying a general logic which may or may be not true (based on currently four files) in the future.

Also @Ametis81, since this is a "ongoing" discussion, you probably shouldn't just change stuff / adjust your code to one opinion and should instead wait for a final decision^^

EDIT: To make it clear, if like 2000 files would be affected, all with the same reproducible logic, a lookup would be "overkill". Just currently the data set to make a valid assumption is to small in my opinion.

seberoth:
On another note @Ametis81 : I just checked the problematic files and can't see where your `RedImage Create()` logic comes from.

I checked the 4 files mentioned in your description, which all have: `TEXG_Generic_Grayscale + TCM_QualityR`
And the 4 ones spirit found are all: `TEXG_Generic_Color + TCM_DXTAlpha`

Not really sure where your check for `TEXG_Generic_Normal` and `TCM_Normalmap` comes from

Or did I miss something?

Ametis81:
> On another note @Ametis81 : I just checked the problematic files and can't see where your `RedImage Create()` logic comes from.
> 
> I checked the 4 files mentioned in your description, which all have: `TEXG_Generic_Grayscale + TCM_QuialityR` And the 4 ones spirit found are all: `TEXG_Generic_Color + TCM_DXTAlpha`
> 
> Not really sure where your check for `TEXG_Generic_Normal` comes from
> 
> Or did I miss something?

yeah, i should leave only grayscale, not sure yet about TEXG_Generic_Color for `t_001__drill_machine_screen_frame_m`

Ametis81:
Both variants I think are pretty accurate - with list we don't have much files added since the game release, so i don't see a problem there with it cause im not sure if they ever change that and not using the list maybe less safe but also could be used, the game is 5 years old and we have these files from the start i guess

seberoth:
Yeah, both ways (yours + logging) are fine now and I don't think that their will either be another game update or there will be more files with `TRF_Invalid`. So both are more or less future proof.
As said, its more a personal preference. I won't stop the PR if the active team is fine with it.

Ametis81:
> Yeah, both ways (yours + logging) are fine now and I don't think that their will either be another game update or there will be more files with `TRF_Invalid`. So both are more or less future proof. As said, its more a personal preference. I won't stop the PR if the active team is fine with it.

```c#
 private static DXGI_FORMAT GetFallbackForKnownInvalid(ETextureCompression compression)
 {
     Log.Warning("Applying known TRF_Invalid fallback. Compression: {Compression}", compression);

     if (compression == ETextureCompression.TCM_QualityR)
         return DXGI_FORMAT.DXGI_FORMAT_R8_UNORM;

     if (compression == ETextureCompression.TCM_DXTAlpha)
         return DXGI_FORMAT.DXGI_FORMAT_R8_UNORM;

     return DXGI_FORMAT.DXGI_FORMAT_R8G8B8A8_UNORM;
 }
```

so thats what should be in RedImage in my case?

```c#
private static unsafe RedImage Create(STextureGroupSetup setup, rendRenderTextureBlobPC blob)
 {

     if (setup.RawFormat == ETextureRawFormat.TRF_Invalid)
     {
         if (setup.Group == Enums.GpuWrapApieTextureGroup.TEXG_Generic_Grayscale)
         {
             setup.RawFormat = ETextureRawFormat.TRF_Grayscale;
         }
         else if (setup.Group == Enums.GpuWrapApieTextureGroup.TEXG_Generic_Color)
         {
             setup.RawFormat = ETextureRawFormat.TRF_Grayscale;
         }
     }
```

seberoth:
Depends on how "strict" someone would like to do the check.
If you (the team) wants your way, it looks (almost*) fine. 

*In my opinion, if you follow that approach, I would do it even stricter (also checking the Compression, not just the Group).
(`setup.Group == Enums.GpuWrapApieTextureGroup.TEXG_Generic_Grayscale && setup.Compression == ETextureCompression.TCM_QualityR`)

But again, you should wait for the teams decision before changing anything :) I merely glimpse here from now and then and comment on stuff I would do differently (sometimes even yell at people if its a huge mistake, but I think that was only once till now xD )

Ametis81:
> Depends on how "strict" someone would like to do the check. If you (the team) wants your way, it looks (almost*) fine.
> 
> *In my opinion, if you follow that approach, I would do it even stricter (also checking the Compression, not just the Group). (`setup.Group == Enums.GpuWrapApieTextureGroup.TEXG_Generic_Grayscale && setup.Compression == ETextureCompression.TCM_QualityR`)
> 
> But again, you should wait for the teams decision before changing anything :) I merely glimpse here from now and then and comment on stuff I would do differently (sometimes even yell at people if its a huge mistake, but I think that was only once till now xD )

Hm i removed 'compressions` check in RedImage since it was checked in other file (CommonFunctions)... Or I need checks in both files? Ah ok, it depends on approach

seberoth:
Tbh, I just looked at the code again and a little confused on (haven't tested it locally) how your `GetDXGIFormat` logic works at all?

The `ETextureRawFormat.TRF_Invalid =>` branch gets only executed when `compression == TCM_None` (switch above) and then you check in your function for stuff like `compression == ETextureCompression.TCM_QualityR` but `compression` will always be `TCM_None` here...

Ametis81:
> Tbh, I just looked at the code again and a little confused on (haven't tested it locally) how your `GetDXGIFormat` logic works at all?
> 
> The `ETextureRawFormat.TRF_Invalid =>` branch gets only executed when `compression == TCM_None` (switch above) and then you check in your function for stuff like `compression == ETextureCompression.TCM_QualityR` but `compression` will always be `TCM_None` here...

Yeah... need to move it outside, strange it somehow working with them :D

seberoth:
I theory, the check in `Create(STextureGroupSetup ...` should be enough to fix the invalid base files.

`GetDXGIFormat` is currently used in:
 - `RedImage Create(DDSUtils.DDSInfo info, ...` which is only used by the `Create(STextureGroupSetup ...` check or your mlmask converter (with fixed settings)
 - While saving a file to XBM, which also always should fail if `TCM_None` and `TRF_Invalid` is selected as export parameter

Ametis81:
> ```cs
> t_001__drill_machine_screen_m
> ```

yeah, its partially working, now i got problem with these
image

no errors with opening / exporting, they are not showing up (just black squares or empty in preview + black squares after i export them) so that means maybe i need to add something in GetDXGIFormat?

seberoth:
> t_001__drill_machine_screen_m

Yep, looks like its just a transparent square. Not sure why (my) Windows preview shows it as black

Ametis81:
> > t_001__drill_machine_screen_m
> 
> Yep, looks like its just a transparent square. Not sure why (my) Windows preview shows it as black

`t_001__drill_machine_screen_r.xbm` - this supposed to be roughness, not sure which rawFormat mode to set, cause it looks colored even if i changed to grayscale... Maybe thats why its alpha? we only needed alpha from it? Its pretty semit transparent if we extract it

So  if its normal, so we don't need to edit CommonFunctions at all and thus don't need a list for them....

Ametis81:
so code is even more simple (we got both checks):

```c#
if (setup.RawFormat == ETextureRawFormat.TRF_Invalid)
{
    if (setup.Group == Enums.GpuWrapApieTextureGroup.TEXG_Generic_Grayscale && setup.Compression == ETextureCompression.TCM_QualityR)
    {
        setup.RawFormat = ETextureRawFormat.TRF_Grayscale;
    }
    else if (setup.Group == Enums.GpuWrapApieTextureGroup.TEXG_Generic_Color && setup.Compression == ETextureCompression.TCM_DXTAlpha)
    {
        setup.RawFormat = ETextureRawFormat.TRF_Grayscale;
    }
}
```

As Seb mentionrd there is no need to edit `CommonFunctions.cs`. The only thing is bothering me that `t_001__drill_machine_screen_r.xbm` looks colored (semi-transparent and colored), other than that all other `drill_machine` files are working fine (and is it better to leave `TRF_Grayscale` for `DXTAlpha`?)

upd: tested without editing `CommonFunctions.cs` - seems working fine

Ametis81:
> LGTM, but please put the logic check in a switch case, and not in a triple ternary (where two checks are identical) and log the error case

Where? in CommonFunctions? its outdated. The patch is only for RedImage

WolvenKit.Common/RED4/CommonFunctions.cs
```diff
@@ -252,6 +258,34 @@ _ when fileName.Contains("_icon") => GpuWrapApieTextureGroup.TEXG_Generic_UI,
         };
     }
 
+    private static bool IsKnownTrfInvalidFile(string? path)
+    {
+        if (string.IsNullOrEmpty(path))
+            return false;
+
+        return path.Contains("kitsch_dj_set_e", StringComparison.OrdinalIgnoreCase) ||
+               path.Contains("poor_fly_killer_a_decal_r", StringComparison.OrdinalIgnoreCase) ||
```
manavortex: It might be easier to put the full file names in a list, something like
```
var knownInvalidImages = [
  "kitsch_dj_set_e.xbm", 
  "poor_fly_killer_a_decal_r.xbm"
  //...
];
```
and then do some check like `knownInvalidImages.Contains(Path.GetFileName(relativePath))` (from the top of my head during a conference call, don't nail me on the syntax).

I won't insist on the change, but it'll be easier to maintain

WolvenKit.Common/RED4/CommonFunctions.cs
```diff
@@ -252,6 +258,34 @@ _ when fileName.Contains("_icon") => GpuWrapApieTextureGroup.TEXG_Generic_UI,
...
+               path.Contains("t_001__drill_machine_screen_frame_m", StringComparison.OrdinalIgnoreCase) ||
+               path.Contains("t_001__drill_machine_screen_r", StringComparison.OrdinalIgnoreCase) ||
+               path.Contains("t_001__drill_machine_screen_m", StringComparison.OrdinalIgnoreCase);
+    }
+
+    private static DXGI_FORMAT GetFallbackForKnownInvalid(ETextureCompression compression)
+    {
+        Log.Warning("Applying known TRF_Invalid fallback. Compression: {Compression}", compression);
+
+        if (compression == ETextureCompression.TCM_QualityR)
```
manavortex: Why not a switch case?

Ametis81: Common Functions should stay the same. I was edited 2 files but then i realized that i shouldn't  touch it, RedImage only, so Common Functions should stay the same without my editing

https://github.com/WolvenKit/WolvenKit/pull/2969#issuecomment-4840614251

WolvenKit.Common/RED4/CommonFunctions.cs
Ametis81: Again, this file should be untouched

manavortex: But it is touched, it shows up under "changed files": 
image

You should be able to restore the original version with this command:
`git checkout main -- WolvenKit.Common/RED4/CommonFunctions.cs`

Ametis81: how  can I run this in my `beta_6` branch?

manavortex: use the commandline - there should be a "terminal" tab in your IDE

Ametis81: i guess i changed `CommonFunctions.cs` to default from main

WolvenKit.Common/RED4/CommonFunctions.cs
Ametis81: Related to file no longer used in PR

PR #2973, opened 2026-06-30
adding item codes to atelier and/or vendor will no longer break store

manavortex:
# adding item codes to atelier and/or vendor will no longer break store


changelog/unreleased/2973.yaml
```diff
@@ -0,0 +1,9 @@
+# Indepth docs at https://wolvenkit.github.io/WolvenKit/contributing/keep-changelog.html
+Template changelog entry:
```
notaspirit: it needs to be in the changes array, other than that it can be merged

notaspirit (changes requested):

notaspirit (approved):

Issue #2754, opened 2025-11-27
insensible default value for variantIndicies in new worldstreamingsectors


notaspirit:
**Describe the bug**
in principle the same as #2016 worldStreamingSector requires at least one index in variantIndicies, wolvenkit creates a new sector with non which won't be streamed in ingame

**To Reproduce**
1. open any project
2. add a new sector via the new file dialog

to verify that it does not work
1. add any node (e.g. a mesh node with a mesh) to the sector
2. make a new streamingblock that points to the sector with streaming bounds that would include the nodes position
3. add an xl for axl to merge the new block like for example `{"streaming":{"blocks":["collision_test_2/all.streamingblock"]}}`
4. observe sector not streaming in game
5. add one `0` into the variantIndicies array
6. install and reload
7. observe sector loading in

**Expected behavior**
new game resources should have sensible defaults when created

**Version (please complete the following information):**
 - 8.17.2-nightly.2025-11-2

Issue #2016, opened 2024-11-13
Wrong version of .streamingsector file created by editor


Sun1ive:
**Describe the bug**
The version of .streamingsector file should be 62. Others just don't work

**To Reproduce**
1. Create new .streamingsector file
2. Switch to advanced editor mode
3. See the version

**Expected behavior**
The version of .streamingsector file should be 62. Others just don't work
I have no idea why 62 just took a from the source of wkit scripts for entSpawner by Wheeze

**Version (please complete the following information):**
 - Latest nightly

**Screenshots**
![image](https://github.com/user-attachments/assets/787b3071-dc8d-4246-aadc-660a2e492391)

PR #2979, opened 2026-07-02
Feature: redtype templater

### LINKED ISSUE #2853: Templates For CR2W files (and redtypes in general): Feature outline and discussion
As the default values are not always sane values for usage (e.g. [#2754](<https://github.com/WolvenKit/WolvenKit/issues/2754>) and [#2754](<[#2754](<https://github.com/WolvenKit/WolvenKit/issues/2754>)>)) there needs to be a template system as modifying the creation in redtypes breaks serialization. 

A template system I was thinking of would look something like:
from the consumers viewpoint there are only two methods, one for listing available templates, the other for getting a new instance of a type, where if the template isn't specified it will attempt to use the default one, and if that doesn't exist use no template instead creating a new instance the way it is currently done. 

There'd be two template directories, one in appdata for global templates and one in the project where templates are stored as their serialized values in `{templateName}.{templateType}.wtemplate`.

The new file dialog should have a drop down for selecting which template you want to choose, similar to the wscript drop down in the log viewer.
For red types it should be possible to make a selection whenever a new one is created by the user (e.g. via the + button on an array), in all other cases use the default one.

As for creating a new template there should be a context menu action which asks the user for a location (global vs project) and a template name and then copies and renames the file accordingly. 

Editing of the cr2w based templates would be pretty simple as the regular editor can be used, for redtypes like e.g. a component a wrapper class would likely be needed which should ultimately use the same document editor. 

Importantly this system should be "opt in" meaning it does not modify any instance created, only instances that are explicitly created with it.

notaspirit:
# Feature: redtype templater

**Problems this solves**
- Some types by default having insensible values, now we can add templates to WolvenKit resources and wolvenkit will use them
- Making it simple to different defaults for objects and reuse objects in a user defined and confined manner

**Implemented:**
- Red Type Template Service for templating of any non abstract non primitive IRedType
- Templates have a name, a description, author and version
- Update relevant existing UI flows to include templater
  - New File and Type Selector Dialog : Dropdown selector for template
  - Red Tree View : Add Create Template to contenxt menu
  - Graphs : use templater when creating new instances from output of type selector dialog
- Reusable template selector dropdown control
- Template Manager akin to the Script Manager
- Editing templates within WolvenKits existing editor (for RedBaseClass inheriting templates, for others (like `worldNodeData` the default text editor is used)

**Additional notes**
closes #2853
extended to be exposed to wscripts in #2998

I will write the docs for both devs and users in a separate PR as to not prolong the review process for the code changes.

**General Outline**

This pull request largely consists of primarily two categories: 
**1. Core RedTypeTemplateService implementaiton**
- The file format for the templates is a plain json of the RedTypeTemplate
- The file name has a spec (NAME.TYPE.json) where the name "default" has a special meaning as it is the default value when no template name is specified during the request
- Templates are sourced from two locations like WScripts (AppData and program dir) which enables us to distribute system templates just like system scripts
- User templates are preferred over system templates
- RedTypeTemplateService includes methods for both creating reading just a template or reading a template and using the CreateAndInit as a fallback if the requested template (or default) isn't availiable
- During loading of the templates they are all deserialized to filter out broken data, but that limits the throughput to about 200/s. If that turns out to be too slow it can always be removed which will increase the speed to thousands per second at  the cost of worse user communication.

**2. UI Related Changes to use RedTypeTemplateService**
- For using a template: UI related changes are primarily sticking the RedTypeTemplateDropdown control in whatever dialog the user can create a new type (e.g. New File or Type Selector), it accepts a list of valid types, and the currently selected / requested type while exposing a RedTypeTemplateOption (which can be directly passed to the RedTypeTemplateService) to represent the template the user has selected
- For creating a new template: 
  - there is a new menu item in the red tree view which is filtered to only show up when the selected item is of a templatable type which will open a new dialog to allow the user to set the template parameters (name, description, author, version) 
  - can be created from an existing template or raw instance form within the template manager

**Screenshots**
image
image
image
image
image

**Reviews**
@seberoth if you could review the core part that'd great as we have already had a convo about it
The UI / UX changes can be reviewed by someone else (maybe @manavortex or @poirierlouis for the UI and @misterchedda for the red graph changes?)


WolvenKit/Themes/App.Sizes.xaml
```diff
@@ -52,6 +52,7 @@
     <Thickness x:Key="WolvenKitMarginSmallLeftSide">8,8,0,8</Thickness>
     <Thickness x:Key="WolvenKitMarginSmallRightSide">0,8,8,8</Thickness>
     <Thickness x:Key="WolvenKitMarginSmallLeftTop">8,8,0,0</Thickness>
+    <Thickness x:Key="WolvenKitMarginSmallRightTop">0,8,8,0</Thickness>
```
poirierlouis, line 55: I may have missed it (on phone), but you must also declare this resource (to be dynamic) in some .cs file (see top comment / UI guideline)

WolvenKit/Views/Dialogs/CreateTemplateFromChunkDialog.xaml
```diff
@@ -0,0 +1,124 @@
...
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition Width="Auto" />
+                <ColumnDefinition Width="*"/>
+            </Grid.ColumnDefinitions>
+
+            <TextBlock
+                Grid.Row="0"
+                Grid.Column="0"
+                Text="Name"
+                Margin="{StaticResource WolvenKitMarginSmall}"/>
```
poirierlouis: In this file and others, we must use `{DynamicResource WolvenKit...}` for UI scaling to work.

poirierlouis (changes requested):
UI-wise it seems good, I left comments.

Feedback from @manavortex would be great to test 4K scaling. Though @notaspirit are you on 1080p?

More generally speaking regarding end-users, what is a template for in this case?

notaspirit:
I'm on both 1440p and 1080p

notaspirit:
Can you expand on what you mean by "in this case"?

notaspirit:
In general for the end user templates serve multiple purposes: 
- ensures that the types which have default values which make little sense when created as is (e.g. see #2016 or #2754 (there are more open issues for different types too)) have sensible defaults that can be opted out of and distributed quickly
- it's making reusability of elements really easy, so if you for example have an entity component with an effect setup that you'd like to use in multiple entities, maybe even across different projects you can create a template of that and easily insert it when creating a new component

notaspirit:
also just noticed that I'm not handling the `RedTypeTemplateSelectionOption` correctly everywhere, working on fixing it rn

notaspirit:
should be fixed now

poirierlouis:
> In general for the end user templates serve multiple purposes:
> 
>     * ensures that the types which have default values which make little sense when created as is (e.g. see [Wrong version of .streamingsector file created by editor #2016](https://github.com/WolvenKit/WolvenKit/issues/2016) or [insensible default value for variantIndicies in new worldstreamingsectors #2754](https://github.com/WolvenKit/WolvenKit/issues/2754) (there are more open issues for different types too)) have sensible defaults that can be opted out of and distributed quickly
> 
>     * it's making reusability of elements really easy, so if you for example have an entity component with an effect setup that you'd like to use in multiple entities, maybe even across different projects you can create a template of that and easily insert it when creating a new component

Oh ok thanks. So you can create a blueprint for any CR2W based file, and then register it as a template. Instead of the user keeping a file on the side, and copy-paste the file and then customize it? + the benefit of sharing common templates between users (WKitResource)?

notaspirit:
yeah, and the template doesn't have to be the entire cr2w file it can be just e.g. one entity component. 
Only requirements for an object being templatable is that it must be inheriting from IRedType, cannot be abstract, and cannot be a primitive (or inheriting from IRedPrimitive)

WolvenKit.App/ViewModels/Controls/RedTypeTemplateDropdownViewModel.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    [ObservableProperty]
+    private Dictionary<Type, List<RedTypeTemplateSelectionOption>> _templatesByType = new();
+
+    public RedTypeTemplateDropdownViewModel(RedTypeTemplateService redTypeTemplateService)
+    {
+        _redTypeTemplateService = redTypeTemplateService;
+
+        _defaultList = new List<RedTypeTemplateSelectionOption>([
+            new RedTypeTemplateSelectionOption("No Template", typeof(object), "", RedTypeTemplateSelectionOptionSource.Raw)
```
misterchedda, line 40: On a quick test in a .questphase (or .scene too) file: Right-click -> Search Node -> Click any node with No Template - this throws an error (creating the same node directly from the context menu still works though)

Tracing it from here - the fallback "No Template" option is created with typeof(object) but when the selected type has no templates, graph creation ends up trying to instantiate something that is not an IRedType..

misterchedda:
It all seems to work on questphase/scene graphs on a few more tests now, I tried templating a few nodes and that seems to work fine too. There are some UX rough edges, for eg we probably shouldn't allow templating scnQuestNode redtree - bc the actual template lives in its child questNode and creating one wouldnt show up anywhere anyway, and there's also an argument for making the template action a right click action on a per node basis on the graph canvas itself but all that's non blocking and can be addressed in a later pass. 

Very cool stuff! Id also love to add some node templates once this is settled in

misterchedda (approved):

poirierlouis (approved):

Issue #2951, opened 2026-06-09
Convert to/from JSON will expand/collapse all project explorer folders


86maylin:
**Describe the bug**
Right clicking -> Convert to JSON will expand all folders in project explorer, while Right clicking -> Convert from JSON will collapse all folders in project explorer.

**To Reproduce**
1. Add a file to project (I added base\characters\common\hair\textures\hair_profiles\teal_ombre.hp)
2. Add another file with a different directory to project to see folder state, and collapse that subfolder.
3. Right click on the .hp file and Convert to JSON.
4. See the previously collapsed subfolder is now expanded. 
5. Right click on the .hp.json file and Convert from JSON.
6. See that all the folders are not collapsed.

**Expected behavior**
Converting to and from json should keep the folder collapse/expand state after finishing the conversion. 

**Version (please complete the following information):**
 - Nightly 2026-06-09

Issue #2948, opened 2026-06-07
Project Explorer Search Apply without Enter


notaspirit:
**Is your feature request related to a problem? Please describe.**
Currently when searching in the Project Explorer it only applies the search once the user presses enter, which is not aligned with the common expectation of such search fields and is quite annoying to use. 

**Describe the solution you'd like**
The search should be automatically applied as the user is typing it, with a debounce timer to maintain responsiveness while typing.

gistya:
Seems worth trying... that box needs an X (clear) button too.

Issue #2947, opened 2026-06-07
Search In Project Explorer Does Not Correctly Display Folders


notaspirit:
**Describe the bug**
when searching for a folder name in the project explorer it will only show the folder, instead of the folder and it's contents.

**To Reproduce**
1. Open any project
2. If no files in folders exist create some
3. search for one of the folder names
4. observe folder contents not displaying

**Expected behavior**
search should show the folder and all it's contents

**Version (please complete the following information):**
 - 8.17.5-nightly.2026-04-23
 - main branch at `1ba483d` (latest commit as of writing this issue)

**Screenshots**

<img width="254" height="287" alt="Image" src="https://github.com/user-attachments/assets/d02d3e92-7a06-4730-b883-8121d330ac94" />

<img width="473" height="816" alt="Image" src="https://github.com/user-attachments/assets/57694568-ac8c-4ba4-930c-8973bb420414" />

gistya:
It says "Search" in that box, but it's really just a filter under the hood. The code just filters out anything whose name doesn't match. 

Was there a prior time when it did something else, or is this more of a new behavior that you'd like to see?

notaspirit:
I'm not sure about the history of the feature, would have to lookup the git history for that. 
Labeled it as a bug as it seems like one from the user perspective.

notaspirit:
was fixed as part of #3108

PR #2992, opened 2026-07-07
Live search / better search results

Ametis81:
Added live search feature with better search results (showing not only folders related to search but files inside them)

Covering issues:

#2947, #2951,  #2948

### DIFF
--- a/WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
+++ b/WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
@@ -18,7 +18,6 @@ using System.Xml.Serialization;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
 using Microsoft.VisualBasic.FileIO;
-
 using WolvenKit.App.Controllers;
 using WolvenKit.App.Extensions;
 using WolvenKit.App.Helpers;
@@ -100,29 +99,30 @@ public partial class ProjectExplorerViewModel : ToolViewModel
     public ObservableCollection<FileSystemModel> DisplayedFileTree { get; } = new();
     private void ApplySearchFilter()
     {
-        DisplayedFileTree.Clear();
-
-        if (string.IsNullOrWhiteSpace(SearchText))
+        DispatcherHelper.RunOnMainThread(() =>
         {
-            // Normal mode - show full tree
-            foreach (var item in FileTree)
-            {
-                DisplayedFileTree.Add(item);
-            }
-            return;
-        }
+            DisplayedFileTree.Clear();
 
-        // Search mode
-        string search = SearchText.Trim();
-
-        foreach (var root in FileTree)
-        {
-            var filteredNode = FilterNode(root, search);
-            if (filteredNode != null)
+            if (string.IsNullOrWhiteSpace(SearchText))
             {
-                DisplayedFileTree.Add(filteredNode);
+                foreach (var item in FileTree)
+                {
+                    DisplayedFileTree.Add(item);
+                }
+                return;
             }
-        }
+
+            string search = SearchText.Trim();
+
+            foreach (var root in FileTree)
+            {
+                var filteredNode = FilterNode(root, search);
+                if (filteredNode != null)
+                {
+                    DisplayedFileTree.Add(filteredNode);
+                }
+            }
+        });
     }
 
     /// <summary>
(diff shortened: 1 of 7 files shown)


WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -82,6 +82,105 @@ public partial class ProjectExplorerViewModel : ToolViewModel
...
+
+    private FileSystemModel? FilterNode(FileSystemModel node, string search)
+    {
+        bool nameMatches = node.Name.Contains(search, StringComparison.OrdinalIgnoreCase) ||
+                           node.RawRelativePath.Contains(search, StringComparison.OrdinalIgnoreCase);
+
+        var matchingChildren = new ObservableCollection<FileSystemModel>();
+
+        // Normal recursive filtering
+        if (node.Children != null)
```
manavortex, line 145: You can probably LINQ that, something like: 
```c#
matchingChildren.AddRange((node.Children ?? []).Select(child => FilterNode(child, search).Select(n => n is not null))
```

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -82,6 +82,105 @@ public partial class ProjectExplorerViewModel : ToolViewModel
...
+                {
+                    matchingChildren.Add(filteredChild);
+                }
+            }
+        }
+
+        // === When folder matches → return original node (with all its real children) ===
+        if (nameMatches && node.IsDirectory)
+        {
+            return node;
```
manavortex: This should probably be in front of the matchingChildren check, since if this condition is true, nothing is done wtih the results

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -82,6 +82,105 @@ public partial class ProjectExplorerViewModel : ToolViewModel
...
+        }
+
+        if (nameMatches || matchingChildren.Count > 0)
+        {
+            // Only create a copy when we need to show a filtered list of children
+            var copy = new FileSystemModel(node.Parent, node.Name, node.RawRelativePath, node.IsDirectory);
```
manavortex: Not sure about this logic at all - @gistya , can you say something about it?

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -166,6 +265,7 @@ private void AppViewModel_OnInitialProjectLoaded(object? sender, EventArgs e)
         // On first project load, we're already initialized, so this won't fire
         Refresh();
         OnProjectChanged?.Invoke();
+        ApplySearchFilter();
```
manavortex: This happens only on project change - it would probably be better to clear the search filter here?

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
```diff
@@ -367,6 +467,9 @@ private void Refresh()
         {
             _projectWatcher.Refresh();
         }
+
+        // Сообщаем View, что нужно восстановить поиск (если он был активен)
```
manavortex: English pls :D

WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
```diff
@@ -48,6 +49,35 @@ public ObservableCollection<FileSystemModel> TreeItemSource
             set => SetValue(TreeItemSourceProperty, value);
         }
 
+        private void RefreshButton_Click(object sender, RoutedEventArgs e)
+        {
+            try
+            {
+                var vm = ViewModel; // capture on UI thread to avoid Dispatcher access from background threads
```
manavortex: make this 
```c#
if (ViewModel isn't {} vm) 
{
  return;
}
```
and put t in front of the `try`, it's much easier to read

WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
```diff
@@ -48,6 +49,35 @@ public ObservableCollection<FileSystemModel> TreeItemSource
...
+                }
+                else
+                {
+                    // Fallback: directly resume watcher which will trigger a preserve refresh
+                    vm.ResumeFileWatcher();
+                }
+            }
+            catch (Exception ex)
+            {
+                System.Diagnostics.Debug.WriteLine($"[ProjectExplorerView] RefreshButton_Click failed: {ex.Message}");
```
manavortex: Should we do this rather than using the logger?

WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
```diff
@@ -675,20 +793,175 @@ private void CollapseAll_OnClick(object sender, RoutedEventArgs e)
...
+                                TreeGrid.ExpandAllNodes();
+                            }, DispatcherPriority.Background);
+                        }
+                    }
+                }
+
+                DispatcherHelper.RunOnMainThread(() =>
+                {
+                    if (PESearchBar == null)
+                        return;
```
manavortex: Please use Allman Style as per formatting convention, even if it's wasteful – we should stay consistent

WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
```diff
@@ -61,12 +91,14 @@ public ObservableCollection<FileSystemModel> FlatItemSource
         private string _currentFolderQuery = "";
         private bool _isDragging;
         private CancellationTokenSource _deferRefreshTokenSource = new();
+        private int _lastSearchLength = 0;
```
manavortex: This should probably a boolean `_searchApplied`, since you're only checking if the length is 0 or not, rather than doing anything with it

manavortex (commented):
Needs some more work, and input from @gistya and maybe @notaspirit

notaspirit:
How does it relate to #2963? as they both aim to close the same issues

Ametis81:
> How does it relate to #2963? as they both aim to close the same issues

its most advanced in terms of progress i made, so i can close that PR

WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel.cs
Ametis81: can also do like this:

```c#
private FileSystemModel? FilterNode(FileSystemModel node, string search)
{
    bool nameMatches = node.Name.Contains(search, StringComparison.OrdinalIgnoreCase) ||
                       node.RawRelativePath.Contains(search, StringComparison.OrdinalIgnoreCase);

    var matchingChildren = new ObservableCollection<FileSystemModel>();

    // Recursively filter children
    if (node.Children != null)
    {
        foreach (var child in node.Children)
        {
            var filteredChild = FilterNode(child, search);
            if (filteredChild != null)
            {
                matchingChildren.Add(filteredChild);
            }
        }
    }

    // Nothing matches — exclude this node
    if (!nameMatches && matchingChildren.Count == 0)
        return null;

    // Create a copy of the node (safer than mixing original and filtered nodes)
    var copy = new FileSystemModel(node.Parent, node.Name, node.RawRelativePath, node.IsDirectory);

    if (nameMatches && node.IsDirectory)
    {
        // Folder name matches → show all original children (full folder view)
        if (node.Children != null)
        {
            foreach (var child in node.Children)
            {
                copy.Children?.Add(child);
            }
        }
    }
    else
    {
        // Only some children matched → add filtered children
        foreach (var child in matchingChildren)
        {
            copy.Children?.Add(child);
        }
    }

    return copy;
}
```

PR #2995, opened 2026-07-07
Add Linter GitHub Action on Pull Request

notaspirit:
# Add Linter GitHub Action on Pull Request

This will run `dotnet format` on the files changed by the pull request and create suggestions as review comments. 

Unfortunately the linter will also suggest breaking changes (e.g. making a public field readonly or making a public method static) so I can't just lint the entire codebase preemptively and the style suggestions will need to be resolved in the PR that touches that file first.

Despite this downside I think it will be a net positive as it means the reviewer doesn't need to manually check for code style violations and the reviewee gets immediate feedback and fix suggestions. 

If in the medium to long term it turns out to be more annoying than helpful we can always return to manual style verification.

### DIFF
--- a/.github/workflows/lint-pull-requests.yml
+++ b/.github/workflows/lint-pull-requests.yml
@@ -1,0 +1,40 @@
+name: Lint Pull Requests
+on:
+  pull_request_target:
+    paths:
+      - '**/*.cs'
+permissions:
+  contents: read
+  pull-requests: write
+jobs:
+  format-suggest:
+    runs-on: ubuntu-latest
+    env:
+      EnableWindowsTargeting: 'true'
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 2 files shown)


manavortex:
can the linter be trained out of those breaking changes? I'd rather not find out the hard way that it broke something >.<

notaspirit:
I have looked for ways to do that, but the only way would be to manually exclude all the codes that can cause these issues but that'd both not flag it when it's not breaking and you can't fully trust it since you always could've missed a code. 

I do want to have a convo and hear more of your guys' opinion on it before / if we merge as it does have it's downsides too.

manavortex:
Rider gets these suggestions mostly right, but if this is auto-linting, then I'd like to be able to rely on it blindly. Can't trust humans to catch the errors that the tools are making, and reverse centaurs are slow and stupid.

PR #2997, opened 2026-07-08
feat: add comment boxes

misterchedda:
Screenshot 2026-07-08 165142

Adds editor only comment boxes to quest and scene editors

- right-click on empty canvas to create comment
- comments can be moved, resized, renamed with double-click. Also recolored, reset, and deleted
- also: select multiple nodes and right-click on canvas to create comment enclosing the selected nodes
- comment state is persisted in a new `*.comments.json` file (living beside the existing graph layout json)
- shortcuts: "C" on canvas to add comment and "Delete" deletes it

Screenshot 2026-07-08 165722

### DIFF
--- a/WolvenKit.App/ViewModels/GraphEditor/GraphCommentState.cs
+++ b/WolvenKit.App/ViewModels/GraphEditor/GraphCommentState.cs
@@ -1,0 +1,19 @@
+using System.Collections.Generic;
+
+namespace WolvenKit.App.ViewModels.GraphEditor;
+
+internal sealed class GraphCommentState
+{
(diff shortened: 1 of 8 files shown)


manavortex (approved):
LGTM, I love it

notaspirit:
I don't like that it's parsing the json manually, but since it was already doing that before I won't be blocking on that but it should be adjusted to use a class that gets serialized and deserialized for maintainability in the future as a separate pr

misterchedda:
@notaspirit oh true for sure, you know what - i might actually do that (for the comments json atleast) in this pr right now when its cheap than later since it already sits standalone in its own json

PR #3002, opened 2026-07-11
Fix: quest scene graph perf quick fixes

misterchedda:
cherry picked a couple of self contained smaller performance related fixes from #2911 that are quick wins:

> Avoid loading external phase resources just for the phase node count

> Detach graph editors before document close teardown

(I'll keep #2911 as WIP for now, it'll eventually be a bigger graph virtualization refactor)

### DIFF
--- a/WolvenKit.App/ViewModels/GraphEditor/Nodes/Quest/questPhaseNodeDefinitionWrapper.cs
+++ b/WolvenKit.App/ViewModels/GraphEditor/Nodes/Quest/questPhaseNodeDefinitionWrapper.cs
@@ -47,8 +47,7 @@ public class questPhaseNodeDefinitionWrapper : questEmbeddedGraphNodeDefinitionW
             details.Add("Phase Resource", Path.GetFileName(_castedData.PhaseResource.DepotPath.GetResolvedText())!);
         }
         
-        // Add node count for the phase
-        var nodeCount = GetPhaseNodeCount();
+        var nodeCount = GetInlinePhaseNodeCount();
         if (nodeCount > 0)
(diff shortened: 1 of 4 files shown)


WolvenKit.App/ViewModels/GraphEditor/Nodes/Quest/questPhaseNodeDefinitionWrapper.cs
```diff
@@ -143,33 +142,15 @@ void InternalRecalculateSockets(graphGraphDefinition graphDefinition)
...
-    /// Get the total count of nodes in this phase graph
+    /// Get the node count from an inline phase graph without loading an external resource
     /// </summary>
-    private int GetPhaseNodeCount()
+    private int GetInlinePhaseNodeCount()
     {
-        // Check embedded graph first
         if (_castedData.PhaseGraph?.Chunk != null)
         {
             return _castedData.PhaseGraph.Chunk.Nodes?.Count ?? 0;
```
manavortex: This here looks as if the whole method could be condensed to that single line?

WolvenKit/Views/GraphEditor/GraphEditorView.xaml.cs
```diff
@@ -82,10 +82,13 @@ public NodeViewModel SelectedNode
         get => _selectedNode;
         set
         {
+            var previousNode = _selectedNode;
             if (SetField(ref _selectedNode, value))
```
manavortex: Good place for an early return

manavortex (approved):

PR #3003, opened 2026-07-11
Do not hide non empty parametersbuffer

manavortex:
# appearance: property hiding in easy editor mode

- parametersBuffer will no longer be hidden if it's not empty
- partsMasks will now be hidden if it's empty


misterchedda (approved):

notaspirit:
are parametersBuffer and partsMasks readonly?  (as in you'd never create or edit them manually) 
I definitely have concerns about hiding a property depending on it's contents as that can be more confusing than it always being there.

manavortex:
In my entire time modding this, I have never once created one of them, just deleted their contents for NPVs. Anyone who wants to access them is using an advanced workflow and won't mind switching out of easy editor mode.

notaspirit (approved):

PR #3010, opened 2026-07-12
Fix: comment box arrange nodes

misterchedda:
Fixes the 'Arrange Nodes' action to account for comment boxes, currently they just get ignored and nodes inside comment boxes get pulled out breaking comments. Also fixes a bug where nested comments couldn't be selected correctly


manavortex (approved):

gistya (commented):
Can you add tests please?

PR #3015, opened 2026-07-17
Live Search

Ametis81:
Live Search feature for Project Explorer - live search without using Enter button

Better search results - showing all files inside folders
Tested with @gistya latest PR

For Nightly first


Ametis81:
Can be closed since LS is  implemented now in main

PR #3016, opened 2026-07-17
feat: add graph action palette

misterchedda:
Replaces the separate quest and scene graph right click context menus with a shared, searchable action palette

image

This refactor also streamlines node template support:
- Cascading menu to view templates per node type
- Adds `Create Template from Node` to node right-click context actions
- Search works for both node types and template names
~~Adds 34 system templates for common quest nodes~~

image

~~@notaspirit : fyi `23e48d0e6b` also includes a fix in `WolvenKit/WolvenKit.csproj` so that system templates get copied in release builds~~

### DIFF
--- a/WolvenKit.App/ViewModels/Controls/RedTypeTemplateDropdownViewModel.cs
+++ b/WolvenKit.App/ViewModels/Controls/RedTypeTemplateDropdownViewModel.cs
@@ -54,7 +54,7 @@ public partial class RedTypeTemplateDropdownViewModel : ObservableObject
         SelectedRedTypeTemplate = ((List<RedTypeTemplateSelectionOption>)CurrentRedTypeTemplates.Source).First();
 
         RequestedType = typeof(object);
-        _ = Refresh();
+        RefreshFromRegistry();
     }
 
     partial void OnRequestedTypeChanged(Type value)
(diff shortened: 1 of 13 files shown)


manavortex:
Ohh, it's beautiful ♡ I don't have the time to test this right now, so I'd wait for one more green light before throwing this into Nightly?

notaspirit:
but why the copy on build changes? system templates are sourced from [WolvenKit/Wolvenkit-Resources](<https://github.com/WolvenKit/Wolvenkit-Resources>) just like wscripts

misterchedda:
ohhh right i see, that removes a lot of json from this pr :)

notaspirit:
though I just realized that I did mess up the path for system templates, missed a subdir, should be `Resources\Templates`. Opening a pr for it rn

misterchedda:
im adding like 34 templates now, we might need subfolder support soon too

notaspirit:
Not scanning subfolders when indexing templates is a deliberate design decision, if it were to scan sub folders we would need to handle duplicates. Whereas without subfolders the file system enforces unique name type combinations.

notaspirit:
Though we can reevaluate whether or not allowing subfolders is net positive in the future.

notaspirit:
> Ohh, it's beautiful ♡ I don't have the time to test this right now, so I'd wait for one more green light before throwing this into Nightly?

you're always free to merge into the nightly as long as the changes compile.

gistya (approved):
Tested locally, looks like a good addition. I did not test all the possible node menu options, but the state of the PR seems good enough to get merged and then any issues can always be iterated upon subsequently. There is certainly nothing breaking in here. The UI is very clean and nicely done. 

I would like to see some tests at some point in a follow-up PR, maybe UIAutomation tests if anyone has time I can show you how to make them. Might be good to have around this kind of thing. Just a thought.

gistya (approved):

gistya (approved):

misterchedda:
You're right but I might add them in a separate pr, this has been cooking for a while 👍

PR #3019, opened 2026-07-19
fix: ent preview dynamic substitution

manavortex:
# fix: ent preview dynamic substitution

Fixed an exception when trying to display a root entity preview with dynamic substitution.

### DIFF
--- a/WolvenKit.App/Helpers/ArchiveXlHelper.cs
+++ b/WolvenKit.App/Helpers/ArchiveXlHelper.cs
@@ -181,6 +181,8 @@ public static partial class ArchiveXlHelper
     /// Returns any existing depot path, or null. If no substitution is used, it will still check for the file's existence
     /// and return null if it can't be found.
     /// </summary>
+    /// <param name="depotPath">Depot path to resolve. Can be null, can be static.</param>
+    /// <param name="activeProject">As ProjectManager can be null in file, pass this from calling class where available.</param>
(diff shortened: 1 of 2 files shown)


WolvenKit.App/Helpers/ArchiveXlHelper.cs
```diff
@@ -181,11 +181,20 @@ List<string> ExtractVariants()
     /// Returns any existing depot path, or null. If no substitution is used, it will still check for the file's existence
     /// and return null if it can't be found.
     /// </summary>
+    /// <param name="depotPath">Depot path to resolve. Can be null, can be static.</param>
+    /// <param name="activeProject">As ProjectManager can be null in file, pass this from calling class where available.</param>
```
poirierlouis, line 185: wdyt of using `Optional depot path to resolve.` instead of adding `Can be null`?
What is the meaning of `can be static`?

We could explain it is optional and default to the ActiveProject, something like:
`Optional project to resolve with, default to the current ActiveProject.` I'm actually wondering if this argument is ever used with something else?

poirierlouis (approved):
LGTM

WolvenKit.App/Helpers/ArchiveXlHelper.cs
manavortex: The method gets called directly with `GetResolvedText()` - so the resulting string can be null, and it can be static (as in, a non-dynamic string), which means that there's nothing to resolve here

PR #3023, opened 2026-07-22
quest stuff: more cvm descriptors

manavortex:
# quest stuff: more cvm descriptors

I don't want to create a change log every time I do one of these :( they're such minor changes (just a lot of them)

image

### DIFF
--- a/WolvenKit.App/Helpers/SceneEditingHelper.cs
+++ b/WolvenKit.App/Helpers/SceneEditingHelper.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using WolvenKit.RED4.Types;
 
 namespace WolvenKit.App.Helpers
@@ -17,21 +18,21 @@ namespace WolvenKit.App.Helpers
         public static scnActorDef CreateAndAddActor(scnSceneResource sceneResource, string? actorName = null, gameEntityReference? entityRef = null)
         {
             var actor = new scnActorDef();
-            
+
             if (!string.IsNullOrEmpty(actorName))
             {
                 actor.ActorName = actorName;
             }
-            
+
             if (entityRef != null)
             {
                 actor.FindActorInWorldParams.ActorRef = entityRef;
             }
-            
+
             sceneResource.AddActor(actor);
             return actor;
         }
-        
+
         /// <summary>
         /// Creates a new player actor with automatic ID calculation and adds it to the scene
         /// </summary>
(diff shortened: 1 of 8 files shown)


WolvenKit.App/Helpers/StringHelper.cs
```diff
@@ -156,6 +156,15 @@ public static string Stringify(CArray<scnOutputSocket> scnOutputAry)
         return Stringify(nodeIds);
     }
 
+    public static string Stringify(scnAnimTargetBasicData animEvtData, scnSceneResource scene)
```
misterchedda, line 170: can we add handling for prop ids too? you could branch between Actor and Prop, resolving props through TargetPropId and scene.Props

manavortex: I just saw the comment - that'd require me to resolve the prop ID, and brain isn't up to that anymore. Other than that, happily!

manavortex: image
I implemented a bunch of them. Resolving the performerID for game props is of course using a different logic than the one for generated props, but CDPR aren't consistent, so there's no need to be (besides, I'm just trying to resolve these).

misterchedda:
Fixed the performer resolution using the correct formula (which is 2 + propId * 256 for prop performer ids and direct scnPropId fields will remain direct lookups), the previous one would interpret valid actor performer IDs as unrelated props

WolvenKit.App/Helpers/SceneEditingHelper.cs
```diff
@@ -17,21 +18,21 @@ public static class SceneEditingHelper
...
         {
             var actor = new scnActorDef();
-            
+
             if (!string.IsNullOrEmpty(actorName))
             {
                 actor.ActorName = actorName;
             }
-            
+
```
gistya, line 26: Maybe we can revert these whitespace changes before merge?

gistya (approved):
This all looks good to me. Thanks for adding tests with it.

Nit: consider rolling back the unneccessary whitespace changes before merge though?

WolvenKit.App/Helpers/SceneEditingHelper.cs
manavortex: That's probably "trim trailing whitespaces" by autosave action - can just set github's diff to ignore whitespace changes, and all is well

PR #3032, opened 2026-07-23
refactor: unify project pack/build logic between App and CLI

Zhincore:
# refactor: unify project pack/build logic between App and CLI

> [!NOTE]
> Apologies for AI slop, I did my best to keep Claude leashed.

Please, let me know if this is okay, it's a pretty big refactor. I plan to do more project-related CLI tooling, so I would appreciate this change.

RED4Controller and Modkit's Build.cs had drifted duplicate implementations of packing a Cp77 project into packed/ (Build.cs was missing REDmod folder layout, info.json, and script/tweak/sound packing entirely).

**Additional notes:**
Moves Cp77Project (plus FileSystemArchive and ModInfo/ModInfoEntry, both required by the move) from WolvenKit.App into WolvenKit.Modkit, and adds PackProject/CleanPackedDirectory/CreateZip instance methods on Cp77Project merged from RED4Controller's more complete implementation. Build.cs and RED4Controller now both call into the same shared code; RED4Controller keeps only its App-specific orchestration (notifications, install/deploy, game launch). Reference-scanning methods that depended on App-only services were split out into an extension-method file that stays in App.


notaspirit:
What plans do you have for the CLI that it requires the project? 

Right now I'm not fundamentally opposed to it but it is a big shift in ownership as now all the project related changes will need to also be considered for breaking code changes and for it's effect on the CLI. 
Especially relevant right now with the file tracking / project explorer changes and the discussion about project validation.

seberoth:
Merging / Unifying handling for AXL / TXL files should be done, but in the past we decided against of implementing of Redmod in CLI cause we thought it wouldn't be worthwhile to add a CLI wrapper around a CLI tool.

We saw no benefit in it and just much work / hassle just for someone to be able to call `wkit pack -redmod ...` if he could simply call `redmod archive ...` (or whatever its called) directly

Zhincore:
@seberoth I agree with redmod CLI being redundant, my biggest issue was that our build command didn't really do much and spit out a lot of warnings for a regular project, now it properly creates `packed/` as App does.

@notaspirit Project seemed crucial to make the build process consistent. I would like to add project management tools, primarily `wolvenkit.CLI init` to create a new wkit project from CLI. I would like to make wkit cli more useful for script mod devs, I haven't yet figured out everything, maybe I'll want to create a symbolic link to the nested `source/resources/r6/scripts/Modname` and/or add a `<BuildScript>` to projectfile allowing to run red-cli or make with the `wolvenkit build` command.

I'm perfectly fine with waiting in the background and then resolving conflicts myself later, this would be a low prio feature.

Apologies for not discussing this beforehand, didn't seem like a big change before.

notaspirit:
the base premise makes sense to me, the cli and modkit already have some app project handling via the build command (which has been fixed to no longer pack the root of the folder in #3164) so unifying that makes sense. 
@gistya is also done with the large refactor so there shouldn't be conflicts stepping from two major changes at once.

Zhincore:
Since GitHub show many changes I have slopped an overview:

- `Cp77Project.cs` **(moved App → Modkit)** — same class, just relocated so CLI/Modkit can use it directly.
- `Cp77Project.Packing.cs` **(new)** — the actual payload: PackProject/CleanPackedDirectory/CreateZip, ported from RED4Controller's more complete packer (REDmod layout, info.json, script/tweak/sound packing — CLI's old Build.cs had none of this).
- `Build.cs` **(rewritten)** — CLI build command now just builds a Cp77Project and calls into Cp77Project.Packing.cs instead of hand-rolling its own logic.
- `RED4Controller.cs` (−390 lines) — trimmed to App-only orchestration (notifications, install/deploy, launch); packing delegated to the shared code.
- `Cp77ProjectReferenceScanning.cs` **(new)** — GetAllReferencesAsync/ScanForBrokenReferencePathsAsync split out from `Cp77Project.cs` as extension methods, kept in App since they depend on App-only services (unrelated to packing).
- `ModInfo.cs`/`ModInfoEntry.cs` **(moved)** — forced move: Cp77Project.GetInfo() returns ModInfo.
- `FileSystemArchive.cs` **(moved)** — forced move: Cp77Project.AsArchive() returns this type.
- `IGameControllerFactory.cs` (7 lines) — fixed a fully-qualified reference that broke when Cp77Project's namespace changed.
- Everything else (~50 files) — one-line using fixes for the namespace change, no logic touched.

PR #3034, opened 2026-07-23
Fix rows on pinned projects

gistya:
# Fixes an issue where pinned projects weren't scrollable

### DIFF
--- a/WolvenKit/Views/HomePage/Pages/WelcomePageView.xaml
+++ b/WolvenKit/Views/HomePage/Pages/WelcomePageView.xaml
@@ -376,6 +376,7 @@
                     <Grid
                         x:Name="ProjectsScrollContent"
                         Margin="0,0,6,24"
+                        ClipToBounds="False"
                         VerticalAlignment="Top"
                         SizeChanged="ProjectsScrollContent_OnSizeChanged">
                         <Grid.RowDefinitions>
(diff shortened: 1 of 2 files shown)


poirierlouis (changes requested):
Nice.

Are we sure about the header background highlight UX-wise?
We could go with:
- a higher contrast ratio + some padding
OR
- restore background color + add a divider at the bottom of the header + some padding

wdyt?

Scale UI 100%, 1080p screen:
Image

I tried with scale UI at 150% on my 1080p screen, when resizing at the minimum I'm getting this non-scrollable view:
Image

If you have a 4K screen (or maybe @manavortex can check it out) with scale UI at 200%, does resizing the window at the minimum ends up with the same issue?

manavortex:
4k or not 4k doesn't change a thing:
https://github.com/user-attachments/assets/bf3dc40e-9d20-4d7d-b9e2-67e8d034ba9a
https://github.com/user-attachments/assets/f7609585-b486-4018-bd31-f2d746f0d592

My suggestion to make the most of the existing real estate is this:
image

Still have a PR lying around to [make project color configurable](https://github.com/WolvenKit/WolvenKit/pull/2471). I don't insist on finishing it up myself (as evident by the age of the PR) - if anyone is bored and wants to finish it, please be my guest

gistya:
> My suggestion to make the most of the existing real estate is this:

I agree but, changing the header should be a separate PR, this PR is just to fix the scrolling behavior.

manavortex:
yeah, absolutely :) we just discussed it in this context, so I figured i'd drop this here

poirierlouis:
Right for the header.

As mana can also reproduce the issue on 4K when resizing at the minimum, is it something you could look into @gistya ?

gistya:
> Right for the header.
> 
> As mana can also reproduce the issue on 4K when resizing at the minimum, is it something you could look into @gistya ?

I already fixed it

poirierlouis (approved):
LGTM :)

A bonus would be to add corner radius on headers (like buttons on the side).

PR #3035, opened 2026-07-24
cvm: 'overwrite with selection' no longer has index offset by 1

manavortex:
# cvm: 'overwrite with selection' no longer has index offset by 1

**Fixed:**
- cvm: 'overwrite with selection' no longer has index offset by 1


notaspirit:
Do you have repro steps for the issue this solves? I couldn't notice any difference when using "Overwrite Selection with Single" between main and this pr.

manavortex:
Yeah, it was OK for single. Overwrite with multiple.

PR #3037, opened 2026-07-26
Fix graph editor lifecycle cleanup

misterchedda:
Fixes graph editor lifecycle cleanup to prevent stale graph objects from remaining subscribed

Addresses the progressive slowdown (reported by some users in discord) in long quest/scene editor sessions

### DIFF
--- a/WolvenKit/Views/Documents/QuestPhaseGraphView.xaml.cs
+++ b/WolvenKit/Views/Documents/QuestPhaseGraphView.xaml.cs
@@ -75,8 +75,27 @@ namespace WolvenKit.Views.Documents
             InitializeComponent();
             DataContextChanged += OnDataContextChanged;
 
+            QuestPhaseGraphEditor.SourceChanged += OnEditorSourceChanged;
+
             // Add keyboard shortcut for subgraph navigation (Tab key)
             KeyDown += OnKeyDown;
+        }
+
+        private void OnEditorSourceChanged(object? sender, RedGraph? graph)
+        {
+            if (_disposed)
+            {
+                return;
+            }
+
+            if (graph is null)
+            {
+                UnsubscribeFromGraph();
+                return;
+            }
+
+            SubscribeToGraph(graph);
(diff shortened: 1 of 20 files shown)


manavortex (dismissed):

notaspirit:
Can we stop rubber stamping approvals? The test build hadn't even finished and the changelog wasn't added yet and you already approved 300+ LOC that change behavior and introduce new tests. I don't see how anyone can review the changes to a meaningful degree in that time.

manavortex:
I started looking at the code when the commit was pushed, so I had several minutes' head start on the pull request. The logic changes seemed sound and the code looks good to me, so I approved it. If that level of pull request review is not good enough for you, perhaps you should draft up some sort of review guideline, but at that point I expect you to pay me a salary.

notaspirit:
Sorry for having brought it up this way here, wasn't the right thing to do.

WolvenKit/Views/Documents/QuestPhaseGraphView.xaml.cs
```diff
@@ -249,6 +252,8 @@ private void OnDataContextChanged(object sender, DependencyPropertyChangedEventA
                 CleanupEventSubscription();
             }
 
+            UnsubscribeFromGraph();
+
```
gistya: Really great PR but I have one requested change.

I found a serious pre-existing bug (from before your PR) that will interact poorly with these particular changes. 

On Discord I DM'd you a patch file you can apply to fix it, please check it out. I know it wasn't from your PR but it's definitely worth fixing while you're in here under the hood.

Once that's in place, happy to approve!

gistya (changes requested):

WolvenKit/Views/Documents/QuestPhaseGraphView.xaml.cs
misterchedda: great catch! thanks, will integrate

gistya (approved):
thanks for adding that fix. apprroved

PR #3038, opened 2026-07-26
add disclaimer for material preview via a banner

### LINKED ISSUE #2887: add disclaimer to mesh preview with materials
add disclaimer to mesh preview with materials as it is known to sometimes be unreliable, which should be communicated in app and not require external communication to people unfamiliar with the feature.

notaspirit:
# add disclaimer for material preview via a banner

**Implemented:**
- new banner in mesh preview informing the user about the possibility of the material preview being wrong

**Additional notes:**
closes #2887

image
image

### DIFF
--- a/WolvenKit/Views/Documents/RDTMeshView.xaml
+++ b/WolvenKit/Views/Documents/RDTMeshView.xaml
@@ -491,6 +491,18 @@
                         Content="Export Appearance with Rig" />
                 </Grid>
 
+                <Border Background="{StaticResource WolvenKitRed}"
+                        CornerRadius="2"
+                        Margin="{DynamicResource WolvenKitMarginTiny}"
+                        Padding="{DynamicResource WolvenKitMarginTiny}"
+                        HorizontalAlignment="Stretch">
+                    <TextBlock Text="The material preview is experimental and may be incorrect in some cases."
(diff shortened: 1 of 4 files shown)


WolvenKit/Views/Documents/RDTMeshView.xaml
```diff
@@ -491,6 +491,18 @@
                         Content="Export Appearance with Rig" />
                 </Grid>
 
+                <Border Background="{StaticResource WolvenKitRed}"
+                        CornerRadius="2"
```
poirierlouis: I'd recommend using `DynamicResource WolvenKitSmallRadius` to be scale UI-compatible

poirierlouis (approved):
Nice, outside little comment: LGTM

WolvenKit/Views/Documents/RDTMeshView.xaml
notaspirit: I've updated it to use a new dynamic tiny radius which at 100% scale is 2px so it still matches the buttons as small radius with 5px is a bit too round.

poirierlouis: Perfect then :)

PR #3041, opened 2026-07-28
Fix file system model spam

gistya:
# Performance optimization of FileSystemModel

Now that we have GridGuard maintaining a shadow tree, we don't need the domain models in FileWatcher to publish INotifcation events (since there's no grid listening to them). So we can optimize performance by not publishing from non-live-grid models. 

The other thing we can do is not calculate the expensive hashes on every model, instead just calculate the hash when/if it's ever needed. Further we can avoid having to reverse-walk the parents to find the root, instead we can just pass the project dir down through the file tree from parent to son/daughter/etc. 

Profiling revealed these were major performance bottlenecks and this PR cuts the time to rename a folder of 2500 files by more than half.

### DIFF
--- a/WolvenKit.App/Models/FileSystemModel.cs
+++ b/WolvenKit.App/Models/FileSystemModel.cs
@@ -4,8 +4,10 @@ using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using WolvenKit.Common;
+using WolvenKit.Core.Extensions;
 using WolvenKit.RED4.Types;
 
 namespace WolvenKit.App.Models;
@@ -14,6 +16,8 @@ public class FileSystemModel : INotifyPropertyChanged
 {
     public const string ProjectDirName = "<ProjectDir>";
 
+    private readonly bool _shouldPublish;
+    private readonly string _projectDirectory;
     private string _name;
     private string _gameRelativePath = null!;
     private long _fileSize;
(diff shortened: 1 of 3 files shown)


WolvenKit.App/Models/FileSystemModel.cs
```diff
@@ -39,8 +43,28 @@ public string GameRelativePath
...
+                {
+                    return hash;
+                }
+                else
+                {
+                    return ResourcePath.CalculateHash(GameRelativePath);
+                }
+            }
+
+            return 0;
```
gistya, line 63: Only compute this hash when it's asked for. This has massive performance savings.

WolvenKit.App/Models/FileSystemModel.cs
```diff
@@ -82,15 +106,26 @@ public bool IsExpanded
...
         Parent = parent;
+
+        if (Parent == null)
+        {
+            _projectDirectory = rawRelativePath;
+        }
+        else
+        {
+            _projectDirectory = Parent._projectDirectory;
+        }
```
gistya, line 122: Set the new _projectDirectory path from the parent. This will be null iff this is the _projectDir file model itself. Otherwise the parent will have it.

WolvenKit.App/Models/FileSystemModel.cs
```diff
@@ -130,63 +165,57 @@ public void UpdateFileInfo()
...
-        {
-            if (current.Parent?.Name != ProjectDirName && current.Name != ProjectDirName)
-            {
-                hashParts.Add(current.Name);
-            }
-
-            root = current;
-            current = current.Parent;
-        }
-        hashParts.Reverse();
```
gistya, line 147: Don't do any of this expensive computation anymore.

WolvenKit.App/Models/FileSystemModel.cs
```diff
@@ -208,20 +237,36 @@ public static string GetFileSizeStr(long fileSize)
...
             return false;
         }
 
         field = value;
-        OnPropertyChanged(propertyName);
+
+        if (ShouldPublish(propertyName))
+        {
+            OnPropertyChanged(propertyName);
+        }
```
gistya, line 255: Only publish updates to the SyncFusion grids if this model is one of the GridGuard models.

gistya (commented):
Comments for the other reviewers

PR #3043, opened 2026-07-28
Unit Test Checks

notaspirit:
# Unit Test Checks

**Implemented:**
- run unit tests on pr and main / dev pushes
- adjusted checks to only run when relevant based on file extension
- adjust checks to run on prs not targeting main or dev (useful for the breaking changes staging branch for example)

**Additional notes:**
- ~~had to temporarily disable toe `Serialize_Db_ExpectedOutput` unit test that has been failing with commits from as far back as early 2025. With the test each check would fail and be useless.~~
- updated `Serialize_Db_ExpectedOutput` unit test blob version so it passes

### DIFF
--- a/.github/workflows/unit-tests-on-pr.yml
+++ b/.github/workflows/unit-tests-on-pr.yml
@@ -1,0 +1,33 @@
+name: Unit Tests
+on:
+  pull_request:
+    paths:
+      - '**/*.cs'
+      - '**/*.csx'
+      - '**/*.csproj'
+      - '**/*.sln'
+      - '**/*.props'
+      - '**/*.targets'
+      - '**/*.resx'
+      - '**/*.xaml'
+  push:
+    branches: [ main, dev ]
+    paths:
+      - '**/*.cs'
+      - '**/*.csx'
+      - '**/*.csproj'
+      - '**/*.sln'
+      - '**/*.props'
+      - '**/*.targets'
+      - '**/*.resx'
+      - '**/*.xaml'
+
+jobs:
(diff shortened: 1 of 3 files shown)


Tests/WolvenKit.UnitTests/TweakDBTests.cs
seberoth: The blob version is wrong. Should be 8 instead of 5. That's why its failing

notaspirit: Will update it instead of removing the test, thanks for pointing out the specific issue.

.github/workflows/unit-tests-on-pr.yml
```diff
@@ -0,0 +1,33 @@
...
+      - '**/*.props'
+      - '**/*.targets'
+      - '**/*.resx'
+      - '**/*.xaml'
+
+jobs:
+  unit-tests:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
```
poirierlouis: better use latest version (v7)

poirierlouis (approved):
Will be nice to have!
LGTM

.github/workflows/unit-tests-on-pr.yml
notaspirit: they're already up to v7? only ever saw v2 everywhere lol

poirierlouis: Yep, see https://github.com/marketplace/actions/checkout

PR #3044, opened 2026-07-28
Update the .NET SDK to latest version

### LINKED ISSUE #2906: Upgrade from .NET 8 to .NET 10
.NET 8 is going EOL by November 2026, as well as .NET 10 having some compiler optimizations and also being an LTS release. 
.NET 10 does have some breaking changes that might affect wolvenkit which should be checked.

gistya:
# Updates the project to .NET 10.0.302 with Windows 11 target

**Implemented:**
- fixes some new errors/warnings
- removes vulnerabilities
- removes redundant dependencies
- homogenizes the SDK version across all targets

**Additional notes:**
- No regression testing was performed yet but in light use the app seemed to behave fine
- Probably best to deploy on Dev for awhile first

closes #2906


WolvenKit.Modkit/RED4/Tools/Animation/AnimRootMotion.cs
```diff
@@ -164,7 +163,7 @@ public static RootMotionType DecodeRootMotion(out JointsTranslationsAtTimes tran
 
                             if (transform.JointIndex != RootJointIndex)
                             {
-                                throw new ValueOutOfRangeException($"Motion extraction in joint {transform.JointIndex}, unable to handle this!,");
+                                throw new Exception($"Motion extraction in joint {transform.JointIndex}, unable to handle this!,");
```
poirierlouis, line 166: Why change the type of exception?

WolvenKit.Modkit/RED4/Tools/Animation/AnimRootMotion.cs
```diff
@@ -195,7 +194,7 @@ public static RootMotionType DecodeRootMotion(out JointsTranslationsAtTimes tran
...
```
poirierlouis, line 197: Same here

WolvenKit.Modkit/RED4/Tools/Animation/AnimRootMotion.cs
```diff
@@ -349,17 +348,6 @@ public static (RootMotionType rootMotionType, bool rootTransformsStripped) Extra
                     throw new ArgumentOutOfRangeException(nameof(incomingType), "Should never get here, invalid incoming RM type");
             }
 
-            if (incomingType is not RootMotionType.None or RootMotionType.Unknown)
```
poirierlouis, line 352: I'm not familiar with this part. Can you explain why this part can be removed?

poirierlouis: It doesn't seem related to .NET 10 btw

WolvenKit.RED4/Archive/IO/AnimationReader.cs
```diff
@@ -41,7 +41,7 @@ public EFileReadErrorCodes ReadBuffer(RedBuffer buffer)
                             var dataAddr = animBuff.DataAddress;
                             var bytes = new byte[dataAddr.ZeInBytes];
                             animDataBuffers[(int)((uint)dataAddr.UnkIndex)].Seek(dataAddr.FsetInBytes, SeekOrigin.Begin);
-                            animDataBuffers[(int)((uint)dataAddr.UnkIndex)].Read(bytes, 0, (int)((uint)dataAddr.ZeInBytes));
+                            animDataBuffers[(int)((uint)dataAddr.UnkIndex)].ReadExactly(bytes, 0, (int)((uint)dataAddr.ZeInBytes));
```
poirierlouis, line 44: What's the difference between Read and ReadExactly?

poirierlouis (commented):
It seems some changes are not directly related to .NET 10.
I'd appreciate your feedback regarding my comments.

WolvenKit.Modkit/RED4/Tools/Animation/AnimRootMotion.cs
gistya: We removed an insecure dependency that is where this exception type came from. Since the dependency is gone it has to become a different kind of exception. Just went with the default here, but if you have a suggestion please add it as a suggested change and I'll commit it.

WolvenKit.Modkit/RED4/Tools/Animation/AnimRootMotion.cs
gistya: See above reply 👯

WolvenKit.Modkit/RED4/Tools/Animation/AnimRootMotion.cs
gistya: The .NET 10 compiler has improved diagnostics regarding if a switch statement is comprehensive over all the cases. 

Here, it threw a new warning (which is treated as an error in our codebase) that the switch statement above is already comprehensive over all the cases of incomingType. Thus, this particular if statement is 100% redundant and must be removed to clear the warning-as-error.

WolvenKit.RED4/Archive/IO/AnimationReader.cs
gistya: .NET 10's improved compiler diagnostics threw a warning-as-error here because ReadExactly is the idiomatic usage whenever you have an exact byte range (Read is more open-ended). 

Microsoft wants people to migrate code to ReadExactly because they are planning future runtime optimizations around it.

WolvenKit.Modkit/RED4/Tools/Animation/AnimRootMotion.cs
poirierlouis: I see, `IndexOutOfRangeException` seems more explicit.

WolvenKit.Modkit/RED4/Tools/Animation/AnimRootMotion.cs
poirierlouis: Oh I see thanks for the explanation. It makes sense indeed.

WolvenKit.RED4/Archive/IO/AnimationReader.cs
poirierlouis: Ok, nice :)

notaspirit:
I've updated the pr body so it closes the issue for the .net10 upgrade.

Also would need a changelog for the .net upgrade and the removal of wkpackage n stuff from common as well as updated docs to link to .net10 download links.

I would be happy to do those if you don't want or don't have the time to. 

And well as discussed should be merged into main earliest after the August release so the September stable can have all the breaking changes in one release.

notaspirit (approved):

PR #3045, opened 2026-07-29
component operations: exception prevention, UX

manavortex:
# component operations: exception prevention, UX

**Fixed:**
- Exception when trying to force LOD level on empty component array


gistya (approved):
LGTM

PR #3046, opened 2026-07-29
FileValidation script: better error message if not found

manavortex:
# FileValidation script: better error message if not found

This happened to a user the other day, and they fixed it by moving WKit out of Program Files.

### DIFF
--- a/WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
+++ b/WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
@@ -153,7 +153,10 @@ namespace WolvenKit.Views.Documents
                 [
                     "File validation script not found!",
                     "Try deleting '%APPDATA%\\REDModding\\WolvenKit\\Scripts\\Wolvenkit_FileValidation.wscript',",
-                    "then restart Wolvenkit. If that does not help, please get in touch with the devs."
+                    "then restart Wolvenkit. If that does not help, make sure that Wolvenkit is installed in a location ",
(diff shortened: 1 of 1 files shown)


WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
```diff
@@ -153,7 +153,10 @@ private async Task RunFileValidationAsync()
                 [
                     "File validation script not found!",
                     "Try deleting '%APPDATA%\\REDModding\\WolvenKit\\Scripts\\Wolvenkit_FileValidation.wscript',",
-                    "then restart Wolvenkit. If that does not help, please get in touch with the devs."
+                    "then restart Wolvenkit. If that does not help, make sure that Wolvenkit is installed in a location ",
+                    "that is not controlled by windows (e.g. 'C:\\CyberpunkModding\\WolvenKit'.",
```
notaspirit: The example is missing the closing bracket. 
While you're here imo 'restricted' rather than 'controlled' is more precise and gets the point across better.

notaspirit (changes requested):
Just a typo and an optional wording suggestion.

WolvenKit/Views/Documents/RedDocumentViewMenuBar.xaml.cs
manavortex: File a suggestion and I'll approve it

notaspirit:
```suggestion
                    "that is not restricted by windows (e.g. 'C:\\CyberpunkModding\\WolvenKit')",
```

notaspirit (approved):

PR #3047, opened 2026-07-29
AXL generator: aabdi bugfix edition

manavortex:
# AXL generator: aabdi bugfix edition

Fixed incorrect handling of "hide in FPP" and "force hair" for generated items

### DIFF
--- a/WolvenKit.App/Services/ArchiveXlItemService.cs
+++ b/WolvenKit.App/Services/ArchiveXlItemService.cs
@@ -898,24 +898,57 @@ public partial class ArchiveXlItemService
     }
 
     /// <summary>
+    /// Make sure that we have a valid root entity. If one already exists in the project, check
+    /// if we can re-use it (e.g. if the user adds a new item with 'force_Hair', don't re-use one that
+    /// doesn't have it.
+    /// </summary>
+    private CR2WFile? CheckRootEntityPath(ArchiveXlClothingItem clothingItemData, Cp77Project activeProject)
+    {
+        var rootEntityAbsPath = Path.Combine(activeProject.ModDirectory, clothingItemData.RootEntityPath);
+        if (!File.Exists(rootEntityAbsPath))
+        {
+            if (_archiveManager.GetCR2WFile(EquipmentItemData.DefaultRootEntityPath) is not CR2WFile defaultFile)
+            {
+                return null;
+            }
+            ((entEntityTemplate)defaultFile.RootChunk).Appearances.Clear();
+            return defaultFile;
+        }
+
+        // file already exists
+        if (_cr2WTools.ReadCr2W(rootEntityAbsPath) is not { RootChunk: entEntityTemplate rootEntity } f)
+        {
+            return null;
+        }
+
+        var tags = rootEntity.VisualTagsSchema.Chunk ?? new entVisualTagsSchema();
+        tags.VisualTags ??= new redTagList();
+        var tagsArray = tags.VisualTags.Tags ?? [];
+
+        if (!clothingItemData.TagsForceHair || tagsArray.Contains("force_Hair"))
+        {
+            return f;
+        }
+
+        // We can't use this root entity - update relative path
+        clothingItemData.RootEntityPath = Path.Combine(clothingItemData.ControlFilesRelPath,
+            $"_{clothingItemData.ItemName}_root_entity.ent").ToArchiveFilePath();
+
+        if (_archiveManager.GetCR2WFile(EquipmentItemData.DefaultRootEntityPath) is not CR2WFile f2)
+        {
+            return null;
+        }
+        ((entEntityTemplate)f2.RootChunk).Appearances.Clear();
+        return f2;
+    }
+
+    /// <summary>
     /// Creates root entity from base game file and adjusts path and tags
     /// </summary>
     private void AddRootEntity(ArchiveXlClothingItem clothingItemData, Cp77Project activeProject)
     {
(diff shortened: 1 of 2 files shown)


WolvenKit.App/Services/ArchiveXlItemService.cs
```diff
@@ -959,16 +992,6 @@ private void AddRootEntity(ArchiveXlClothingItem clothingItemData, Cp77Project a
...
         }
 
-        if (clothingItemData.TagsHideInFpp && !tagsArray.Contains("EmptyAppearance:FPP"))
-        {
-            tagsArray.Add("EmptyAppearance:FPP");
-        }
-
-        if (clothingItemData.TagsForceHair && !tagsArray.Contains("force_Hair"))
-        {
-            tagsArray.Add("force_Hair");
```
gistya, line 969: Do we need to reinstate adding "force_hair" somewhere else? Seems like this is removed here but I don't see where we put it back. (Probably just missing something)

Just want to make sure if the user is ticking "force hair" in AddArchiveXlFilesDialogViewModel (TagsForceHair, line 116) then we don't now produce a root entity without the "force_hair" tag.

WolvenKit.App/Services/ArchiveXlItemService.cs
```diff
@@ -898,24 +898,57 @@ void AdjustMeshAppearances(CMesh mesh, bool isSecondaryComponent = false)
...
+        {
+            return null;
+        }
 
-        if (File.Exists(rootEntityAbsPath) && _cr2WTools.ReadCr2W(rootEntityAbsPath) is CR2WFile f)
+        var tags = rootEntity.VisualTagsSchema.Chunk ?? new entVisualTagsSchema();
+        tags.VisualTags ??= new redTagList();
+        var tagsArray = tags.VisualTags.Tags ?? [];
+
+        if (!clothingItemData.TagsForceHair || tagsArray.Contains("force_Hair"))
```
gistya, line 928: Below we have the removal:

https://github.com/WolvenKit/WolvenKit/pull/3047/changes#diff-919126a63d7ac51937acc0ef77e2c3a188022834a833aefa5e42d573cca8f246L967-L970

Does that mean the "force_hair" tag is never persisted now? If so then tagsArray.Contains("force_Hair") at line 928 here is always false. 

So with TagsForceHair == true and an existing root entity, does this method always fall through to the "can't use this root entity" conditon? i.e.

`clothingItemData.RootEntityPath = Path.Combine(clothingItemData.ControlFilesRelPath,
    $"_{clothingItemData.ItemName}_root_entity.ent").ToArchiveFilePath();`

WolvenKit.App/Services/ArchiveXlItemService.cs
```diff
@@ -1017,7 +1040,9 @@ private void AddAppFile(ArchiveXlClothingItem clothingItemData, Cp77Project acti
 
             CArray<CName> tags = [.. clothingItemData.HidingTags.Select(t => (CName)t.ToString()).ToArray()];
 
-            appAppearance.Name = name;
+            var suffix = clothingItemData.TagsHideInFpp ? "&camera=tpp" : "";
+            appAppearance.Name = $"{name}{suffix}";
```
gistya, line 1044: What happens  for a hide-in-FPP item, where the persisted name is foo&camera=tpp, which is not in appearanceNames?

Am I reading this wrong or wouldn't the next generation pass find no match and set isAdding = true... then appends a second copy of every appearance? Seems like a hint on lline 1070 ("Appearance already exists and we're just recreating it") except are now or ..?

Just something to check

Maybe thelookup needs to use the suffixed name, e.g.: compute suffix before the loop, then check:

`s == $"{name}{suffix}`

or such

WolvenKit.App/Services/ArchiveXlItemService.cs
gistya, line 308: What happens here if we now have multiple `_root_entity.ent` files? 

Or can we not end up from the force-hair case, with a `_{itemname}_root_entity.ent`? Seems like if that could happen there may then be nondeterministic behavior?

WolvenKit.App/Services/ArchiveXlItemService.cs
manavortex: I don't understand the question. Anything without the suffix will be treated as a new appearance. That one would show regardless of camera.

WolvenKit.App/Services/ArchiveXlItemService.cs
manavortex: If the item can't fit into the first generated root entity, a new entity will be created for each item. Could probably refactor this to use the tags as suffixes instead of the item names, but... can do that later.

WolvenKit.App/Services/ArchiveXlItemService.cs
manavortex: if the existing root entity has this tag, then no item without it can go in here. That was the initial finding that triggered this PR.

WolvenKit.App/Services/ArchiveXlItemService.cs
gistya: Maybe easier if I show you with tests. I wouldn't trust my intuition without tests anyway. I'll post a branch that adds some tests to make sure that I am understanding this correctly.

WolvenKit.App/Services/ArchiveXlItemService.cs
gistya: Well OK, just need some tests to make sure. I'll write em and post PR to this PR

notaspirit (approved):

PR #3052, opened 2026-07-30
Fix some Live Search issues

gistya:
# Fixes some live search issues reported today

- fixes issue where folders of search results weren't expanding
- - fixes issues where expansion state remained after search result was cleared

### DIFF
--- a/WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
+++ b/WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
@@ -69,10 +69,26 @@ namespace WolvenKit.Views.Tools
         }
 
         private string _currentFolderQuery = "";
+
+        private HashSet<string> _searchVisiblePaths;
+
         private readonly DispatcherTimer _searchDebounceTimer;
         private bool _isDragging;
         private ISettingsManager _settingsManager;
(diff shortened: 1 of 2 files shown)


WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
```diff
@@ -533,6 +550,8 @@ private void OnToggleFlatMode(object sender, EventArgs e)
                 TreeGridFlat.SetCurrentValue(VisibilityProperty, Visibility.Collapsed);
             }
 
+            //ReapplyCurrentSearchFilter(expandAllForSearch: !string.IsNullOrWhiteSpace(_currentFolderQuery));
+
             PESearchBar_OnSearchStarted(this, new FunctionEventArgs<string>(_currentFolderQuery));
```
gistya: CHASE - which of these should we keep?

gistya (commented):
just a question

WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
Ametis81: Keep only the direct ReapplyCurrentSearchFilter call. Drop the PESearchBar_OnSearchStarted line (and the commented one).

PR #3055, opened 2026-07-31
Fix for CR2WReader

seberoth:
# Fix for CR2WReader

**Fixed:**
- Prevent an issue, where buffer readers could try to access uninitialized buffers

### DIFF
--- a/WolvenKit.RED4/Archive/IO/CR2WReader.File.cs
+++ b/WolvenKit.RED4/Archive/IO/CR2WReader.File.cs
@@ -133,18 +133,19 @@ public partial class CR2WReader
             ReadChunk(i);
         }
 
+        var clones = new List<RedBuffer>();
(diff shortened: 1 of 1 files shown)


seberoth:
The animation reader should be refactored in the feature, since its the one where the problem occurs with the switch to `ReadExactly`. The buffers in question gets parse multiple (unneeded) times and tries to access all the other buffers.

In the past that was fine, since the first iterations would yield invalid data but the last one would have access to all of the buffers and succeed.

seberoth:
Only needed for dev for now. So please merge to dev if approved.

gistya (commented):
While I lack the ability to approve PRs in this repo, it looks good to me.

Should this be part of the .NET 10 update PR since it fixes the ReadExactly stuff?

seberoth:
Closed in favor of #3044

PR #3060, opened 2026-08-01
Fix MorphTarget import for GLBs with GarmentSupport

seberoth:
# Fix MorphTarget import for GLBs with GarmentSupport

**Fixed:**
- Fixed MorphTarget import for GLBs with GarmentSupport

**Additional notes:**
The import checked for the attributes `_GARMENTSUPPORTWEIGHT` and `_GARMENTSUPPORTCAP` before. These are not needed for proper GS support and aren't always available.
This caused issues then the MorphTarget for `GarmentSupport` were available but the attributes not.
Also if the attributes are in the GLB but not a MorphTarget with the name `GarmentSupport` this should be ignored here, since it would skip an MorphTarget witch would also be wrong in this case.

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/MorphTargetImportTools.cs
+++ b/WolvenKit.Modkit/RED4/Tools/MorphTargetImportTools.cs
@@ -2,18 +2,13 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using Microsoft.Extensions.DependencyModel;
+using System.Text.Json;
 using SharpGLTF.Schema2;
-using WolvenKit.Common.FNV1A;
 using WolvenKit.Common.Model.Arguments;
 using WolvenKit.Core.Exceptions;
-using WolvenKit.Core.Extensions;
 using WolvenKit.Modkit.RED4.GeneralStructs;
-using WolvenKit.Modkit.RED4.RigFile;
 using WolvenKit.Modkit.RED4.Tools;
-using WolvenKit.RED4.Archive;
 using WolvenKit.RED4.Archive.CR2W;
-using WolvenKit.RED4.CR2W;
 using WolvenKit.RED4.Types;
 
 using Vec3 = System.Numerics.Vector3;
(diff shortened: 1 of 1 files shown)


WolvenKit.Modkit/RED4/Tools/MorphTargetImportTools.cs
```diff
@@ -59,27 +54,36 @@ public bool ImportMorphTargets(FileInfo inGltfFile, Stream inTargetStream, GltfI
...
-                    submesh.Primitives[0].GetMorphTargetAccessors(t)
-                );
+                var primitive = submesh.Primitives[0];
+                var targets = Enumerable.Range(0, primitive.MorphTargetsCount)
+                    .Select(primitive.GetMorphTargetAccessors)
+                    .ToArray();
 
-                var hasGS = submesh.Primitives[0].GetVertexAccessor("_GARMENTSUPPORTWEIGHT") != null &&
-                            submesh.Primitives[0].GetVertexAccessor("_GARMENTSUPPORTCAP") != null;
+                var targetNames = submesh.Extras?.Deserialize<MeshExtras>()?.TargetNames?.ToList() ?? [];
```
gistya, line 62: targetNames falls back to [] when extras are missing, so any submesh with morph targets and no extras.targetNames now throws instead of importing. (previously i think that would have worked so its worth checking if this could be a regression).

seems low risk (limited to third party or manipulated glbs, not blender/wk ones), but note `MeshImportTools.ExtractMorphTargets` takes the opposite behavior on identical input... it returns silently when TargetNames is null and only throws on a count mismatch.

WolvenKit.Modkit/RED4/Tools/MorphTargetImportTools.cs
```diff
@@ -59,27 +54,36 @@ public bool ImportMorphTargets(FileInfo inGltfFile, Stream inTargetStream, GltfI
...
 
-            var garmentSupportPerSubmesh = gsAndTargetsPerSubmesh.Select(_ => _.gs).ToArray();
-            var realTargetsPerSubmesh = gsAndTargetsPerSubmesh.Select(_ => _.real).ToArray();
+                if (targetNames.Skip(gsIndex + 1).Contains("GarmentSupport"))
+                {
+                    throw new ArgumentException($"Submesh '{submesh.Name}' contains multiple GarmentSupport targets, but only one is allowed per submesh.");
+                }
 
-            var hasAtMostOneGsPerSubmesh = garmentSupportPerSubmesh.All(gs => gs.Length <= 1);
+                var realTargets = targets.Length == 1 ? [] : targets.Where((_, i) => i != gsIndex).ToArray();
```
gistya, line 80: this seems like a dead branch -- reaching here implies that gsIndex == 0, i.e. `Where` always yields `[]`.

gistya (commented):
Good fix! 

I lack the ability to approve PRs so I just left a couple of minor comments.

WolvenKit.Modkit/RED4/Tools/MorphTargetImportTools.cs
seberoth: Yes, this assumes what the glb comes from our Blender addon, which always exports these names (at least I hope so, haven't checked...)

Before the change they weren't needed, since it only checked for the attributes (which need to have a name in the GLB, thats why no extra property is needed for them.

But if we switch to check if there is an actual GS MorphTarget in the GLB, i only can do it via its names, where GLBs doesn't have names for. So I need to rely on this extra field.

WolvenKit.Modkit/RED4/Tools/MorphTargetImportTools.cs
seberoth: Reaching here implies that `gsIndex > -1`.
So it separates the GS MorphTargets (`targets[gsIndex]`) from the normal MorphTargets (everything besides gsIndex).

Or am I blind?

gistya: sorry.. my comment was unclear. to be clearer- the existing code is not a bug at all, it just has an extremely minor redundancy. 

i just meant that `targets.Length == 1 ? []` is redundant, because the Where on its own already produces an empty array when targets == ["GarmentSupport"]

1. gsIndex != -1 like you said
2. gsIndex is now guaranteed to be the index of "GarmentSupport" in the targets array
3. if targets.Length == 1 then targets == ["GarmentSupport"] and gsIndex == 0
4. thus `["GramentSupport"].Where((_, i) => i != gsIndex).ToArray()` already gives `[]`
5. so you can replace line 80 with `var realTargets = targets.Where((_, i) => i != gsIndex).ToArray();`

anyway sorry sometimes i'm too detail oriented lol you can ignore me

gistya (approved):
looks good.

WolvenKit.Modkit/RED4/Tools/MorphTargetImportTools.cs
gistya, line 80:
```suggestion
                var realTargets = targets.Where((_, i) => i != gsIndex).ToArray();
```

gistya (commented):
minor fix if u want it

manavortex (approved):

PR #3063, opened 2026-08-02
Quest graph loading optimization

gistya:
# Quest graph loading optimizations

**Fixed:**
- shaves 45 seconds off of a 53-second graph load by setting SyncFusion to use the default style (makes no visible difference because nothing under Views/GraphEditor even uses a single Syncfusion control, so that work was pure overhead)
- parallelize graph building in a threadsafe manner and optimize the tree algorithm, bringing it down from O(N^2) to O(N). 
- fixed the issue where adding a node would do O((N^2)/2+N) work; now it just does O(1)
- Previously two identical TextBlocks with complementary Subtitle triggers - one for scene nodes, one for quest nodes - which rendered the same thing either way, accounted for 2/3rds of title rendering. Fixed by collapsing them into one always-visible element
- index once instead of scanning per-saved-node
- cache resource lookups for brush color and title strings (going to Application.Resources is expensive)

**Additional notes:**

Nodify has no virtualization features at all. So there will be a follow-up PR to implement lazy loading, but that's a riskier change and actually loads slower overall so it probably belongs as a separate PR.

### DIFF
--- a/WolvenKit/Views/GraphEditor/GraphEditorView.xaml
+++ b/WolvenKit/Views/GraphEditor/GraphEditorView.xaml
@@ -428,31 +428,15 @@
                                     IsConnected="{Binding IsConnected}">
                                     <nodify:NodeInput.Header>
                                         <StackPanel Orientation="Horizontal">
-                                            <!-- Scene nodes: Show Title FIRST and BIG (socket name) -->
+                                            <!--
+                                                Title (socket name) for Scene or Quest nodes.
+                                            -->
                                             <TextBlock
                                                 VerticalAlignment="Center"
                                                 Foreground="{DynamicResource TextFillColorPrimary}"
                                                 FontSize="14"
                                                 FontWeight="Normal"
-                                                Text="{Binding Title}">
-                                                <TextBlock.Style>
-                                                    <Style TargetType="TextBlock">
-                                                        <Style.Triggers>
-                                                            <!-- Hide if no subtitle (quest nodes) -->
-                                                            <DataTrigger
-                                                                Binding="{Binding Subtitle}"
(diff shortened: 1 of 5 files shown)


WolvenKit/Views/GraphEditor/GraphEditorView.xaml
```diff
@@ -428,31 +428,15 @@
...
-                                            <!-- Scene nodes: Show Title FIRST and BIG (socket name) -->
+                                            <!--
+                                                Title (socket name) for Scene or Quest nodes.
+                                            -->
                                             <TextBlock
                                                 VerticalAlignment="Center"
                                                 Foreground="{DynamicResource TextFillColorPrimary}"
                                                 FontSize="14"
                                                 FontWeight="Normal"
-                                                Text="{Binding Title}">
```
manavortex, line 437: Now there's no subtitle anymore?

gistya: > Now there's no subtitle anymore?

Nothing should be missing. 

Please let me know if, when you try the branch, anything looks or works differently. The only thing this PR should do is make the graph load faster.

If there is a regression then I'll revert the part of the PR that affected it.

misterchedda: I think it's just a duplicate, they become one shared binding

misterchedda (approved):

WolvenKit/Views/GraphEditor/GraphEditorView.xaml
gistya: That was the idea, it was to save on render cycles.

PR #3069, opened 2026-08-03
refactor: unify graph node context menu construction

misterchedda:
Refactors the right-click graph node context menus into a shared implementation across the quest, scene, and behavior graphs

No user facing changes except adding tooltips for node actions

image

### DIFF
--- a/WolvenKit/Views/GraphEditor/GraphContextMenuBuilder.cs
+++ b/WolvenKit/Views/GraphEditor/GraphContextMenuBuilder.cs
@@ -1,0 +1,120 @@
+using System;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Controls.Primitives;
+using System.Windows.Media;
+using WolvenKit.Views.Templates;
+
+namespace WolvenKit.Views.GraphEditor;
+
+#nullable enable
+internal sealed class GraphContextMenuBuilder
+{
+    private readonly ContextMenu _menu;
+    private readonly Style _itemStyle;
+    private readonly Style _separatorStyle;
+    private bool _separatorPending;
+
+    public GraphContextMenuBuilder(ContextMenu menu, Style itemStyle, Style separatorStyle)
+    {
+        _menu = menu;
+        _itemStyle = itemStyle;
+        _separatorStyle = separatorStyle;
+        _menu.Items.Clear();
+    }
+
+    public void StartSection() => _separatorPending = _menu.Items.Count > 0;
+
+    public MenuItem AddAction(string header, string iconKind, Action click, string? description = null) =>
+        AddAction(header, iconKind, null, click, description);
+
+    public MenuItem AddAction(string header, string iconKind, string? iconColor, Action click, string? description = null)
+    {
+        var item = CreateMenuItem(header, iconKind, iconColor, click, description);
+        AddItem(item);
+        return item;
+    }
+
+    public MenuItem AddCategory(string header, string? description = null)
+    {
+        var item = CreateMenuItem(header, "FolderOutline", null, null, description);
+        AddItem(item);
+        return item;
+    }
+
+    public void AddItem(MenuItem item, bool applyDefaultStyle = true)
+    {
+        AddPendingSeparator();
+        if (applyDefaultStyle)
+        {
+            item.Style = _itemStyle;
+        }
+
+        _menu.Items.Add(item);
+    }
+
+    public void Open(ContextMenuEventArgs e)
+    {
+        _menu.SetCurrentValue(ContextMenu.IsOpenProperty, true);
+        e.Handled = true;
+    }
+
+    public static MenuItem CreateMenuItem(
+        string header,
+        string iconKind,
+        string? iconColor,
+        Action? click,
+        string? description = null)
+    {
+        var item = new MenuItem
+        {
+            Header = header,
+            Padding = (Thickness)Application.Current.Resources["WolvenKitMarginTiny"]!
+        };
+
+        if (!string.IsNullOrWhiteSpace(description))
+        {
+            item.ToolTip = description;
+            ToolTipService.SetPlacement(item, PlacementMode.Right);
+            ToolTipService.SetHorizontalOffset(item, 8);
+        }
+
+        if (iconKind is not null)
+        {
+            var hasIcon = iconKind != "Empty";
+            var icon = new IconBox
+            {
+                IconPack = hasIcon ? IconPackType.Material : IconPackType.Empty,
+                Kind = hasIcon ? iconKind : "",
+                Margin = new Thickness(4, 0, 2, 0),
+                Size = (double)Application.Current.Resources["WolvenKitIconMicro"]!
+            };
+
+            if (!string.IsNullOrEmpty(iconColor) &&
+                Application.Current.Resources[iconColor] is Brush brush)
+            {
+                icon.Foreground = brush;
+            }
+
+            item.Icon = icon;
+        }
+
+        if (click is not null)
+        {
+            item.Click += (_, _) => click();
+        }
+
+        return item;
+    }
+
+    private void AddPendingSeparator()
+    {
+        if (!_separatorPending)
+        {
+            return;
+        }
+
+        _menu.Items.Add(new Separator { Style = _separatorStyle });
+        _separatorPending = false;
+    }
+}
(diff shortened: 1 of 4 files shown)


Akiway:
In Scene graph, we lost the help link that redirect to `https://wiki.redmodding.org/wolvenkit/wolvenkit-app/editor/scene-editor`.
image

I think it can could be great to also add a link in quest graph to link to `https://wiki.redmodding.org/wolvenkit/wolvenkit-app/editor/quest-editor`.

WolvenKit/Views/GraphEditor/GraphEditorView.xaml.cs
```diff
@@ -318,260 +318,28 @@ private void Node_ContextMenuOpening(object sender, ContextMenuEventArgs e)
...
-            infoItem.Click += (_, _) => {
-                try
-                {
-                    System.Diagnostics.Process.Start(new System.Diagnostics.ProcessStartInfo("https://wiki.redmodding.org/wolvenkit/wolvenkit-app/editor/scene-editor") { UseShellExecute = true });
-                }
-                catch (Exception)
-                {
-                    // Silently handle any exceptions when opening the link
-                }
-            };
```
gistya, line 567: This normally opens a wiki link, but I notice in the refactor you don't include the help link. Should we keep that or was it meant to be removed by the refactor?

WolvenKit/Views/GraphEditor/GraphContextMenuBuilder.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        if (iconKind is not null)
+        {
+            var hasIcon = iconKind != "Empty";
+            var icon = new IconBox
+            {
+                IconPack = hasIcon ? IconPackType.Material : IconPackType.Empty,
+                Kind = hasIcon ? iconKind : "",
+                Margin = new Thickness(4, 0, 2, 0),
```
gistya, line 89:
```suggestion
                Margin = (Thickness)Application.Current.Resources["WolvenKitMarginTiny"]!,   // = uniform 4 (App.Sizes.xaml:57) ... or add whichever one that you feel is appropriate
```

 Should we change this to use an app style? Typically prefer to see consistency on use of stylesheet/resources etc. over hardcoded values.

WolvenKit/Views/GraphEditor/GraphContextMenuBuilder.cs
```diff
@@ -0,0 +1,120 @@
...
+        _menu.SetCurrentValue(ContextMenu.IsOpenProperty, true);
+        e.Handled = true;
+    }
+
+    public static MenuItem CreateMenuItem(
+        string header,
+        string iconKind,
+        string? iconColor,
+        Action? click,
+        string? description = null)
```
gistya, line 67: It seems this method quietly also changes the look of other context menus besides just Graphs:

1. Editor background menu (Arrange Items / Hide-unhide sockets / Paste Node) (GraphEditorView.xaml.cs:266-274)
2. Grouping-node menu (GraphEditorView.xaml.cs:447-466)
3. "Search Node ..." (GraphEditorView.xaml.cs:885)
4. Connection menu ("Delete Connection") (GraphEditorView.xaml.cs:970)
5. Behavior child menus (Move Up/Down, Replace Child, Wrap Child With, Delete Child Branch) (GraphEditorView.xaml.cs:1009-1048)

Are we intending to change all of those or is this just supposed to impact the context menu? 

Because none of those others seem to get GraphContextMenuStyle or GraphContextMenuItemStyle, so they end up with the new icon spacing and the old menu chrome.

WolvenKit/Views/GraphEditor/GraphEditorView.xaml.cs
```diff
@@ -1065,33 +833,8 @@ private uint CreateNode(RedGraph graph, Type type, System.Windows.Point position
...
-            {
-                IconPack = (iconKind == "Empty") ? IconPackType.Empty : IconPackType.Material,
-                Kind = (iconKind == "Empty") ? "" : iconKind,
-                Margin = (Thickness)Application.Current.Resources["WolvenKitMarginTiny"]!,
-                Size = (double)Application.Current.Resources["WolvenKitIconMicro"]!
-            };
-
-            if (iconColor != null)
-            {
-                icon.Foreground = (Brush)Application.Current.Resources[iconColor] ?? Brushes.White;
```
gistya, line 1088: We used to give a default White color to the background as a fallback option but it seems now we don't?

WolvenKit/Views/GraphEditor/GraphContextMenuBuilder.cs
```diff
@@ -0,0 +1,120 @@
...
+                Kind = hasIcon ? iconKind : "",
+                Margin = new Thickness(4, 0, 2, 0),
+                Size = (double)Application.Current.Resources["WolvenKitIconMicro"]!
+            };
+
+            if (!string.IsNullOrEmpty(iconColor) &&
+                Application.Current.Resources[iconColor] is Brush brush)
+            {
+                icon.Foreground = brush;
+            }
```
gistya, line 97:
```suggestion
            if (!string.IsNullOrEmpty(iconColor) &&
                Application.Current.Resources[iconColor] is Brush brush)
            {
                icon.Foreground = brush;
            }
            else
            {
                icon.Foreground = Brushes.White;
            }
```

See my comment below on the removed line that this replaces. We used to give a fallback color of white. So I might suggest adding a fallback here so we're not changing the appearance unintentionally, unless that's your intent to remove the white fallback.

WolvenKit/Views/GraphEditor/GraphEditorView.xaml
```diff
@@ -21,25 +21,60 @@
         <!-- Node icon converter -->
         <converters:NodeIconConverter x:Key="NodeIconConverter" />
 
-        <!-- Style for help/info menu items -->
         <Style
-            x:Key="HelpMenuItemStyle"
+            x:Key="GraphContextMenuStyle"
+            TargetType="ContextMenu">
```
gistya, line 26: Should we add a `BasedOn` property for this Style? Seems it fully replaces ContextMenu style rather than extending it, which is inconsistent with GraphContextMenuItemStyle below, which does use BasedOn="{StaticResource {x:Type MenuItem}}".

This could be an issue because the default WPF ContextMenu template wraps the items host in a ScrollViewer; a bare StackPanel doesn't scroll or scroll-into-view on keyboard navigation, so an overly tall menu just clips. 

To me this seems like a blocker unless I'm mistaken, since we don't want unscrollable menus where you can't reach all the options... so I'm going to withhold approval until this comment is answered/resolved.

If we don't want to base off the WPF ContextMenu then at least wrap the content in a ScrollViewer?

(If I'm wrong then plz let me know)

WolvenKit/Views/GraphEditor/GraphContextMenuBuilder.cs
```diff
@@ -0,0 +1,120 @@
...
+        };
+
+        if (!string.IsNullOrWhiteSpace(description))
+        {
+            item.ToolTip = description;
+            ToolTipService.SetPlacement(item, PlacementMode.Right);
+            ToolTipService.SetHorizontalOffset(item, 8);
+        }
+
+        if (iconKind is not null)
```
gistya, line 82:
```suggestion
```

This is always true, don't need it.

WolvenKit/Views/GraphEditor/GraphContextMenuBuilder.cs
```diff
@@ -0,0 +1,120 @@
...
+        AddAction(header, iconKind, null, click, description);
+
+    public MenuItem AddAction(string header, string iconKind, string? iconColor, Action click, string? description = null)
+    {
+        var item = CreateMenuItem(header, iconKind, iconColor, click, description);
+        AddItem(item);
+        return item;
+    }
+
+    public MenuItem AddCategory(string header, string? description = null)
```
gistya, line 38: I notice none of the callers of this function are passing the "description" property. Should they be?

WolvenKit/Views/GraphEditor/GraphEditorView.NodeContextMenus.cs
```diff
@@ -0,0 +1,291 @@
...
+
+    private void AddDestroyDeletionMarkerAction(GraphContextMenuBuilder menu, NodeViewModel node) =>
+        menu.AddAction(
+            "Destroy Deletion Marker",
+            "CloseBoxOutline",
+            "WolvenKitRed",
+            () => Source.RemoveNode(node),
+            "Permanently removes this deletion marker from the graph");
+
+}
```
gistya, line 291:
```suggestion
            "Permanently removes this deletion marker from the graph");
}
```

Fix extra blank line here

gistya (changes requested):
Really solid, nice PR, nice work. I do have one small change request and a few minor questions. 

Happy to approve once those are resolved. Thanks MC!

gistya:
Also one other thing, it needs a changelog

PR #3071, opened 2026-08-03
Fix LoggerService performance issues

gistya:
# Fix LoggerService performance issues

**Fixed:**
- Fixes an issue where AddLog appended to LogEntries and let a CollectionChanged handler re-derive FilteredLogEntries: it re-scanned every entry ever logged, then Cleared the whole collection. This Reset in turn makes the ItemsControl tear down and regenerate every single row... and then force a ScrollToBottom layout pass. All of that **per message**, so the slowdown cost increased with the size of the log.
- Entries are now appended at O(1) per message. The full re-filter only runs when the user changes the level filter manually, and the scroll is coalesced to once per change set instead of once per entry.

Notes:

- This fix should help with an issue many users have reported where the app slows down over long usage sessions.
- A burst of log lines from a background thread would make the cost quadratic, starving the UI thread badly enough to visibly stall animations; archive loading writes one line per archive, which is why the welcome-screen fade dropped to a couple of frames per second while it ran. Fixing this log issue therefore unblocks moving archive loading to app launch, which will greatly improve load times for users getting into being able to edit a project.

### DIFF
--- a/WolvenKit/Views/Tools/LogView.xaml.cs
+++ b/WolvenKit/Views/Tools/LogView.xaml.cs
@@ -45,8 +45,6 @@ namespace WolvenKit.Views.Tools
             ViewModel = Locator.Current.GetService<LogViewModel>();
             DataContext = ViewModel;
 
-            LogEntries.CollectionChanged += LogEntries_CollectionChanged;
-
             var sink = Locator.Current.GetService<MySink>();
             _ = sink.Connect()
                 .ObserveOn(RxApp.MainThreadScheduler)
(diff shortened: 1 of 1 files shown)


manavortex (approved):

manavortex:
Works great - log is so much faster now :)

poirierlouis (approved):
LGTM.

I think the performance issue reported by users was mostly due to Syncfusion. I tried to fix data/virtualization without success in a past issue/PR. Someone else tried too but I can't remember who...
Though that's definitely a nice fix to have, thanks.

Issue #3067, opened 2026-08-02
Exporting/Importing some animations (base game, unedited, not custom) breaks them


PhilWright47:
Exporting and importing certain animations breaks them.
Example: `synced__sq027_09_new_camp_wakeup__player_kisses_panam__01_man_player_1` from `synced__sq027_09_new_camp_wakeup__player_male.anims` will no longer play at the intended position, and tries to play so far away that it triggers a teleport/loading screen

**To Reproduce**
1. Add the `base\animations\quest\side_quests\sq027\sq027_09_new_camp_wakeup\synced\synced__sq027_09_new_camp_wakeup__player_male.anims` file to project
2. Export the file
3. Import the file back in
4. See the issue in either the original scene (kiss panam after waking up in the new camp as part of the quest) or [easier] try to kiss Judy in her apartment as part of Judy Romanced Enhanced

**Expected behavior**
Nothing should change, the animation isn't being edited at all - I can't even get to the stage of editing it because of this bug

**Version:**
 - Latest seen WolvenKit-8.19.1-nightly.2026-08-01
 - Originally seen WolvenKit-8.17.2-nightly.2026-01-12

jono-loomy:
seems to be a buffer ordering of transform keys issue,  where original is sorted by Time ->TransFormType (rot/loc),  and wkit sorts it by TransformType->Time.  Assuming buffers are read from start to end,  ur translations and rotations are being processed seperately rather than simultaneously.   

*UPDATE*  Sorting the keys by time>type has solved this issue.

jono-loomy:
Fix:  https://github.com/WolvenKit/WolvenKit/pull/3074

PR #3072, opened 2026-08-04
Animation Buffer - Fix AnimKey Ordering

jono-loomy:
# Animation Buffer - Fix AnimKey Ordering

**Fixed:**
- Animation Buffer Data Ordering of Transforms/animKeys  (Sort Keys by Bone->Time->KeyType (pos->rot->sca),   instead of Bone->KeyType->Time), 

**Additional notes:**
addresses https://github.com/WolvenKit/WolvenKit/issues/3067

### DIFF
--- a/WolvenKit.Modkit/RED4/Tools/Animation/AnimSpline.cs
+++ b/WolvenKit.Modkit/RED4/Tools/Animation/AnimSpline.cs
@@ -228,6 +228,47 @@ namespace WolvenKit.Modkit.RED4.Animation
                 ((List<JointsScalesAtTimes>)[incomingAnimBufferData.Scales, incomingAnimBufferData.ConstScales]).All(_ =>
                     _.All(_ => _.Value.All(_ => EqWithEpsilon(_.Value, Scale1to1))));
 
+            // Sort Keys by Bone->Time->KeyType (pos->rot->sca)
+            animKeys.Sort((a, b) =>
+            {
+                int cmp = a!.Idx.CompareTo(b!.Idx);
+                if (cmp != 0) return cmp;
+            
+                cmp = a.Time.CompareTo(b.Time);
+                if (cmp != 0) return cmp;
+            
+                return string.Compare(
+                    a.GetType().Name,
+                    b.GetType().Name,
+                    StringComparison.Ordinal);
+            });
+            animKeysRaw.Sort((a, b) =>
+            {
+                int cmp = a!.Idx.CompareTo(b!.Idx);
+                if (cmp != 0) return cmp;
+            
+                cmp = a.Time.CompareTo(b.Time);
+                if (cmp != 0) return cmp;
+            
+                return string.Compare(
+                    a.GetType().Name,
+                    b.GetType().Name,
+                    StringComparison.Ordinal);
+            });
+            constAnimKeys.Sort((a, b) =>
+            {
+                int cmp = a!.Idx.CompareTo(b!.Idx);
+                if (cmp != 0) return cmp;
+            
+                cmp = a.Time.CompareTo(b.Time);
+                if (cmp != 0) return cmp;
+            
+                return string.Compare(
+                    a.GetType().Name,
+                    b.GetType().Name,
+                    StringComparison.Ordinal);
+            });
+    
             newCompressedBuffer = new animAnimationBufferCompressed()
             {
                 Duration = incomingAnimBufferData.Duration,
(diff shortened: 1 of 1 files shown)


WolvenKit.Modkit/RED4/Tools/Animation/AnimSpline.cs
```diff
@@ -228,6 +228,47 @@ internal static void EncodeAnimationData(out animAnimDataChunk newAnimDataChunk,
                 ((List<JointsScalesAtTimes>)[incomingAnimBufferData.Scales, incomingAnimBufferData.ConstScales]).All(_ =>
                     _.All(_ => _.Value.All(_ => EqWithEpsilon(_.Value, Scale1to1))));
```
gistya, line 238: Suggestion: rather than copy this same code three times below, make it a reusable function. Then add unit test for it. 

I'd also prefer to see us rely on something more robust than string comparison. String comparison depends on the naming being P < R < S alphabetically, but this could change and silently break this code. I'd suggest to use a simple enum switch like below so that the ordering is statically defined and can never break.

Ideally ComponentOf method could live inside `animAnimationBufferCompressed` and refactor it to use this also, but that's out of scope here. 

```suggestion
        // TODO: migrate this to animAnimationBufferCompressed and refactor to use it 
        public static ushort ComponentOf(animKey animKey) => animKey switch
        {
            animKeyPosition => 0,
            animKeyRotation => 1,
            animKeyScale => 2,
            _ => 0,
        };

        private static readonly Comparison<animKey?> AnimKeyBufferOrder = (a, b) =>
        {
              int cmp = a!.Idx.CompareTo(b!.Idx);
              if (cmp != 0) return cmp;
          
              cmp = a.Time.CompareTo(b.Time);
              if (cmp != 0) return cmp;
          
              return ComponentOf(a).CompareTo(ComponentOf(b));
         };
```

WolvenKit.Modkit/RED4/Tools/Animation/AnimSpline.cs
```diff
@@ -228,6 +228,47 @@ internal static void EncodeAnimationData(out animAnimDataChunk newAnimDataChunk,
...
+                if (cmp != 0) return cmp;
+            
+                cmp = a.Time.CompareTo(b.Time);
+                if (cmp != 0) return cmp;
+            
+                return string.Compare(
+                    a.GetType().Name,
+                    b.GetType().Name,
+                    StringComparison.Ordinal);
+            });
```
gistya:
```suggestion
            animKeys.Sort(AnimKeyBufferOrder);
            animKeysRaw.Sort(AnimKeyBufferOrder);
```

WolvenKit.Modkit/RED4/Tools/Animation/AnimSpline.cs
```diff
@@ -228,6 +228,47 @@ internal static void EncodeAnimationData(out animAnimDataChunk newAnimDataChunk,
...
```
gistya, line 264: Why are we sorting constAnimKeys here? Const keys are a different sub-buffer with a different record layout (WriteBuffer writes them idx-first, then time. See the `// NB order of idx, time is different to AnimKeys*` comments) and are documented in the source as "Single keyframe per joint index".

If every const key sits at t=0 (the normal case), sorting by Idx -> Time -> Type is the same as Idx -> Type, which is identical to what the current loop already emits. So this would be a no-op.

If a const key has a non-zero timestamp, the new sort interleaves const P/R/S by time, which is a behaviour change from the current code, but what is the rationale for changing that? (I understand the rationale for the first two but not for this one)

WolvenKit.Modkit/RED4/Tools/Animation/AnimSpline.cs
gistya: For these first two, can we separate out the duplicated logic into its own function?

WolvenKit.Modkit/RED4/Tools/Animation/AnimSpline.cs
```diff
@@ -228,6 +228,47 @@ internal static void EncodeAnimationData(out animAnimDataChunk newAnimDataChunk,
...
```
gistya:
```suggestion
```

Then here you don't need this second section.

gistya (commented):
PR looks good but have some suggestions and a question.

WolvenKit.Modkit/RED4/Tools/Animation/AnimSpline.cs
jono-loomy: great suggestions all round,  cheers.

Regarding ConstKeys, I included them in the sort just for the small handful of fringe cases where u have sets of 2 to 4 keys per bone/transform channel (sampled values at % duration). However, should rather be consolidating those instead.

PR #2945, opened 2026-06-06
Batch Loading + Expansion State Saving

gistya:
# Batch Loading + Expansion State Saving

For easy review, click these chunks. Note that most of the changes in the PR are in the tests.

- [refactor of WatcherService](https://github.com/WolvenKit/WolvenKit/pull/2945/changes/2b36eed76f6737fd5f55498559d7f4eab05d1084..1e564e2d0ebe5720fedfd49278e08b1b1bb44dbf)
- [refactor of ProjectExplorer](https://github.com/WolvenKit/WolvenKit/pull/2945/changes/8509d3dd4bb8d54933b464cbd1c95a70a9ebba83..1e89127a8d9fc3a57efe6b3517175a3a9d4542c2)
- [refactor of ProjectExplorerView, including addition of LiveSearch](https://github.com/WolvenKit/WolvenKit/pull/2945/changes/1e89127a8d9fc3a57efe6b3517175a3a9d4542c2..8d2c34d150c8a3ffa1c83c138d4ce9bcdb047271)
- [refactor of DispatcherHelper for better concurrency](https://github.com/WolvenKit/WolvenKit/pull/2945/changes/8d2c34d150c8a3ffa1c83c138d4ce9bcdb047271..3f243909abc960251bb6b9bdc2753eb05791475f)
- [various bugfixes and uptake of the refactor](https://github.com/WolvenKit/WolvenKit/pull/2945/changes/8d2c34d150c8a3ffa1c83c138d4ce9bcdb047271..3f243909abc960251bb6b9bdc2753eb05791475f) ••• [also here](https://github.com/WolvenKit/WolvenKit/pull/2945/changes/b6b1ff3185be5c5e95f27de4c3fb21b7fa410a7c..f40d61a177ac3eced6cb03fb25c60cce86e44123) ••• [and also here](https://github.com/WolvenKit/WolvenKit/pull/2945/changes/5bd5e046f41e1a951ebb7e5748c9f942d86b39fe..b021653141d6ecbbb941288594d28c3f186200eb)
- [refactor FileSystemModel to resolve observable property spam](https://github.com/WolvenKit/WolvenKit/pull/2945/changes/8992c86a9bb56e6ecb86363847fec95a444165b6)
- [add a Settings toggle for auto-expand folders on launch](https://github.com/WolvenKit/WolvenKit/pull/2945/changes/a9e5d1c821fcd4884a24b5fea6549c93967bb26f)
[...]

PR #3073, opened 2026-08-04
Improve Archive Manager Load Times

gistya:
# Improve Archive Manager Load Times

**Fixed:**
- Fixes an issue where the app would wait until the user opens a project before it starts loading the CP77 game archives, slowing down project loading.
- Now the app starts loading the CP77 archives in the background as soon as possible after launch. This dramatically speeds up being able to start working on your projects after a fresh launch of WolvenKit.

**Additional notes:**
- Will become part of #2945 once its merged to dev and verified
- Works best once #3071 is also merged. 
- Resolves the below longstanding TODO from @rfuzzo by relocating the LoadCustomHashes method to the new class ArchiveManagerLoader:

https://github.com/WolvenKit/WolvenKit/blob/eadb315540ea4e70b9b791b33922aaa0c5e73086/WolvenKit.App/Controllers/RED4Controller.cs#L83-L87

### DIFF
--- a/WolvenKit.App/Services/AppArchiveManager.cs
+++ b/WolvenKit.App/Services/AppArchiveManager.cs
@@ -32,14 +32,14 @@ public class AppArchiveManager : ArchiveManager, IAppArchiveManager
         _progressService = progressService;
         _settings = settings;
     }
-    
+
     #region Fields
 
     private readonly IHashService _hashService;
     private readonly ILoggerService _logger;
     private readonly IProgressService<double> _progressService;
     private readonly ISettingsManager _settings;
-    
+
     private readonly SourceList<RedFileSystemModel> _rootCache = new();
 
     private readonly SourceList<RedFileSystemModel> _modCache = new();
@@ -218,11 +218,17 @@ public class AppArchiveManager : ArchiveManager, IAppArchiveManager
 
         var progress = -1;
         var numTotalArchives = (float)GetModArchives().Count();
-        
+        var reportEvery = Math.Max(1, (int)(numTotalArchives / 100));
+
         foreach (var archive in GetModArchives())
         {
             progress++;
-            _progressService.Report(progress / numTotalArchives);
+
+            if (progress % reportEvery == 0)
+            {
+                _progressService.Report(progress / numTotalArchives);
+            }
+
             ArgumentNullException.ThrowIfNull(archive.ArchiveRelativePath,
                 $"{nameof(archive.ArchiveRelativePath)}, archive name: ${archive.Name}");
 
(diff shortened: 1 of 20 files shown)


seberoth:
> Fixes an issue where the app would wait until the user opens a project before it starts loading the CP77 game archives

That isn't an issue. This was wanted in the past.
You can open wkit without a project and it isn't always needed / wanted that the archives gets loaded then.
Thats why we have the "Load Archives" button in the AssetBrowser

manavortex:
If loading is fully asynchronous, I don't see a harm in it. People opening WKit without a project are usually clueless newbies who have simply forgotten to make one (I still want that "you have no project, create one?" overlay in the project explorer, akin to how the asset browser behaves).

seberoth:
I actually like it how it is. That I'm not forced to load the game archives.
I use it quite often, but if you (the active team) feel this does more harm then it helps feel free to change that

manavortex:
OK, I didn't know that you had a preference for it to stay this way, that does change things! I definitely want to streamline UX for the majority of users, but I also _definitely_ want to accommodate power users where possible. 

IMO it makes sense to decouple these loading tasks with "I want to use these right now", since directly after startup, the app is probably just sitting there while the user clicks on the project they want to load. (Same for Tweak Browser, possibly).

What would be your preferred outcome?
- There's a setting to opt out of this change and preserve the old behaviour
- Archive loading is so async that you don't notice because it happens in the background
- Something else entirely

seberoth:
Settings sounds good to me (for my use case)

Async loading wouldn't help me here, since its mostly about having dev environments where I just have the exe/oodle dll without any game archives^^

gistya:
> Async loading wouldn't help me here, since its mostly about having dev environments where I just have the exe/oodle dll without any game archives^^

Well certainly in my current PR, if the archives don't exist, then the archives simply won't load, and it will just behave the same as today, showing the Load Archives button.

Can you try the branch and make sure if it works for your use case?

manavortex:
I'd be in favour of the settings toggle - but make it opt-out (as in, the new behaviour should be default). People who know what they're doing will only have to set it once, and everyone else gets to benefit :)

gistya:
> I'd be in favour of the settings toggle - but make it opt-out (as in, the new behaviour should be default). People who know what they're doing will only have to set it once, and everyone else gets to benefit :) 

Supporting two behaviors is more fragile. We could do it but it means double the tests. 

If Sebs only concern is "I want to be able to use the app in a dev environment where there's no game install" then this PR does not impact his use case at all. Because if the archives are not present, then the "Load Archives" button will show and the app will behave identical to today. 

All this PR does is move back the point in time where automatic loading occurs, to when MainView loads, instead of waiting until you open a project. Because right now it means you wait an extra 20 seconds or so, and that slows down all the integration and UI tests.

WolvenKit.App/Services/AppArchiveManager.cs
```diff
@@ -32,14 +32,14 @@ public AppArchiveManager(
         _progressService = progressService;
         _settings = settings;
     }
-    
+
```
gistya, line 35: I can revert the whitespace changes before merge.

gistya:
this was merged into "the big pr" #2945

PR #3075, opened 2026-08-04
Quest graph lazy loading

gistya:
# Implements lazy loading of quest graphs for better user experience

**Implemented:**
- Graphs now load in chunks of several nodes at a time without blocking the UI thread. There existing animated loading indicator now works as expected and the progress indicator updates to accurately represent progress as the graph loads. 

**Fixed:**
- Fixes an issue where opening quest files would freeze the UI without a progress bar while they loaded and would appear all jumbled until layout completed.
- Fixes an issue where opening a document would not save the open documents list right away

### DIFF
--- a/WolvenKit.App/ViewModels/GraphEditor/RedGraph.Quest.cs
+++ b/WolvenKit.App/ViewModels/GraphEditor/RedGraph.Quest.cs
@@ -1558,6 +1558,9 @@ public partial class RedGraph
             }
         }
 
+        var inputs = graph.Nodes.Sum(n => n.Input.Count);
+        var outputs = graph.Nodes.Sum(n => n.Output.Count);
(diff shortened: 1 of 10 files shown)


misterchedda:
nice work! It seems quite smooth on testing on a few big questphase and scene files

However I see that only `QuestPhaseGraphView` subscribes to the new `CanvasRealized` but we'd want all graph consumers to do so too like .scene's `SceneGraphView` and .behavior `BehaviorGraphView`) too ideally since they are affected by this change so that users get the same loading indicator everywhere. Also - within questphase, inline phase navigation and breadcrumb/back navigation would also need it

gistya:
> but we'd want all graph consumers to do so too

i'm not familiar with all the other places. can you make a patch that implements it in the other spots or send me a list of the filenames that also need migration? thanks

PR #3077, opened 2026-08-05
WelcomePageView UX improvments for scrolling

gistya:
# WelcomePageView UX improvments for scrolling

**Fixed:**
- Fixed an issue where mousewheel scrolling would not work with the cursor inside the pinned projects area. Now you can see more pinned projects and scroll all projects independently of the rest of the welcome screen.


notaspirit (changes requested):
The first column images are clipping a bit, and for the dark accent the corner radius doesn't match the other elements. 
(The clipping has been that way in main too, so you don't have to fix it as part of this pr) 

Also I would prefer a lighter color instead for that bar that matches the sidebar, the current dark tone stands out a lot.

The recent and pinned project bars don't quite line up

Image

Image

notaspirit (approved):

PR #3081, opened 2026-08-05
Fix ChunkVM: to properly handleTweakDB  exceptions

gistya:
# Fix ChunkVM: to properly handleTweakDB  exceptions

**Fixed:**
- Fixes an issue where ChunkViewModel does not handle exceptions thrown by `GetRedTypeFromCSType` -- currently it just dumps a stacktrace into the app logs. 
- Now we'll catch the exception and run `_loggerService.Error` for a more sane output.

**Additional notes:**
closes_issue_number

### DIFF
--- a/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
@@ -804,8 +804,16 @@ public partial class ChunkViewModel : ObservableObject, ISelectableTreeViewItemM
     {
         get
         {
-            var redName = GetRedTypeFromCSType(PropertyType, Flags);
-            return redName != "" ? redName : PropertyType.Name;
+            try
+            {
+                var redName = GetRedTypeFromCSType(PropertyType, Flags);
+                return redName != "" ? redName : PropertyType.Name;
+            }
+            catch (WolvenKitException ex)
+            {
+                _loggerService.Error($"File type could not be determined for a Tweak. Exception: ${ex.Message}");
+                return "";
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


notaspirit (approved):

notaspirit (changes requested):
actually hold on wouldn't it make more sense to return PropertyType.Name?

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
```diff
@@ -804,8 +804,26 @@ public string Type
...
+                    _loggerService.Error(
+                        $"File type could not be determined for a Tweak with name: {PropertyType.Name}. Exception: {ex.Message}.");
+                }
+                catch (Exception ex2)
+                {
+                    _loggerService.Error(
+                        $"File type could not be determined for a Tweak whose name could not be determined. Exceptions: {Environment.NewLine}1. {ex.Message}{Environment.NewLine}2, {ex2.Message}.");
+                }
+
+                return $"{PropertyType.Name}";
```
notaspirit: I don't understand the $"{}" for a value that is already a string

notaspirit (changes requested):

seberoth:
Would rename the message to something like: `RedType could not be determined for the property...`
Since it neither has to do with file types nor it is tied to tweak related stuff.

WolvenKit.App/ViewModels/Shell/ChunkViewModel.cs
gistya: Sorry I did not realize it was a string lol will fix
```suggestion
                return PropertyType.Name;
```

notaspirit (approved):

PR #3087, opened 2026-08-06
Vehicle Damage Model

seberoth:
# Vehicle Damage Model

**Implemented:**
- Vehicle Damage Model for mesh export / import

Thought after 4 years "working" on it, it should finally end up in WKit...
Preview disabled for now, since its not very usable. If someone wants to tackle that...


seberoth:
Didn't do much testing (today). Just brought back my old code and tested:
- Import/Export without blender (looked fine...)
- Import glb in Blender, looked fine

File used `base\vehicles\sport\v_sport1_rayfield_caliburn\entities\meshes\v_sport1_rayfield_caliburn__ext01_body_01.mesh`

gistya (approved):
LGTM, maybe add tests? All I can think of

PR #3093, opened 2026-08-08
fix search: restore forced property initialization on shift

manavortex:
# fix search: restore forced property initialization on shift

image


WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.UserInteractionStates.cs
```diff
@@ -54,27 +54,39 @@ public void SetEditorLevel(EditorDifficultyLevel level)
 
     private bool _isPropertiesInitialized;
 
+    ///
```
gistya:
```suggestion
```

gistya (dismissed):
Looks good to me. just one minor change suggested, not blocking tho

gistya (approved):

PR #3103, opened 2026-08-09
Reduce OnPropertyChanged spam from FileSystemModel

gistya:
# Reduce OnPropertyChanged spam from FileSystemModel

**Fixed:**
- Fixes a performance issue related to excessive OnPropertyChange notifications generated by FileSystemModel

**Additional notes:**
- Tests added in stacked PR #3102 guarantee that this change does not introduce regressions, and that the refactored FileSystemModel behaves identically to the prior one.

### DIFF
--- a/WolvenKit.App/Models/FileSystemModel.cs
+++ b/WolvenKit.App/Models/FileSystemModel.cs
@@ -4,8 +4,11 @@ using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Runtime.CompilerServices;
+using Microsoft.EntityFrameworkCore.Query.Internal;
 using WolvenKit.Common;
+using WolvenKit.Core.Extensions;
 using WolvenKit.RED4.Types;
 
 namespace WolvenKit.App.Models;
@@ -14,14 +17,11 @@ public class FileSystemModel : INotifyPropertyChanged
 {
     public const string ProjectDirName = "<ProjectDir>";
 
+    private readonly string _projectDirectory;
     private string _name;
-    private string _gameRelativePath = null!;
-    private long _fileSize;
-    private string _fileSizeStr = null!;
-    private string _extension = "default";
 
     [Browsable(false)] public FileSystemModel? Parent { get; }
-    
+
     public string Name
     {
         get => _name;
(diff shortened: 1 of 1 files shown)


seberoth (changes requested):
`GameRelativePath` (and Hash) should show what the file path in the archives would be.
Therefor it should always use `ResourcePath.DirectorySeparatorChar` as directory separator for all operating systems.

gistya:
> `GameRelativePath` (and Hash) should show what the file path in the archives would be. Therefor it should always use `ResourcePath.DirectorySeparatorChar` as directory separator for all operating systems.

Fixed now

https://github.com/WolvenKit/WolvenKit/pull/3103/changes#diff-f80fabe38a981f4ea9365f1f1fa0a3c3c3cbd1a72a160d191670ad0594ce4d12R211-R215

WolvenKit.App/Models/FileSystemModel.cs
```diff
@@ -110,63 +160,59 @@ public void UpdateFileInfo()
...
+                    case Constants.RawDirectoryTop:
+                        GameRelativePath = RawRelativePath[4..];
+                        break;
+                    case Constants.ResourceDirectoryTop:
+                        GameRelativePath = RawRelativePath[10..];
+                        break;
+                    default:
+                        var split =
+                            RawRelativePath.Split(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)[1..];
+                        GameRelativePath = Path.Combine(split);
```
seberoth: Not relevant for hashes, since their aren't hashes for dir, but probably to use `ResourcePath.DirectorySeparatorChar` here too

gistya: Actually if we need `\\` used in GameRelativePath what we really want is:

```
var split = RawRelativePath
    .Split(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)[1..];

GameRelativePath = string.Join(ResourcePath.DirectorySeparatorChar, split);
```

RawRelativePath uses `/` on Linux so if we switched to split on `\\` instead it would break. But this should work

seberoth: Yeah, sorry. Didn't want to suggest to change the `RawRelativePath.Split`
Was just meant for the `Path.Combine(...`

seberoth (approved):

PR #3106, opened 2026-08-09
Improve Project Loading Speed

gistya:
# Improve Project Loading Speed

**Implemented:**
- Vastly improved project loading speed
- Smooth fade-out animation from Home Screen to Docking View

**Fixes**
- Fixes an issue where BeginDeferredRefreshContext would not await its action like it was supposed to. This was hidden due to using a cancellation token, but now we don't have to. It was fixed by changing the signature to `Func<Func<Task>, Task>? BeginDeferredRefreshContext`. This has been fixed on dev for awhile.

**Additional notes:**
- This PR is a reworking of the project loading speed improvements from #2945.
- It relies upon silencing collection change spam from the FileList thanks to a tip from Sebroth (ObservableCollectionEx).
- It retains existing WatcherService behavior and ProjectExplorer behavior, so there should not be any regressions such as we kept experiencing with the #2945 PR. 
- Other aspects of the #2945 PR like LiveSearch and improved loading indicators will come in subsequent stacked PRs.

## Key things to note here for reviewers

- Most of the PR is tests. To view the non-test changes only, click [here](https://github.com/WolvenKit/WolvenKit/pull/3106/changes/7796ae15c6b0a1be1e982494d41554fd5a1450dc).
- Switching to using `SfDataGrid` instead of `SfTreeGrid` for the "flat" view of ProjectExplorer resolves the SyncFusion "object does not refer to an instance of an object" stacktraces that testers kept running into while testing #2945. 
- Switching to using a subclass of ObservableCollectionEx (`DispatchedObservableCollectionEx`) allows for batch adding of large numbers of files with great performance without having to rely on the `AddRange` command. The reason is because ObservableCollectionEx lets us tell the `FileList` not to publish events while we're adding large numbers of files. 
- Using `Add` instead of `AddRange` keeps SfTreeGrid happy. SfTreeGrid works best when files are added one-by-one. This change seems to result in not having to use a DeferredRefresh cycle for so many other file operations in the app, letting us simply rely on the existing FileWatcher for most of those without regressions. 
- Preserving more of the original behavior of the app keeps the changes to a minimum and reduces risks of regressions. 
- The changes to `AppViewModel` have been on `dev` for quite awhile. (It just defers starting project load until after the fade-out completes in order that fade-out will render more smoothly).


WolvenKit.App/Services/WatcherService.cs
```diff
@@ -410,20 +447,59 @@ private void InternalRefresh()
...
+        {
+            foreach (var fileSystemInfo in allFiles)
+            {
+                var name = fileSystemInfo.FullName[(_projectDirectory.Length + 1)..];
+                _fileChanges.Enqueue(new FileSystemEventArgsWrapper(
+                    new FileSystemEventArgs(WatcherChangeTypes.Created, _projectDirectory, name)));
+            }
+
+            return;
+        }
```
gistya: If the project's file count is less than 250, load it the old fashioned way.

WolvenKit.App/Services/WatcherService.cs
```diff
@@ -410,20 +447,59 @@ private void InternalRefresh()
...
+        FileList.SuppressNotification = true;
+        WeakReferenceMessenger.Default.Send(new ChalkboardService.WillStartLoadingProjectFiles());
+
+        CreateBaseItems(allFiles);
+
+        Parallel.ForEach(allFiles, _parallelOptions, fileSystemInfo =>
+            CreateInstantly(fileSystemInfo.FullName[(_projectDirectory.Length + 1)..]));
+
+        FileList.SuppressNotification = false;
+        WeakReferenceMessenger.Default.Send(new ChalkboardService.DidFinishLoadingProjectFiles());
```
gistya: If the file count is 250 or greater, take measures to speed up loading. The effect of this is similar to the DeferredRefresh logic but doesn't rely on a delegate pattern.

WolvenKit.App/Services/WatcherService.cs
```diff
@@ -410,20 +447,59 @@ private void InternalRefresh()
...
+        foreach (var fileInfo in allFiles)
+        {
+            var rawRelPath = fileInfo.FullName[(_projectDirectory.Length + 1)..];
+
+            if (!rawRelPath.Contains(Path.DirectorySeparatorChar))
+            {
+                CreateInstantly(rawRelPath);
+            }
+        }
+    }
```
gistya: Create the root-level items for Fie Tree in order first so that all the rest of the file additions are guaranteed to have their parent present. (Parallel.Foreach can be nondeterministic.)

WolvenKit.App/Services/WatcherService.cs
```diff
@@ -410,20 +447,59 @@ private void InternalRefresh()
...
+
+    public bool? GetExpansionStateOrNull(string relPath) => ExpansionStateDictionary.TryGetValue(relPath, out var state) ? state : null;
```
gistya, line 522: Moved expansion state to WatcherService since it's now tracked through FileSystemModel's `isExpanded` state, which is the best practice way of doing it in SyncFusion.

WolvenKit.App/Services/WatcherService.cs
```diff
@@ -44,6 +49,13 @@ public partial class WatcherService : ObservableObject, IWatcherService
     [ObservableProperty]
     private DispatchedObservableCollection<FileSystemModel> _fileTree = new();
 
+    private static readonly int _maxDoP = Environment.ProcessorCount > 1 ? Environment.ProcessorCount : 1;
```
manavortex, line 52: Math.Max?

manavortex (approved):

Issue #3110, opened 2026-08-10
Drag and drop does not work properly in TreeView of ProjectExplorer


gistya:
**Describe the bug**

Drag and Drop doesn't work if you drag a single file that was not already selected.

**To Reproduce**
1. Go to ProjectExplorer.
2. Select a folder.
3. Click to drag and drop a file without selecting it first.
4. Observe that nothing happens

**Expected behavior**
The file should become selected and then get properly moved/copied.

**Version (please complete the following information):**
 - Version 8.20.0

gistya:
Resolved by #3111

PR #3111, opened 2026-08-10
Fix unselected single-item drag-and-drop

gistya:
# Fix unselected single-item drag-and-drop

**Fixed:**
- Fixes an issue where dragging and dropping a single item would fail if it was not already selected.

**Additional notes:**
Closes issue #3110

### DIFF
--- a/WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
+++ b/WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
@@ -745,8 +745,41 @@ namespace WolvenKit.Views.Tools
             }
         }
 
-        private void RowDragDropController_DragStart(object sender, TreeGridRowDragStartEventArgs e) =>
+        private void RowDragDropController_DragStart(object sender, TreeGridRowDragStartEventArgs e)
+        {
             _isDragging = true;
+
+            if (ViewModel is not { } vm)
+            {
+                return;
+            }
+
+            vm.IsDragging = true;
+
+            var draggingItems = e.DraggingNodes.Select(x => x.Item as FileSystemModel).ToList();
+
+            if (vm.SelectedItems is { } selectedItems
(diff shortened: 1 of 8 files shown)


WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
```diff
@@ -745,9 +745,42 @@ private void CollapseAll_OnClick(object sender, RoutedEventArgs e)
             }
         }
 
-        private void RowDragDropController_DragStart(object sender, TreeGridRowDragStartEventArgs e) =>
+        private void RowDragDropController_DragStart(object sender, TreeGridRowDragStartEventArgs e)
+        {
             _isDragging = true;
```
manavortex, line 750: should starting another cancel the ongoing drag event? or maybe not start another?

manavortex (dismissed):

WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
gistya: I don't think it's possible to start another one before the first one stops. Unless you have two cursors and two mice?

manavortex: Lag, I have lag :P

manavortex (dismissed):

WolvenKit/Views/Tools/ProjectExplorerView.xaml.cs
gistya: My Pr doesn't change anything about the existing way that double-events like that would be handled. If it's a reproducible issue let me know how to repro it so I can make a separate PR to address it? (Or if it's an easy fix I can just fix it here, but i think i'd want to add new tests etc)

notaspirit (approved):

PR #3122, opened 2026-08-13
Project loading indicator improvements

gistya:
# Project loading indicator improvements

**Implemented:**
- Project loading indicator now shows earlier as soon as a project starts loading for better UX, and during long batch operations when one should not interact with the file tree lest bad things happen. 
- If project loading fails or is somehow canceled, the app now gracefully handles this, preventing the Loading indicator from being able to get stuck.

**Additional Notes**
- broken out from PR #2945
- resolves issue #3126

### DIFF
--- a/WolvenKit.App/ViewModels/HomePage/Pages/WelcomePageViewModel.cs
+++ b/WolvenKit.App/ViewModels/HomePage/Pages/WelcomePageViewModel.cs
@@ -17,6 +17,7 @@ using WolvenKit.App.Interaction;
 using WolvenKit.App.Models.ProjectManagement;
 using WolvenKit.App.Services;
 using WolvenKit.App.ViewModels.Shell;
+using WolvenKit.App.ViewModels.Tools;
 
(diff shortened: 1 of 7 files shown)


manavortex:
Could you make the project tree display an overlay when no project is loaded, similar to the asset browser? We've talked about it, this PR might be a good place to do it

gistya:
> Could you make the project tree display an overlay when no project is loaded, similar to the asset browser? We've talked about it, this PR might be a good place to do it

I agree that would be nice. But I don't want to expand the scope of these PRs as it will cause merge conflicts with the remaining merge-ins from the big PR that I'm trying to dice up into digestible reviewable chunks. 

So can you please file a GitHub issue to request that as a new feature and I will make a separate PR for it after the big PR's stuff is all fully merged in?

manavortex:
Sure thing - I lost track of that one, thought it was merged already :) I think we have an issue already. If not, I'll try to make one tomorrow (today is almost over)

manavortex (approved):

Issue #3139, opened 2026-08-14
Make WatcherService part of ProjectExplorer for testability and SRP


gistya:
# The problem

Currently, WatcherService owns the FileList and FileTree, which are the data models that back the ProjectExplorerView. 

This is an architectural no-no in MVVM. The ViewModel is supposed to own the source of truth for the View. Its whole job is to own the *models*. The models should not be owned by a service that can be interacted with freely by any part of the app. 

It also violates Single Responsibility Principle because the WatcherService (if it is to exist at all) should only be responsible for watching the file system, not owning the data model. Since its only role relates to ProjectExplorer, arguably it should not be a public service open to the rest of the app for interaction.

Further, there is no abstract API for adding or removing files from a project, that could be used in tests to simulate project behavior. This makes writing tests for the ProjectExplorer problematic. 

# Proposed solution

Steps:

1. ~~Rename `WolvenKit.App/Services/WatcherService.cs` to `WolvenKit.App/ViewModels/Tools/ProjectExplorerViewModel+Watcher.cs` using `git mv` to retain history.~~ done in #3140
2. ~~Make WatcherService become part of ProjectExplorerViewModel's partial class.~~ done in #3140
3. ~~Make a new ProjectEvents messaging interface that allows for publishing and subscribing to various project-related events such as adding files, removing files, moving files, renaming files, opening/closing a project, etc.~~ done in #3141 and #3144
4. Replace dependencies on WatcherService and ProjectExplorerViewModel with dependencies upon the ProjectEvents abstraction.
5. ~~Implement a main thread serial queue that ensures all messaging on ProjectEvents occurs in a serial fashion on stack frame at a time in order to protect against threading issues and race conditions.~~ done in #3141 (and supplemented in #3142)
6. Migrate usages in the app of "Suspend" / "Resume" on ProjectWatcher with calls that go through ProjectEvents.
7. Cover the whole thing in copious tests. partially done in #3145

One benefit of messaging services like ProjectEvents or ChalkboardService is that tests can also monitor these events to ensure that things are happening in the right sequence and the right input/outputs are being given.

notaspirit:
What is the breaking change that would require it to be part of 9.0.0?

Or is this more of a "planned to be done by" usage of the milestone?

gistya:
"Major improvements to UX that are planned to be part of the next major release. Includes minor behavioral changes that may alter your workflow."

[...]

PR #3140, opened 2026-08-14
Subsume WatcherService Under ProjectExplorer

gistya:
# Subsume WatcherService Under ProjectExplorer

This PR has two commits. 

In the first one, we move rename WatcherService.cs to ProjectExplorerViewModel+Watcher.cs via:

`git mv .\WolvenKit.App\Services\WatcherService.cs WolvenKit.App\ViewModels\Tools\ProjectExplorerViewModel+Watcher.cs`

Using `git mv` ensures that git history for the file is retained across the move.

In the second commit, we refactor the code so that as much of the old WatcherService as possible becomes `private` without changing any existing code paths or behaviors. 

This change allows ProjectExplorerViewModel to own its own models (FileTree, FileList), as is best practice in MVVM. 

This work is broken out from PR #2945 and is preliminary to merging in more changes from that PR in smaller chunks, particularly the big new test suite.

**Implemented:**
- Refactor of WatcherService to be part of ProjectExplorerViewModel.

**Additional notes:**
Part of the work for issue #3139


WolvenKit.App/ViewModels/Shell/AppViewModel.cs
```diff
@@ -1530,15 +1527,15 @@ private void AddAxlFiles()
             return;
         }
 
-        _watcherService.Suspend();
+        GetToolViewModel<ProjectExplorerViewModel>().Suspend();
```
notaspirit: wouldn't it be cleaner if it got it once into a local variable so it's not refetching it for the later calls?

notaspirit (approved):
just the one minor comment

WolvenKit.App/ViewModels/Shell/AppViewModel.cs
gistya: Will implement in next push.

PR #3141, opened 2026-08-14
Add ProjectEvents Messaging Service & ProjectEventsTests

gistya:
# Add ProjectEvents Messaging Service

This PR adds the ProjectEvents messaging service and associated unit tests but does not wire up ProjectEvents into anything in the live app yet. (That will be a subsequent PR that will be stacked on this one.)

We also add a MainThreadInlineScheduler that avoids race conditions with SyncFusion model updates by ensuring that data model mutations happen before SyncFusion refresh events and not during them.

At this stage, ProjectEvents is not wired up to anything in the app yet, so this PR should be an easy approve. 

ProjectEvents was on dev for a couple of weeks and was previously reviewed, so I'm confident this is solid.

**Implemented:**
- Adds the ProjectEvents messaging service from PR #2945.

**Additional notes:**
Part of the work for implementing better testability for file and project operations #3139. This is a prerequisite for a lot of tests to come.
and breaking apart PR #2945 into chunks

### DIFF
--- a/WolvenKit.App/Services/IProjectEvents.cs
+++ b/WolvenKit.App/Services/IProjectEvents.cs
@@ -1,0 +1,142 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using WolvenKit.App.ViewModels.Documents;
+using WolvenKit.Core.Interfaces;
+
+namespace WolvenKit.App.Services;
+
+/// <summary>
+/// Service for published events related to file operations.
+/// Provides an alternative to monitoring of file system events.
+/// </summary>
+public interface IProjectEvents
+{
+    /// <summary>
+    /// Consumers of the published events should subscribe here.
+    /// </summary>
+    IObservable<FilesImportedMessage> FilesImported { get; }
+
+    /// <summary>
+    /// Consumers of completed-move events should subscribe here.
+    /// </summary>
+    IObservable<FilesMovedMessage> FilesMoved { get; }
+
+    /// <summary>
+    /// Consumers of completed-deletion events should subscribe here.
+    /// </summary>
+    IObservable<FilesDeletedMessage> FilesDeleted { get; }
+
+    /// <summary>
+    /// Consumers of completed change events should subscribe here.
+    /// </summary>
+    IObservable<FileChangedMessage> FileChanged { get; }
+
+    /// <summary>
+    /// To be called when a known batch of files are being added to the project.
+    /// E.g.: when adding files from AssetBrowser or exporting JSON, etc.
+    /// NOTE: Be sure to disable monitoring of file system events before use.
+    /// </summary>
+    /// <param name="message"></param>
+    void PublishFilesImported(FilesImportedMessage message);
+
+    /// <summary>
+    /// Publish that a single file was added to the project at <paramref name="absolutePath"/> (e.g.
+    /// copied in from outside the project). Call this AFTER the file exists on disk. The project
+    /// explorer materializes the node and any missing parent directories. Idempotent.
+    /// </summary>
+    /// <param name="absolutePath">Absolute path of the added file.</param>
+    void PublishFileImported(string absolutePath);
+
+    /// <summary>
+    /// Publish an authoritative set of file moves that have <b>already happened on disk</b>.
+    /// Call this AFTER the move completes and after any overwrite/confirmation prompts have been
+    /// resolved, so the project explorer reflects what actually happened rather than what was
+    /// intended. The apply is idempotent, so it is safe to publish whether or not the OS file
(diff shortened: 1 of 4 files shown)


WolvenKit.App/Services/MainThreadInlineScheduler.cs
```diff
@@ -0,0 +1,65 @@
...
+/// <summary>
+/// An <see cref="IScheduler"/> that runs work <b>inline</b> (synchronously) when the caller is already on
+/// the WPF UI thread, and otherwise defers to a <paramref name="fallback"/> scheduler (normally
+/// <c>RxApp.MainThreadScheduler</c>, which posts onto the dispatcher).
+///
+/// <para>
+/// Why this exists: <c>ObserveOn(RxApp.MainThreadScheduler)</c> ALWAYS re-posts each notification to a
+/// later dispatcher turn — even for a publisher that is already on the UI thread. For the project-event
+/// pipeline that means a UI-thread flow which publishes and then <i>immediately</i> releases a Syncfusion
+/// deferred-refresh (e.g. renaming a file) can redraw the grid from domain models the WatcherService
```
notaspirit: mentions watcherservice which has been refactored away downstack

WolvenKit.App/Services/IProjectEvents.cs
```diff
@@ -0,0 +1,142 @@
...
+/// A message published when a file has changed, i.e. when it has been saved or modified
+/// without the filename changing.
+/// </summary>
+public abstract record FileChangedMessage
+{
+    /// <summary>
+    /// A published IFileSystemViewModel that has been changed without being renamed.
+    /// </summary>
+    /// <param name="File"></param>
+    public sealed record Document(IDocumentViewModel File): FileChangedMessage;
```
notaspirit, line 119: Is it necessary to tie the file changed message to an open document? since files can also be changed externally

WolvenKit.App/Services/MainThreadInlineScheduler.cs
gistya: Thanks. I'll also push a global search and replace for appearances of "WatcherService" in the comments, there were several others. Replaced them all with "ProjectExplorerViewModel". I'm going to make this change in the "watcher_service_rename" PR however it'll propagate up into this one when I rebase the stack.

WolvenKit.App/Services/IProjectEvents.cs
gistya: > Is it necessary to tie the file changed message to an open document?

Yes, this is specifically to support a future PR in which we can refresh open documents when they change. One issue that I discovered when working on the "big PR" was that sometimes open documents could become stale, for example if you move and rename the file, then the open document now refers to a file that no longer exists on disk and so it needs to be refreshed to refer to the new file's location.

WolvenKit.App/Services/MainThreadInlineScheduler.cs
gistya:
```suggestion
/// deferred-refresh (e.g. renaming a file) can redraw the grid from domain models the ProjectExplorerViewModel
```

WolvenKit.App/Services/IProjectEvents.cs
notaspirit: Could it expose the file path directly and the document only if one is open? As it is now changes for files which aren't open would never get an event as far as I understand.

gistya: Well OK so the reason for using `abstract record` for the wrapper type `FileChangedMessage` is that anytime we need to add support for a new kind of data type (be it a string path, or a `FileSystemModel`, or `IGameFile`, etc.) then we can simply add another nested `sealed record` type. 

Right now I only have the types of messages that have actually been implemented in the big PR branch and in the tests. But we are not limited to just supporting those types as it is trivial to add new ones when the need should arise, easily enough:

```csharp
public abstract record FileChangedMessage
{
    public sealed record Document(IDocumentViewModel File): FileChangedMessage;
    /// New type:
    public sealed record ExternallyChangedFile(string FullPath): FileChangedMessage;
}
```

That's the advantage of using `abstract record` instead of say, `enum` or `sealed record` as the parent type that subscribers listen for. Another advantage is that it lets us use strongly-typed message payloads, that way subscribers can be specialized to a particular subtype of that message (for example, AppViewModel listening for Document changes to take action on open documents). 

But yeah we're totally not limited to just that. It's outside the scope of this PR to add in more types than we currently actually use but we're also not limiting ourselves with the architecture here. It's built to be extensible.

notaspirit: Makes sense

notaspirit (approved):

PR #3148, opened 2026-08-16
Fix opus export skipping the last pak and mangling wav paths

Zhincore:
# Fixed opus export skipping the last pak and mangling wav paths

**Implemented:**
- Unit tests for `OpusTools.ExportAllOpus` covering pak enumeration, unsorted `PackIndices`, and progress reporting — no game install or fixture files needed

**Fixed:**
- Exporting all sounds from OpusPaks never opened the highest-numbered pak, so those sounds could not be extracted at all. On current game files that silently dropped the 27 sounds in `sfx_container_1878.opuspak`. The bound also no longer assumes `PackIndices` is sorted, and progress now reaches 1.0.
- Opus export and import built the `.wav` path with `Replace("opus", "wav")` on the full path, which rewrites the directory part too — output was lost whenever the directory contained "opus" in its name. Now uses `Path.ChangeExtension`.

**Additional notes:**
No API changes, both are behavioural fixes. Verified the new tests fail against the previous implementation and pass with the fix.

Full unit suite is 488/489; the one failure (`SceneDescriptorTests.AnimTargetDescribesStopAndStaticPosition`) is pre-existing on clean main and locale-dependent — it asserts on `1.25` but a comma-decimal locale renders `1,25`.

`

### DIFF
--- a/Tests/WolvenKit.UnitTests/Modkit/OpusToolsTests.cs
+++ b/Tests/WolvenKit.UnitTests/Modkit/OpusToolsTests.cs
@@ -1,0 +1,112 @@
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using WolvenKit.Common;
+using WolvenKit.Modkit.RED4.Opus;
+using WolvenKit.RED4.Types;
+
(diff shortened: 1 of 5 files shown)


notaspirit (approved):

notaspirit:
locale dependent serialization my beloved

Issue #3150, opened 2026-08-20
Moving an open document's file and then saving the file creates a new copy of the file


gistya:
**Describe the bug**
Moving an open document's file and then saving the file creates a new copy of the file. This happens because of insufficient tracking of file changes. 

**To Reproduce**
1. Go to a project and open a file (just some CSV, doesn't matter).
2. In ProjectExporer tree view, click & drag to move the file to a different folder.
3. Press the save button at the top of the screen.
4. Observe that a new file gets created where the old one used to be, rather than the moved file being saved.

**Expected behavior**
The app should keep track of where the file has moved to and save it there. Further if there would be any loss of tracking then the user should be notified somehow, at least in Warning logs if not via a dialog box, or the file move/rename/delete should be disallowed.

**Version (please complete the following information):**
 - Version 8.20.0, 8.20.1 nightly (as of 8/20/26, coincidentally)

PR #3151, opened 2026-08-20
Fix issues with reloading open docs after changes

gistya:
# Fix issues with reloading open docs after changes

**Fixed:**
- Fixed a bug where moving/renaming an open document and then resaving that document could create an unintended extra copy at the pre-move/pre-rename path.

**Additional notes:**
Closes issue #3150.

### DIFF
--- a/WolvenKit.App/ViewModels/Shell/AppViewModel.cs
+++ b/WolvenKit.App/ViewModels/Shell/AppViewModel.cs
@@ -444,7 +444,7 @@ public partial class AppViewModel : ObservableObject/*, IAppViewModel*/
         string? projectPathToOpen = null;
 
         // Will be overwritten if the launch args contain a project path
-        if (args.Contains("-reopenProject") && (SettingsManager.LastUsedProjectPath is string projectPath) && SettingsManager.ReopenLastProject)
+        if (args.Contains("-reopenProject") && (SettingsManager.LastUsedProjectPath is string projectPath) || SettingsManager.ReopenLastProject)
         {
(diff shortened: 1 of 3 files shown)


WolvenKit.App/ViewModels/Shell/AppViewModel.cs
```diff
@@ -442,7 +444,7 @@ private bool OpenFileFromLaunchArgs()
         string? projectPathToOpen = null;
 
         // Will be overwritten if the launch args contain a project path
-        if (args.Contains("-reopenProject") && SettingsManager.LastUsedProjectPath is string projectPath)
+        if (args.Contains("-reopenProject") && (SettingsManager.LastUsedProjectPath is string projectPath) && SettingsManager.ReopenLastProject)
```
manavortex:
```suggestion
        if (args.Contains("-reopenProject") && (SettingsManager.LastUsedProjectPath is string projectPath) || SettingsManager.ReopenLastProject)
```

manavortex: I haven't configured the setting, but I use the command line argument when I run from source

manavortex (approved):

WolvenKit.App/ViewModels/Shell/AppViewModel.cs
gistya: Actually on second thought. Might as well fix while we're in here. Let me see why it was misbehaving, I was just having an issue where it was trying to reopen projects even when I was normally launching the app. Let me figure out why

PR #3153, opened 2026-08-22
show changelog after update

### LINKED ISSUE #3028: Show changelog in app
For stable releases which have a detailed changelog it'd be useful if it showed up as a popup after updating. 
That way users don't have to go out to github or discord to read the changelog and as the changelog is in the same location as the updater.

notaspirit:
# show changelog after update

All changelogs between the version that was updated from upto and including the version updated to will be shown on startup (assuming they are present in CHANGELOG.md which contains changelogs since 8.18.0)

**Implemented:**
- added SyncFusions Markdown packages
- implemented showing changelog after an update
- MarkdownInfoDialog: dialog accepting a custom title and body accepting arbitrary markdown with an ok button

**Steps to Verify**
- for the showing of the changelog after an update:
- create a file under `%appdata%\REDmodding\WolvenKit\changelog.md` with some markdown content 
- build and run the branch
- see changelog info dialog
- observe file aforementioned file being deleted

- for the updater part:
- downgrade the version in WolvenKit.csproj to an older version like e.g. `8.18.1`
- build and run WolvenKit
- trigger an update
- observe `%appdata%\REDmodding\WolvenKit\changelog.md` being created with the changelog containing all changelogs available since last installed version

**Additional notes:**
closes #3028 

The inclusion of the new syncfusion packages should be checked with @rfuzzo to ensure they are covered under our existing license. 

image

### DIFF
--- a/WolvenKit.App/Services/UpdateService.cs
+++ b/WolvenKit.App/Services/UpdateService.cs
@@ -5,6 +5,7 @@ using System.IO.Compression;
 using System.Linq;
 using System.Net.Http;
 using System.Security.Cryptography;
+using System.Text;
 using System.Threading.Tasks;
 using Semver;
 using WolvenKit.App.Helpers;
@@ -21,6 +22,9 @@ public class UpdateService : IUpdateService
     private readonly ILoggerService _loggerService;
     private readonly ISettingsManager _settingsManager;
     private readonly HttpClient _httpClient;
+
+    private readonly string _localChangelogPath = Path.Join(ISettingsManager.GetAppData(), "changelog.md");
+    private const string s_remoteChangelogPath = "https://raw.githubusercontent.com/WolvenKit/WolvenKit/refs/heads/main/CHANGELOG.md";
 
     public UpdateService(ILoggerService loggerService, ISettingsManager settingsManager)
     {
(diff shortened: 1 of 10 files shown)


notaspirit:
Actually should be fine to merge even without fuzzo checking (since he seems to be busy rn), don't get a pop up complaining about the license and `SyncfusionLicenseProvider.ValidateLicense([Platform.UIComponent, Platform.WPF, Platform.Markdown, Platform.Utility])` returns true so I highly doubt that it doesn't cover it.

WolvenKit.App/Services/UpdateService.cs
```diff
@@ -199,4 +221,58 @@ private string GetRepositoryName() => _settingsManager.UpdateChannel == EUpdateC
...
+                localVersionStartIndex = i;
+            }
+
+            if (remoteVersionStartIndex != -1 && localVersionStartIndex != -1)
+            {
+                break;
+            }
+
+            i++;
+        }
```
poirierlouis, line 263: Is it required to compare version numbers in files, instead of computing and comparing the SHA256 of both files?

poirierlouis (approved):
I left a comment, LGTM otherwise.

WolvenKit.App/Services/UpdateService.cs
notaspirit: Which two files are you referring to?

poirierlouis: Oh my bad, completely misunderstood what this function does.

PR #3155, opened 2026-08-25
Remove Mod Manager

### LINKED ISSUE #2988: Sunset Mod Manager
As discussed, the mod manager is not useful nor used (as users should just use a dedicated mod manager of their choice) and as such is just a liability for maintenance and should be removed entirely.
However since it is technically a "breaking change" as an entire feature is removed it should be bundled with the major release.

notaspirit:
# Remove Mod Manager

**Implemented:**
- removes mod manager which nobody used either way

**Additional notes:**
closes #2988


gistya (approved):
I can't speak to whether anyone actually uses the Mod Manager... I'd imagine most people are using third party stuff? Vortex etc.? But anyway, the PR looks fine from a code standpoint. App boots up, runs, etc.

notaspirit:
Yeah there is no real reason to use the one in wolvenkit and in my years of modding cyberpunk I've never seen anyone use it. Removing it has been discussed in the dev channel.

PR #3159, opened 2026-08-27
Scene Editor: Import dialogue creates section node

Akiway:
# Scene Editor: Import dialogue creates section node

**Implemented :**
- Importing dialogues now creates a pre-filled section node (optional, enabled by default)

**Changed :** 
- Import format changed (v1 -> v2) :
  - Added property `startTime` to dialogue lines
  - Durations properties are now in milliseconds (they were in seconds before)

https://github.com/user-attachments/assets/c5b4d932-7272-41de-a304-034c4dff2163

### DIFF
--- a/WolvenKit.App/Helpers/DialogueImportHelper.cs
+++ b/WolvenKit.App/Helpers/DialogueImportHelper.cs
@@ -6,6 +6,8 @@ using System.Linq;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using WolvenKit.RED4.Types;
+using static WolvenKit.RED4.Types.Enums;
 
 namespace WolvenKit.App.Helpers;
 
@@ -17,7 +19,7 @@ namespace WolvenKit.App.Helpers;
 public sealed class ImportedDialogueLine
 {
     /// <summary>Locstring id the recording is registered under.</summary>
-    public ulong LocStringId { get; init; }
+    public CRUID LocStringId { get; init; }
 
     /// <summary>
     /// The one subtitle the scene's locstore gets for this line, picked from the variants below.
@@ -69,14 +71,11 @@ public sealed class ImportedDialogueLine
     /// </summary>
     public uint? StartTimeMs { get; init; }
 
-    /// <summary>
-    /// The <c>locVoiceoverContext</c> the line is recorded under, by name - "Vo_Context_Quest" and
-    /// the like. Kept as the export wrote it; whether the game knows the name is checked later.
-    /// </summary>
-    public string VoContext { get; init; } = "";
+    /// <summary>Voiceover context for the line.</summary>
+    public CEnum<locVoiceoverContext>? VoContext { get; init; }
 
-    /// <inheritdoc cref="VoContext"/>
-    public string VoExpression { get; init; } = "";
+    /// <summary>Voiceover expression for the line.</summary>
+    public CEnum<locVoiceoverExpression>? VoExpression { get; init; }
 
     /// <summary>
     /// Where the line falls in the conversation, counting from 1, or 0 where the export did not say.
@@ -118,33 +117,6 @@ public sealed class DialogueImportPayload
     public bool IsValid => Error is null && Lines.Count > 0;
 
     public static DialogueImportPayload Failed(string error) => new() { Error = error };
-}
-
-/// <summary>
-/// A screenplay line the scene already carries. An import never writes one of these a second time -
-/// two entries on one recording cannot be told apart afterwards - but a section can still play the
-/// entry that is there, which is how a conversation imported in pieces gets a section.
-/// </summary>
-public sealed class ExistingSceneLine
-{
-    /// <summary>Locstring id the entry is keyed on, which an import is matched against.</summary>
-    public required ulong LocStringId { get; init; }
-
-    /// <summary>
-    /// The entry's item id, which a section's dialogue event points at to play it. Null where the
-    /// entry was left on <see cref="SceneEditingHelper.UnassignedScreenplayItemId"/> - the id an
-    /// event means "points at nothing" by - in which case the import gives it one.
-    /// </summary>
-    public required uint? ScreenplayLineId { get; init; }
-
-    /// <summary>
-    /// Who the scene has saying it, or null where the entry names nobody. Read off the entry, not
-    /// the export: the entry is not being rewritten, so the scene's answer holds.
-    /// </summary>
-    public uint? SpeakerActorId { get; init; }
-
-    /// <inheritdoc cref="SpeakerActorId"/>
-    public uint? AddresseeActorId { get; init; }
 }
 
 /// <summary>
@@ -301,7 +273,7 @@ public static class DialogueImportHelper
     /// <c>startTime</c>. Version 1 wrote seconds, and the unit cannot be told from the number - 2 is
     /// either two seconds or a fifth of a frame - so only the version says which.
     /// </summary>
-    private const int MillisecondTimingVersion = 2;
+    private const int millisecondTimingVersion = 2;
 
     /// <summary>
     /// The longest a line's stated length can be and still be believed. No single recording runs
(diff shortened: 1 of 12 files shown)


WolvenKit.App/Helpers/DialogueImportHelper.cs
```diff
@@ -82,6 +120,33 @@ public sealed class DialogueImportPayload
...
+
+    /// <summary>
+    /// Who the scene has saying it, or null where the entry names nobody. Read off the entry, not
+    /// the export: the entry is not being rewritten, so the scene's answer holds.
+    /// </summary>
+    public uint? SpeakerActorId { get; init; }
+
+    /// <inheritdoc cref="SpeakerActorId"/>
+    public uint? AddresseeActorId { get; init; }
+}
```
notaspirit, line 148: What's the purpose of this, why not use scnscreenplayDialogLine? Also use red engine types rather than reducing to basic C# primitives.

WolvenKit.App/ViewModels/Documents/SceneGraphViewModel.cs
```diff
@@ -758,11 +796,29 @@ private void ImportDialogue()
...
-                    (createEmbeddedText ? $", {addedTexts} with embedded text" : "") +
-                    (addedActors > 0 ? $", {addedActors} speaker/addressee assignment(s)" : "") +
+                    (wroteToStore
+                        ? $"Imported {addedLines} dialogue line(s)" +
+                          (addedOptions > 0 ? $" and {addedOptions} choice option(s)" : "") +
+                          (createEmbeddedText ? $", {addedTexts} with embedded text" : "") +
+                          (addedActors > 0 ? $", {addedActors} speaker/addressee assignment(s)" : "") +
+                          (reusedLines > 0 ? $". Reused {reusedLines} line(s) the scene already had" : "")
+                        : $"Reused {reusedLines} line(s) the scene already had, writing nothing to the " +
+                          "screenplay store") +
```
notaspirit, line 813:
```suggestion
                        ? $"Imported {addedLines} dialogue line(s)" +
                          (addedOptions > 0 ? $",\n {addedOptions} choice option(s)" : "") +
                          (createEmbeddedText ? $",\n {addedTexts} with embedded text" : "") +
                          (addedActors > 0 ? $",\n and {addedActors} speaker/addressee assignment(s)" : "") +
                          (reusedLines > 0 ? $". Reused {reusedLines} line(s)" : "")
                        : $"Reused {reusedLines} line(s), no new lines added.") +
```

the "and" was in the middle of the list, with new lines it's easier to read and adjusted the wording on the reused lines.

WolvenKit.App/ViewModels/Documents/SceneGraphViewModel.cs
```diff
@@ -771,6 +827,68 @@ private void ImportDialogue()
...
+        /// <summary>
+        /// Lays an import out as a section node in the graph, where the user asked for one.
+        /// </summary>
+        /// <param name="lines">
+        /// The lines, each already carrying the item id it plays. Empty for an import of nothing but
+        /// choice options, since an option is picked rather than played.
+        /// </param>
+        /// <returns>What was built, or null where nothing was.</returns>
+        private BuiltDialogueSection? CreateSectionForImport(
+            DialogueImportDialogViewModel dialog,
```
notaspirit, line 839: dialog isn't in the xml docs

WolvenKit.App/Helpers/SceneSectionBuilder.cs
```diff
@@ -0,0 +1,298 @@
...
+        return wasEstimated
+            ? EstimateDuration(line.Text)
+            : Math.Clamp(line.DurationMs, MinLineDurationMs, MaxLineDurationMs);
+    }
+
+    /// <summary>
+    /// What to list the section under in the scene's notable points, which is what the canvas puts
+    /// on the node's marker bar. The conversation name, cut down to an identifier.
+    /// </summary>
+    public static string GetNotablePointName(string? conversationName)
```
notaspirit, line 218: the naming and description are a bit misleading, it's not really making it a unqiue identifier it's just sanitizing the name

WolvenKit.App/Helpers/SceneSectionBuilder.cs
```diff
@@ -0,0 +1,298 @@
...
+    public uint? StartTimeMs { get; init; }
+
+    /// <summary>Used only to estimate a duration the export did not give.</summary>
+    public string Text { get; init; } = "";
+
+    /// <summary>A <c>locVoiceoverContext</c> by name, e.g. "Vo_Context_Quest". Empty for the default.</summary>
+    public string VoContext { get; init; } = "";
+
+    /// <summary>A <c>locVoiceoverExpression</c> by name, e.g. "Vo_Expression_Spoken".</summary>
+    public string VoExpression { get; init; } = "";
```
notaspirit, line 42: As mentioned in another comment, use redtypes and enums when available. 
Using C# primitives just muddies the intention when reading the code and creates a failure point during conversion to the type. 
If you need a string representation you easily convert to a given primitive (e.g. number or string) when you have to at the end point or point of comparison.

WolvenKit.App/Helpers/SceneSectionBuilder.cs
```diff
@@ -0,0 +1,298 @@
...
+        var voParams = new scnDialogLineVoParams();
+
+        if (Enum.TryParse<Enums.locVoiceoverContext>(line.VoContext, ignoreCase: true, out var context) &&
+            Enum.IsDefined(context))
+        {
+            voParams.VoContext = context;
+        }
+
+        if (Enum.TryParse<Enums.locVoiceoverExpression>(line.VoExpression, ignoreCase: true, out var expression) &&
+            Enum.IsDefined(expression))
```
notaspirit, line 274: I fail to see how IsDefined can return false here as if it isn't a valid enum value TryParse would already return false.

WolvenKit.App/Helpers/SceneSectionBuilder.cs
```diff
@@ -0,0 +1,298 @@
...
+    /// rather than cut to the estimate cap, so a timed monologue is not truncated. Matches the
+    /// ceiling <see cref="DialogueImportHelper"/> accepts a stated length at; move the two together.
+    /// </summary>
+    public const uint MaxLineDurationMs = 300_000;
+
+    /// <summary>
+    /// Characters a second to read an untimed line at. Near Cyberpunk's own delivery, and a guess
+    /// the user can drag on the timeline afterwards.
+    /// </summary>
+    private const double EstimatedCharactersPerSecond = 14;
```
notaspirit, line 108:
```suggestion
    private const double s_estimatedCharactersPerSecond = 14;
```

adjusted to follow our naming convention

WolvenKit.App/Helpers/DialogueImportHelper.cs
```diff
@@ -419,5 +629,15 @@ private sealed class LineDto
         public string? FemaleLipsyncAnim { get; set; }
         public string? MaleLipsyncAnim { get; set; }
         public string? Kind { get; set; }
+
+        /// <summary>A <see cref="JsonElement"/> for the same reason the locstring id is.</summary>
```
notaspirit, line 633: what's up with the json elements? locstringid doesn't have an explanation nor does the pr introducing it.

If it's about the read write logic use a json converter to handle the conversion once on json serialization.

WolvenKit/Views/Dialogs/DialogueImportDialogView.xaml
```diff
@@ -490,29 +493,54 @@
...
+
+                            <CheckBox
+                                Margin="{DynamicResource WolvenKitMarginTinyTop}"
+                                HorizontalAlignment="Right"
+                                VerticalAlignment="Center"
+                                FontSize="{DynamicResource WolvenKitFontSubTitle}"
+                                Content="Create a section node"
+                                ToolTip="Adds a section node to the graph that plays the imported lines"
+                                IsChecked="{Binding CreateSectionNode}" />
+                        </StackPanel>
```
notaspirit, line 543: should be left aligned, at least within their own group so that the check boxes line up vertically

notaspirit (changes requested):
There are some blocking issues as well as some non blocking ones, some suggestions and questions.

WolvenKit.App/Helpers/SceneSectionBuilder.cs
Akiway: `https://github.com/WolvenKit/WolvenKit/blob/main/docs/CONTRIBUTING.md` point to [general .NET naming conventions](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-fields) which says the exact opposite.

WolvenKit.App/ViewModels/Documents/SceneGraphViewModel.cs
Akiway: Regarding new lines, I only applied them for each new sentence. Otherwise, the visual layout looks odd.

WolvenKit.App/Helpers/SceneSectionBuilder.cs
Akiway: camelCase used until we decide on a naming convention

WolvenKit.App/Helpers/DialogueImportHelper.cs
Akiway: I replaced the JsonElement fields with typed DTO properties and custom converters.

PR #3160, opened 2026-08-27
Fix folders collapsing on import

gistya:
# Fix folders collapsing on import

**Fixed:**
- fixes an issue where folders would collapse on import (or fail to properly expand)

**Additional notes:**
Addresses an issue reported on discord.


notaspirit (changes requested):
It seems to only open the last saved state of the folders when adding files to a project from e.g. the asset browser. 
When the expected behavior is that it opens all folders in the added items path regardless of if they were last seen opened or closed.

gistya:
> It seems to only open the last saved state of the folders when adding files to a project from e.g. the asset browser. When the expected behavior is that it opens all folders in the added items path regardless of if they were last seen opened or closed.

Latest push addresses your concern. 

Had to rearchitect how this works as a result of decoupling of FileTree from the Flat view.

notaspirit (approved):

PR #3179, opened 2026-09-10
update default layout optimizing for document space

notaspirit:
# update default layout optimizing for document space

The previous default layout dedicated very little space to the document tabs which is one of the most important and most used UI elements given that it is where file contents are modified. 

Import and Export windows have been moved to the sidebar, ensuring they do not take up unnecessary space when not used while still being visually available to the user.

The asset browser has been moved down to share horizontal space with the log viewer, keeping it's general aspect ratio. 

Old Layout vs Proposed with content at 1080p
WolvenKit Previous Default Layout 1080p With Content
WolvenKit Previous New Layout 1080p With Content

Old Layout vs Proposed without content at 720p
WolvenKit Previous Default Layout 720p
WolvenKit Previous New Layout 720p


manavortex:
This will lead to people not finding the Import/Export tool because they're used to its old position since 8.5 or earlier. Hell, I didn't find it without reading where it was supposed to be first. 
Plus, this will require updating a _lot_ of screenshots across the wiki.

notaspirit:
With the current layout you get the issues as in mod dev chat where people literally don't find the document editor

notaspirit:
But I do agree that this change needs to be communicated well.

notaspirit:
I'll merge it into the nightly with an announcement in the updates channel so we can gather user feedback on this.

manavortex:
The person in mod dev chat didn't look at any of the linked guides, which had screenshots that would have helped alleviating the problem, and generaly expected to be carried. I'm not saying the current layout is perfect, but the solution should be something like "if the display resolution is lower than X on first startup, reduce the UI scale" and not "change the established UI for literally everyone on behalf of one lazy freeloader".

If you insist on merging this, please update the wiki screenshots as well.

notaspirit:
Of course, updating the wiki is a given as it would cause a lot of confusion for new users if there was a large discrepancy.
I'm not opposed to feedback and iterations on the specifics of the implementation but I do see the layout dedicating so little space to the documents as a real issue that needs resolving. 

Having less about equally sized elements competing for space and user attention will overall make it less intimidating as there is a clear hierarchy.
The currently proposed layout follows what most IDEs have as their default layout so it is not something completely novel. 

Perhaps we can create around three proposed layouts, have a poll about which ones users prefer and put the winner into the nightly for further feedback.

manavortex:
I'd rather that this not hit main without a full update of all wiki screenshots. Plus, I think the current implementation will have any layout changes affecting everyone who hasn't saved a personal default laoyut? Which would mean pretty much everyone.

notaspirit:
Not immediately everyone, the default layout gets copied from the app directory to appdata only when the default layout in appdata doesn't exist or the user deliberately resets it. 

So it would only apply after someone reset their default layout via the ui or the file / entire appdata folder got deleted. 

Nevertheless it is a change affecting many users hence the plans to make a poll and put the resulting favorite into the nightly for a while with explicit communication about it.

notaspirit:
Another possible layout could look something like this, the import export tools aren't set to only be in the side ribbon like in the layout proposed above, instead they are merely stacked. 

WolvenKit Default Layout Proposition 2 1080p With Content

notaspirit:
@poirierlouis would love to hear your opinion on this as well

poirierlouis:
This is the default layout I use in my case:
image

pros:
- no *lost* tabs in the vertical right side
- focus on one task at a time (e.g. explore using Asset Browser > Add to project > use Export Tool)
- easier for new users (a feature = a tab = a view/tool), can discover them one after the other
- don't overload the UI with X different features

cons:
- Logs may take a lot of space for nothing
- no multi-task
- what you say?

Additional notes:
- we need to show a dedicated, explicit, colored icon for Asset Browser, Import Tool, Export Tool (better UX in all cases)
- wiki shouldn't need refactoring of all images
  - user is taken by the hand too much
  - *and/or*
  - explanation/content is too much dependant on the layout
- we could define a new default layout (whatever is accepted) and preserve the old one:
  - View -> Layout -> can contain a new button to select the `Legacy layout` / `Old layout`
  - on update, a popup can allow the user to keep current `Legacy layout` or try the new one.
  - on first startup, a popup (like above) ask the user to select between `Legacy layout` or `Default layout`.

Regarding the wiki, as always, I'm not used to its users and the bare minimum to assume. The better UX idea is hopefully a good way to show what to look for (without getting into default layout, new layout, me layout, you layout...).

manavortex:
> user is taken by the hand too much
Yes, and there's a reason for that. I invite you to spend some time in the dev channels to help the complete newbies finding their way around things.

In fact, that's the target audience for the default layout: people who are opening WKit the first time. We need to keep that in mind before hiding stuff away.

poirierlouis:
Which channel do you have in mind?

Hum, so new users are still asking their way around even with the current layout showing all tools at once?

manavortex:
mod-dev-chat, mostly? Total newbs come by every few days.

Most people are incredibly overwhelmed with basic workflows in WKit, since it'S the first time they're working with that kind of software. Guides are very step-by-step and detailed, but they're written with the assumption that e.g. the export tool will be visible by default, and so the steps go like "use the export tool to export your mesh" and not "find your export tool (link or explanation on how to make it visible)". 

Having specialized layouts for specialized tasks is sure nice (I have a different project layout saved in my quest project), but experienced users know how to customize layouts.

poirierlouis:
Alright was wondering how much dev channels are used for WKit stuff 😅.

Well that's where my idea of adding dedicated icons may help.
In my layout, all tools are accessible with one click.
In case this one (or another) is used, including dedicated icon... wiki might "only" require updating messages from:

> use the export tool to export your mesh

to

> use the export tool [insert icon] to export your mesh

It shouldn't need to change all images, it should be explicit for any users.
wdyt?

seberoth:
Just some crazy idea from my end:
Ship multiple default layouts and let the user pick one:
- On project creation (with option to disable this / always used a preferred layout)
- Or on first start
- Via menu

Could do a little "preview" (mock up) for selecting it on start / project creation (with a small description) and tell them that the current one is used in the wiki for tutorials.

Like:
 - Layout 1 (current): The default Layout used in Wikis and tutorials. A good place to start
 - Layout 2 (yours): A more minimalist layout with focus on file (CR2W) editing

notaspirit:
That would be interesting, especially it has a system and user layout section where system ones are sourced from the wolvenkit app dir (like the current default layout) and from the repo (like scripts and templates) with user ones being local only as always.

Ofc more work to implement but also a good opportunity to look into the issues casued by resetting the layout in the current implementation.

PR #3181, opened 2026-09-13
Lipmap : display scene path as DepotPath

Akiway:
# Lipmap : display scene path as DepotPath

**Implemented:**
Display `scenePaths` and `scenePreviewPaths` as DepotPath

image


WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.Value.cs
```diff
@@ -120,18 +124,7 @@ public void CalculateValue()
...
-            {
-                Value = depotPath.GetResolvedText().NotNull();
-            }
-            else
-            {
-                Value = depotPath == ResourcePath.Empty
-                    ? "null"
-                    : $"{(ulong)depotPath}{_hashService.GetGuessedExtension(depotPath)}";
-            }
+            Value = GetDepotPathValue(rr.DepotPath);
```
manavortex, line 127: I think the fallback is needed to resolve hashes that were calculated with previous versions of the hashing algorithm, like when you click the target button on the right side of the mod browser search bar

manavortex (changes requested):
LGTM, one comment (think that shouldn't be removed, probably?)

WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.Value.cs
Akiway: image

manavortex (approved):

WolvenKit.App/ViewModels/Tools/ChunkViewModelExtended/ChunkViewModel.Value.cs
Akiway: _(Screenshot is unrelated to your question, sorry)_

To answer your comment : I did not removed the fallback, it the if/else condition has simply be moved as-is to the `GetDepotPathValue` function.

manavortex (approved):

PR #3182, opened 2026-09-14
Scene Editor : Generate lipsync anim sets

Akiway:
# Scene Editor : Generate lipsync anim sets

**Implemented:**
Add a new scene menu, in the Scene Editor, with a button to generate vanilla lipsync anim sets.

What it does, in short : **it generates all the files needed in order for the lipsync of a scene to work, and connect them in the scene**.

In long : 
1. It list all the dialogue lines in `screenplayStore` that are tied to an actor (playerActors are ignored), and it list all the voiceTags set on the actors.
2. It looks for already existing `.lipmap` and `.anims` files in the project, to reuse existing elements and reduce process duration (useful when updating a scene for which you already generated lipsync).
3. It seaches in vanilla `.lipmap` all the data needed, for each actor's voiceTag in the scene (it fallbacks to a total search if the dialogue line doesn't match the voiceTag).
4. In the project, it generates the files, per localization available in the Wkit install : one `.lipmap`, one `.anims` per actor, and a `.xl` file to append the lipmap to the vanilla one.
5. In the scene, it set the `resourcesReferenes.lipsyncAnimSets` with the newly created `.anims` and assign them to each actor.

___

The new button : 
image

Generated files : 
image

_Note : `.anims` are grouped by scene-named groups._


manavortex (approved):

WolvenKit.RED4/Types/ClassesExt/Properties/animLipsyncMapping.cs
```diff
@@ -9,4 +9,63 @@ public CArray<CUInt64> ScenePreviewPaths
         get => GetPropertyValue<CArray<CUInt64>>();
         set => SetPropertyValue<CArray<CUInt64>>(value);
     }
-}
\ No newline at end of file
+
+    partial void PostConstruct() => ScenePreviewPaths = [];
```
notaspirit: You can't modify the the constructor, post constructor or default values manually. They are generated from native definitions. This causes serialization issues. 
(Default) templates are specifically designed for solving this kind of need without affecting serialization.

notaspirit (changes requested):
Haven't verified anything else but the RED4 changes cannot be merged as is.

WolvenKit.RED4/Types/ClassesExt/Properties/animLipsyncMapping.cs
manavortex: It's in ClassesExt, I thought that one was only used for the app?

notaspirit: PostConstructor() is called at the end of the constructor and as such is equivalent to modifying the constructor.

Akiway: I removed the PostConstructor

manavortex: I remember running into this before now - you can probably create a factory (or use one of the templates) :)

